PR #234, opened 2017-05-23
Fix system files format (ignoring files of local git artifact)

alexey-igrychev:
* System files must always consist default build directory path.
* External build directory cannot affect on stages signatures.
* Build directory can be on any level of project file hierarchy (but isn't good practice).


alexey-igrychev:
#237

PR #309, opened 2017-06-29
Kube secret extract

alexey-igrychev:
* kube secret extract
  * --values options: decode secret values file
  * visible input
* kube doc reference


alexey-igrychev:
#313

PR #431, opened 2017-11-09
dapp dimg cleaning commands: modify caching and refactor

alexey-igrychev:


alexey-igrychev:
close https://github.com/flant/dapp/issues/359

PR #615, opened 2018-03-07
Cleanup repo: Scan additional resources for used images in cluster.

diafour:
- Support for 1.6, 1.7, 1.8, 1.9 API endpoints
- client get version from cluster on first api call


alexey-igrychev:
много отступов лишних по файлам

PR #846, opened 2018-08-07
 dapp dimg cleanup repo: customized tags politics

alexey-igrychev:


alexey-igrychev:
close https://github.com/flant/dapp/issues/816

PR #891, opened 2018-08-28
New article on dapp secret

may-cat:
Article on dapp secret refactored


may-cat:
Closes https://github.com/flant/dapp/issues/850

PR #1142, opened 2018-12-16
Add more watchers in deploy command

diafour:
- watch for jobs that are not hooks
- watch for pods, daemonsets, statefulsets

A support of https://github.com/flant/kubedog/pull/33


distorhead:
https://github.com/flant/dapp/issues/1050

PR #1602, opened 2019-07-04
Fix render dappfile when path contains spaces

andrew-demb:


andrew-demb:
CI fail not related

distorhead:
This fix will still fail when running on a project name with single quote in the name.

https://github.com/flant/werf/pull/1604

PR #1659, opened 2019-07-22
Optimize git worktree switch procedure; fixes for git operations cache

distorhead:


distorhead:
Fixes https://github.com/flant/werf/issues/1644.

PR #1831, opened 2019-10-10
[tests] Git stages and images cleanup/purge

alexey-igrychev:


scripts/ci/git.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash -e
+
+git config --global user.name "borya"
+git config --global user.email "borya@flant.com"
```
diafour: This should be `boris` =)

PR #1976, opened 2019-12-06
[stapel] Yum ansible module support in stapel for centos 6 and 7

distorhead:


distorhead:
Related to
- https://github.com/flant/werf/issues/1983
- https://github.com/flant/werf/issues/1891
- https://github.com/flant/werf/issues/1798

PR #1999, opened 2019-12-14
Fix link to complete feature list

andrew-demb:


distorhead:
Hmmm, why change? Original link works ok

andrew-demb:
Checked from mobile - github collapses readme and anchor doesn't scroll view to heading.
Now checks from desktop - works, as expected.

If we don't care about mobile - we can close it.

PR #2072, opened 2020-01-20
Use `allowFailuresCountMultiplier` 1 for jobs

### LINKED ISSUE #2071: werf.io/failures-allowed-per-replica cannot be used with k8s job
werf ignores `werf.io/failures-allowed-per-replica` annotation for job and always fail deploy, when at least one failure reached.

I think reason in this line - https://github.com/flant/werf/blob/648426b69f1d8c75fb632f02ebf4e7d99934fedc/pkg/deploy/helm/resources_waiter.go#L135

Werf provide `0` (zero) multiplier for annotation value.

**Possible solution:** use `1` (one) multiplier here.

andrew-demb:
Fixes #2071


distorhead:
By this PR allowed failures count will be equal 1 for Job _by default_, which is incompatible with older versions.

Fixed by splitting up default-failures-per-replica and multiplier in https://github.com/flant/werf/pull/2072.

PR #2122, opened 2020-02-09
Add matrix-tests to werf

nabokihms:
### Description:
Allows running matrix tests with `werf helm matrix-tests` command.
FIle `.helm/matrix_test.yaml ` should exist in project directory to make it works.

For more info check README.md file in the matrix-tests source code directory.


pkg/deploy/helm/matrix_tests/README.md
```diff
@@ -0,0 +1,133 @@
...
+Gitlab example:
+```yaml
+Deploy Production:
+  stage: tests
+  script:
+    - type multiwerf && source <(multiwerf use 1.0 beta)
+    - type werf && source <(werf ci-env gitlab --tagging-strategy tag-or-branch --verbose)
+    - werf deploy --stages-storage :local
+      --namespace ${CI_ENVIRONMENT_SLUG}
+      --set "global.env=prod"
```
distorhead: Need to change `--set` example, because `global.env` is set automatically by werf itself.

pkg/deploy/deploy.go
```diff
@@ -134,7 +137,7 @@ func patchLoadChartfile(chartName string) {
 			Version: "0.1.0",
 			Engine:  helm.WerfTemplateEngineName,
 		}
-
+		mu.Unlock()
```
distorhead: It is better to do `defer mu.Unlock()` right after `mu.Lock()` so that failed `LoadChartfileFunc` function (err != nil) in some thread will not deadlock loading process for other threads.

pkg/deploy/helm/matrix_tests/README.md
distorhead: There is parameter of werf deploy named `--env`, that used to set `global.env` and generate helm release name and kubernetes namespace. In gitlab however it is not specified for the `werf deploy` command because `werf ci-env` command sets this parameter implicitly using `WERF_ENV` environment variable.

pkg/deploy/helm/matrix_tests/README.md
```diff
@@ -0,0 +1,133 @@
...
+Create the file named matrix_test.yaml in your chart directory (.helm).
+
+Example for 1x1 matrix (matrix_test.yaml):
+```yaml
+replicas: 1
+```
+
+To increase your matrix dimensional you should use special pointers:
+
+**NOTE:** Matrix tests generate values.yaml files, witch will be combined with values.yaml in your chart directory.
```
distorhead: witch -> which

pkg/deploy/deploy.go
nabokihms: Good point. I fixed this.

pkg/deploy/helm/matrix_tests/README.md
nabokihms: I added variables block to GitLab ci example with WERF_ENV environment variable set to `prod`.

PR #2259, opened 2020-03-13
Remote storages support: added container runtime package and refactor stages storage for distributed builds

distorhead:


distorhead:

PR #2628, opened 2020-08-19
Flant: ivan.khramov fix django article

mastanggt:


may-cat:
moved to https://github.com/werf/werf-guides

PR #2648, opened 2020-09-01
[dockerfile] Handle unsupported Media Type (especially v1 manifest version)

alexey-igrychev:


pkg/container_runtime/container_runtime.go
```diff
@@ -31,6 +31,15 @@ func (runtime *LocalDockerServerRuntime) GetImageInspect(ctx context.Context, re
 	return inspect, err
 }
 
+// PullImage only avaliable for LocalDockerServerRuntime
```
distorhead, line 34:
```suggestion
// PullImage only available for LocalDockerServerRuntime
```

distorhead (changes requested):

distorhead (approved):

Issue #1606, opened 2019-07-05
Helm 3 support


donbeave:
Hello, guys!

Amazing project! Looks like exactly what I’m looking for.

Just some probably weird question for me, if Helm 3 is supported? I know it’s still alpha, but it still works well for us even in this stage, will be happy to use werf with third version of Helm. If not supported yet, pls let me know your roadmap to upgrade to next version of Helm.

Best regards,
Alexey Zhokhov.

distorhead:
Hi! Thanks!

Werf now have built-in helm 2 client and tiller. It works the same way as helm 2 (same release storage, fully compatible, etc.). But the only difference is that tiller process and werf process are the same single process. So to use helm 3 we must compile werf agains new helm 3 codebase.

Werf is not compiled with helm 3 for now. The main reasons are: there should be clear and durable way to migrate from helm 2 installations to helm 3 (https://github.com/helm/helm/issues/5892); it is alpha.

We will certainly update to helm 3 some day.

Public roadmap will also be available at the https://werf.io soon.

donbeave:
@distorhead https://github.com/helm/helm/releases/tag/v3.0.0 just released!

StupidScience:
Any plans/ETA here?

javad-hajiani:
👍

distorhead:
Some notes for future implementation:
 - werf should not auto-purge first failed release, remove this dangerous feature when ported to helm 3;
 - update masterminds/sprig library;
 - CTRL-C + wait + atomic release status updates;
 - 3 way merge should solve a problem when a kubernetes cluster has been updated, but there are resources deployed which have outdated api versions definitions in the release.

distorhead:
> Any plans/ETA here?

It is not a priority task for now, because it will add nothing new and valuable to the werf. Werf do have all main features that Helm 3 have: three way merge and tillerless architecture.

But we should switch to Helm 3 in v1.2 version.

Menzorg:
Good day. Thanks, could you say, when 1.2 is planning?

distorhead:
**Update.** Happy to announce, that we have "migration to Helm 3" task in the nearest plans of big tasks. I think we will implement usable Helm 3 support in 1 month approximately.

distorhead:
As for now, there is very early alpha version: v1.1.22-alpha4 — which available in the alpha channel. This version embeds all helm-v3 commands into `werf helm-v3` subcommand and adds support for kubedog-based waiting for resources. For example:

```
[...]

PR #2658, opened 2020-09-04
[helm-v3] Extra annotations and labels support for helm v3

distorhead:
Set werf.io/version annotation by default in werf helm-v3 upgrade/install commands.


distorhead:
Refs https://github.com/werf/werf/issues/1606

PR #2660, opened 2020-09-04
Ability to enable/disable checking of all kube namespaces during cleaning images from registry

identw:
My case: there are several projects in one cluster, each project has access only to certain namespaces. Therefore, the werf cleanup does not work as it tries to extract images from the entire cluster. For example:
```bash
627.191299ms Using werf config render file: /private/var/folders/82/qx_s8s_n3c3016fsx1rbbjsh0000gn/T/werf-config-render-088482510
633.265926ms Using images repo docker registry implementation: default
633.301918ms Using images repo mode: multirepo
633.337162ms -- LocalDockerServerStagesStorage.GetManagedImages cross-stitch
792.055737ms Managed images names: [backend migrations notifications static]
851.228204ms 
851.264639ms ┌ Running images cleanup
863.395068ms │ ┌ Fetching repo images data
863.425751ms │ │ ┌ Fetching repo images
2m42.6553478 │ │ └ Fetching repo images (161.78 seconds)
2m42.6554011 │ └ Fetching repo images data (161.79 seconds)
2m42.6554378 │ 
2m42.6554631 │ ┌ Skipping repo images that are being used in Kubernetes
2m42.6622105 │ │ Getting deployed docker images (context stand) ... (0.51 seconds) FAILED
2m43.1712105 │ └ Skipping repo images that are being used in Kubernetes (0.52 seconds) FAILED
2m43.1817046 └ Running images cleanup (162.33 seconds) FAILED
2m43.2183839 
2m43.2183878 Running time 163.22 seconds
2m43.2320993 Error: cannot get deployed imagesRepoImageList: cannot get Pods images: pods is forbidden: User "system:serviceaccount:example-project:deploy"   ↵
2m43.2600260 cannot list resource "pods" in API group "" at the cluster scope
```
 I've added the --check-all-namespaces (WERF_CHECK_ALL_NAMESPACES) flag that controls this behavior. The default behavior has not changed, but for my case this flag can be turned off, in which case the werf will only check the namespaces that are specified in the contexts of the configuration file.

I may not have corrected the code very nicely, I had to add an additional argument with map namespaces to many functions. But I can change it if necessary for another option, for example, add information about namespaces to `KubernetesContextsClients`.

### DIFF
--- a/cmd/werf/cleanup/cleanup.go
+++ b/cmd/werf/cleanup/cleanup.go
@@ -25,6 +25,7 @@ var commonCmdData common.CmdData
 var cmdData struct {
 	GitHistoryBasedCleanup    bool
 	GitHistoryBasedCleanupV12 bool
+	CheckAllNamespaces        bool
 }
 
 func NewCmd() *cobra.Command {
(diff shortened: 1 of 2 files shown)


distorhead:
This PR enables cleanup scanning only for default kubernetes namespace per each context when `--check-all-namespaces=false`. Is that right? Why exactly default namespace for context?

identw:
> This PR enables cleanup scanning only for default kubernetes namespace per each context when `--check-all-namespaces=false`. Is that right? Why exactly default namespace for context?

No.
`--check-all-namespaces=false` - Scans images only in namespaces that are specified in contexts. For example
```
- context:
    cluster: stand
    namespace: example
    user: stand
  name: stand
- context:
    cluster: prod
    namespace: example1
    user: prod
  name: prod
```

Will scan the namespace "example" of the "stand" context and the namespace "example1" of the "prod" context.

distorhead:
I propose to rename option to the more precise: `--check-context-namespace-only=[true|false]` which is false by default. Because when you specify `--check-all-namespaces=false` it is not clear for the user which namespace will be checked.

Otherwise, PR looks good.

distorhead:
> I propose to rename option to the more precise: `--check-context-namespace-only=[true|false]` which is false by default. Because when you specify `--check-all-namespaces=false` it is not clear for the user which namespace will be checked.
> 
> Otherwise, PR looks good.

Renamed, added in-cluster config support and added option for `werf images cleanup` command in the https://github.com/werf/werf/pull/2663

identw:
@distorhead 
Oh, sorry, I missed a couple of days. Thanks a lot for the refactor

Issue #2867, opened 2020-11-06
Add --follow param for werf-run, werf-converge and werf-compose-up commands


distorhead:
In the follow mode werf should track new commits into project git repository and automatically restart werf-run, werf-converge or werf-compose-up command.

Follow mode should be enabled by a `--follow` flag.

PR #2870, opened 2020-11-06
Follow git HEAD mode (--follow)

alexey-igrychev:


alexey-igrychev:
https://github.com/werf/werf/issues/2867

Issue #1956, opened 2019-11-29
Cross-platform dockerfile stage signature


alexey-igrychev:
Dockerfile signature includes hash sum of files related to _ADD_ and _COPY_ Dockerfile instructions.

There are the following issues:
- Windows and Unix-systems have different EOL and it affects stage signature.
- Docker adds a source with system EOL, therefore, such built images cannot be used cross-platform and have the same signature.

Stapel builder does not have such a problem, each stage works properly and has a permanent signature.

Issue #2874, opened 2020-11-06
Determinism mode


distorhead:
Werf should work in the determinated mode by default.

In this mode:
 - `mount` directive disabled;
 - `env` function disabled in go-templates in werf.yaml;
 - `.Files.Get` go-templates function can only fetch files from the git;
 - `env` function disabled in the helm templates;
 - take `werf.yaml` only from the current git commit;
 - take all helm-related files (templates and values) onlhy from the current git commit.

PR #2926, opened 2020-11-18
[dockerfile, determinism] Get context and dockerfile from git

alexey-igrychev:
fixes https://github.com/werf/werf/issues/1956


distorhead:
Refs https://github.com/werf/werf/issues/2874

PR #3112, opened 2021-01-17
small cleanups and fixes

cristaloleg:
```
pkg/true_git/ls_tree/result.go:69:4: SA4006: this value of `err` is never used (staticcheck)
pkg/build/conveyor.go:1235:16: printf: Errorf format %s reads arg #4, but call has 3 args (govet)
pkg/deploy/lock_manager/lock_manager.go:39:58: SA1012: do not pass a nil Context, even if a function permits it; pass context.TODO if you are unsure about which Context to use (staticcheck)
pkg/path_matcher/git_mapping.go:93:28: unslice: could simplify f.includePaths[:] to f.includePaths (gocritic)
pkg/path_matcher/git_mapping.go:94:28: unslice: could simplify f.excludePaths[:] to f.excludePaths (gocritic)
```

### DIFF
--- a/pkg/build/conveyor.go
+++ b/pkg/build/conveyor.go
@@ -5,6 +5,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"io/ioutil"
 	"os"
 	"path"
 	"path/filepath"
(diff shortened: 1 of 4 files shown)


distorhead:
Needs rebase onto latest master.

PR #3140, opened 2021-01-22
remove dummy golang/example package

cristaloleg:
Package name suggests that it's something for the reference.

:) https://go-proverbs.github.io/

:) https://www.youtube.com/watch?v=PAAkCSZUG1c&t=9m28s


distorhead:
@cristaloleg Please add unit test for this Reverse function.

PR #3141, opened 2021-01-25
Refactor parsing of list arguments

### LINKED ISSUE #3128: WERF_VALUES_*, WERF_SECRET_VALUES_* and WERF_SET_* params sorting
It is hard to understand in which order werf will process `WERF_VALUES_*`, `WERF_SECRET_VALUES_*` and `WERF_SET_*` params. Order of processing in turn will affect resulting values, because the next param will override previous.

Proposal:
 - Change prefix for these params from `WERF_VALUES` to `WERF_VALUES_`, from `WERF_SET` to `WERF_SET_`, etc. So user will always specify explicitly some suffix.
 - Sort params by alphanumeric rules. So user will have an ability to specify `WERF_VALUES_002=second.yaml`, `WERF_VALUES_001=first.yaml` and the `second.yaml` will always override `first.yaml`.

Also fix CLI documentation examples.

### LINKED ISSUE #3137: WERF_SET(_STRING) env vars can't be combined with command line arguments --set(-string)
`--set` argument passed via command line breaks all `WERF_SET_*` environment variables, so that all the values set with `WERF_SET_*` won't be passed to templates. Same thing with `--set-string` and `WERF_SET_STRING_*`.

You can combine `--set` with `WERF_SET_STRING*` and `--set-string` with `WERF_SET_*` though.

Related: #3128

ilya-lesikov:
Fixes https://github.com/werf/werf/issues/3128
Fixes https://github.com/werf/werf/issues/3137


cmd/werf/common/common.go
```diff
@@ -1073,6 +1064,45 @@ func GetNamespace(cmdData *CmdData) string {
 	return *cmdData.Namespace
 }
 
+// FIXME: inconsistent behaviour:
+// "WERF_SET_TEST=test=1val" and "--set "test=2val"" results in "test: 2val", but
+// "WERF_SET_TEST=test=3val" and "--set "test=2val"" results in "test: 3val";
+// "...test=true" and "...test=false" in any order results in "test: true"
+func GetSet(cmdData *CmdData) []string {
+	envSet := predefinedValuesByEnvNamePrefix("WERF_SET_", "WERF_SET_STRING_", "WERF_SET_FILE_")
```
distorhead, line 1072: Why WERF_SET_ + WERF_SET_STRING_+ WERF_SET_FILE_ here?

The desireable behaviour of `GetSet` is to return only `WERF_SET_` params without `WERF_SET_STRING_` and `WERF_SET_FILE_`.

cmd/werf/common/common.go
```diff
@@ -1073,6 +1064,45 @@ func GetNamespace(cmdData *CmdData) string {
...
+	result := util.UniqStrings(append(envSet, *cmdData.Set...))
+	sort.Strings(result)
+	return result
+}
+
+func GetSetString(cmdData *CmdData) []string {
+	envSetString := predefinedValuesByEnvNamePrefix("WERF_SET_STRING_")
+	result := util.UniqStrings(append(envSetString, *cmdData.SetString...))
+	sort.Strings(result)
```
distorhead, line 1081: Sort in the predefinedValuesByEnvNamePrefix function by the environment variable names.

distorhead (changes requested):

cmd/werf/common/common.go
ilya-lesikov: sort removed from all common.GetSet*/Get*Values functions

PR #3342, opened 2021-03-11
Generalization of the calculating git repository files checksum 

alexey-igrychev:
- Simplify GetOrCreateChecksum method interface
  - the method creates or gets a checksum from the cache
  - the previous logic was specific to git mapping (it is taken out and based on the new implementation)
- Use caching calculating the dockerfile build context checksum

based on https://github.com/werf/werf/pull/3341


alexey-igrychev:
solved in https://github.com/werf/werf/pull/3342

PR #3344, opened 2021-03-11
[docs] added news and updates block

valent-ex:


z9r5:
Cherry-picked into #3346

PR #3509, opened 2021-04-21
[docs] refactoring to use .base_werf in ci

alexey-gavrilov-flant:


alexey-gavrilov-flant:
there is a drawback when switching the version, it does not change in the examples, but I do not know what to replace it with so that it is a dynamic version

alexey-igrychev:
> there is a drawback when switching the version, it does not change in the examples, but I do not know what to replace it with so that it is a dynamic version

Each werf version has its own documentation.

Issue #3661, opened 2021-08-11
How is it possible to access werf image names from subcharts?


kayanme:
I am using some base template charts for creating different services using parameters in derived templates.
One of parameters is image repo/tag, so I need to take it from werf variables.
Actually, as I can see, there are two ways:
1. Using .Values.global. According to changelog "image" was the global template variable (currently not), so it is impossible now.
2. Using dirty hack aliasing subcharts as "werf" in dependencies. They will see the image map, but different dependencies, which use the same alias, rewrite template files of each other, which is also not good.

How can I give the subcharts image map in correct way?
Werf version is 1.2.

ilya-lesikov:
Hi, there is the ongoing issue #3411 to make it possible to pass the image name/tag to the subcharts and it is high on our priority list. We will start working on this soon.

For now you either keep all of your templates in `.helm/templates` without using subcharts, or, if you have to use subcharts, use `--report-path` parameter of `werf build` to save image names/tags into a file, then get them from the file and pass them via `--set` flags to `werf converge` into your subcharts.

kayanme:
I made a patch for me to put an image map back to helm global values again.
I can make a pull request if this solution is acceptable.

Issue #3411, opened 2021-03-23
Forward values with werf image path to subcharts


tuchinsky:
Use custom parameters in dependencies list in parent `Chart.yaml` for forward value with image path to subchart

For example:
```
apiVersion: v2
name: parentChart
version: 1.0.0
dependencies:
- name: subChart
  version: 1.0.0
  repository: https://charts.flant.com
  condition: subchart.enabled
  alias: subchart
  werfAllowImages:
  - frontend
  - backend
```

in subchart value with image path can be `.Values.global.werf.image.frontend` or `.Values.subchart.werf.image.frontend`

PR #3667, opened 2021-08-13
Added image map to global values

kayanme:
#3361


alexey-igrychev:
We have dropped the use of global in v1.2 because we think it is the wrong way.  

In the coming weeks, we plan to implement the ability to explicitly specify the images – that need to be cast in sub-charts – in _Chart.yaml_ (https://github.com/werf/werf/issues/3411). 

At the moment, these values can only be passed manually when the `--set` option (https://github.com/werf/werf/issues/3661#issuecomment-897423720).

In any case, we welcome feedback and outside contributions.

PR #3705, opened 2021-09-10
Fixing kubernetes synchronization short notation bug

preved911:
Hello.
On latest version, when `WERF_SYNCHRONIZATION` is specified in short fomat(like `kubernetes://default`) i get error, if i'm using custom kubeconfig path(defined over `KUBECONFIG` env var).


distorhead:
Hi! Thanks for pointing out and fixing this problem.

Also created additional PR with some more corrections:
1. Use all default kube settings, not only kube-config-path-merge-list, when --synchronization param lacks explicit configuration specs.
2. Do not use default kube settings when --synchronization param contains explicit kube-config path.

Issue #3768, opened 2021-09-30
Race condition between werf-build and werf-host-cleanup procedures


distorhead:
Werf-host-cleanup procedure could delete git-archives and git-patches that are in use by some werf-build process.

Possible solution: only delete git-archives and git-patches that are older than ~5hours. This should work without any locks.

distorhead:
Fixed by the https://github.com/werf/werf/pull/3821

PR #3821, opened 2021-10-11
fix(host-cleanup): "permission denied" errors, do not wipe git-patches on every run

distorhead:
* Fix bug related to wiping of old git-patches by cache-version: fixed typo.
* Do not remove git data entries which has been used recently (wait for 3 hours).


distorhead:
Fixes https://github.com/werf/werf/issues/3768

PR #4246, opened 2022-03-14
fix(build): do not store images into final repo when --skip-build is set

distorhead:
~* By default in 'werf converge', 'werf bundle publish', 'werf bundle export', 'werf export', 'werf render' and 'werf run' commands.~
~* In the 'werf build' command only when `--verify-built-images` option specified.~
* Respect --skip-build option when final-repo is used:
    * do not push into final repo built images from the primary repo;
    * only check stage image existance in the final repo in such case.


distorhead:
Removed verify images in container registry functionality, and only saved a fix to final repo + skip build logic:

> fix(build): do not store images into final repo when --skip-build is set

PR #4748, opened 2022-08-10
chore(telemetry): update projectID dependencies

alexey-igrychev:


cmd/werf/common/telemetry.go
```diff
@@ -120,21 +120,41 @@ func getTelemetryProjectID(ctx context.Context) (string, error) {
...
+	ep, err := transport.NewEndpoint(url)
+	if err != nil {
+		return "", fmt.Errorf("bad repo origin url %q: %w", url, err)
+	}
+
+	var formatPath string
+	{
+		formatPath = ep.Path
+
+		paramsIndex := strings.Index(formatPath, "?")
```
ilya-lesikov, line 145: Could there be the `?` symbol somewhere before the GET parameters? I would've Split the `ep.Path` by `/` and would've checked only the last part for `?` occurence.

alexey-igrychev: The parameters can include `/` (e.g. `?path=dir/file.txt`) . And I am not sure about the possibility of using `?` in the rest of path.

PR #5144, opened 2022-12-12
Update API. Closes #5143

### LINKED ISSUE #5143: Werf for Selectel CRaaS version 3
### Before proceeding

### Problem

Api methods was updated before CRaaS public release. We need to update our code.

### Solution (if you have one)

Update API methods

### Additional information

[CRaaS API Docs](https://developers.selectel.ru/docs/selectel-cloud-platform/craas_api/)

gyrter:
API Update for relesed Selectel CRaaS


alexey-igrychev:
- Added _full log_ does not include werf log. Please, re-add it. Nonetheless, I do not see any problem with rejected images (a special tag that werf add in the case when the related image is broken), and cleanup should end without any error — it is a valid case.
- We restructured documentation, and now you can find the relevant section [here](https://github.com/werf/werf/blob/main/docs/pages_ru/usage/cleanup/cr_cleanup.md#selectel-craas).
- _OFFTOP_. What do you need for alternative Tags method implementation for the Selectel container registry? It seems you can fully remove it, and everything will work properly. https://github.com/werf/werf/blob/1aa250a6bc342f57e6522cd950693cd082004095/pkg/docker_registry/selectel.go#L94-L100

PR #5568, opened 2023-04-26
fix: speedup docs rendering

diafour:
- Fix werf compose up for "real time" development.
- Add jekyll_base image (as in werf/website).
- Remove assets.precompile - lead to errors with updated Jekyll and sprockets.
- Use include-cached plugin to render sidebar.
- Change js that activate sidebar item.


diafour:
TODO fix js: expand first level items on page load.

z9r5 (approved):

PR #5658, opened 2023-05-26
fix: hide build log in export command

diafour:
- always print grouping frame for images
- move log catching logic into Conveyor
- add DeferBuildLog to ConveyorOptions
- refactor ExportPhase into standalone Exporter
- print log using logboek OutStream instead of fmt.Print


distorhead:
LGTM.

PR #5669, opened 2023-05-29
fix: integration tests should not fail

diafour:
- fix construction of the image name with the digest for default implementation


diafour:
AWS API uses sha:256 for digest: https://docs.aws.amazon.com/AmazonECR/latest/APIReference/API_BatchDeleteImage.html

Same is Selectel: https://developers.selectel.ru/docs/selectel-cloud-platform/craas_api/

diafour:
test:e2e / daily is green now

PR #5814, opened 2023-10-04
Werf helm template consumes "Chart.yaml file is missing" error

BrotifyPacha:
Calling `helm template` in an empty directory through werf results in empty string and calling it directly returns an error (as seen in image below).

image

It happens due to the fact that werf uses chart_extender to modify original chart metadata, and `AutosetChartMetadata` function doesn't check whether incoming metadata is nil or not.

This PR emulates behavior of the original `helm template` command.

image


ilya-lesikov:
Thanks!

ilya-lesikov:
Looks like it broke all other non-helm commands (werf converge, ...):
```
  Release "***-test-converge-24-912f565e-test" does not exist. Installing it now.
  Running time 1.57 seconds
  Error: helm upgrade have failed: Chart.yaml file is missing
```

I guess it was like that because default chart in `.helm` does not need Chart.yaml to be present for all commands except `werf helm ...`, since commands like `werf converge` dynamically generate Chart.yaml if it's missing. I'll revert this change, `werf converge` must not break.

BrotifyPacha:
Okay, got you, will try to come up with a better solution

PR #6306, opened 2024-08-08
chore(ci): enable docs unit test

alexey-igrychev:


alexey-igrychev:
fixed here https://github.com/werf/werf/pull/6307

Issue #6315, opened 2024-08-19
Panic building multi-platform images


jarojasm95:
### Before proceeding

- [X] I didn't find a similar [issue](https://github.com/werf/werf/issues)

### Version

2.9.1

### How to reproduce

I haven't bee able to reproduce consistently but given the stack trace it seems specific to multiplatform images and rebuilding.

This seems related to this change https://github.com/werf/werf/pull/5618/files#diff-ea9b317bcabdca97716079f76aa2a660f8097031e01d4798ee5bb6e7dd3cb231R483

It seems that previously it could return an empty list which would avoid the null pointer exception, now it always expects a value even though `GetMultiplatformImage` can potentially return `null`

### Result

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x18 pc=0x34575d5]

goroutine 1 [running]:
github.com/werf/werf/v2/pkg/build.(*BuildPhase).createReport(0xc000bfa6e0, {0x50b83f0, 0xc000c96330})
        /git/pkg/build/build_phase.go:493 +0xbf5
github.com/werf/werf/v2/pkg/build.(*BuildPhase).AfterImages(0xc000bfa6e0, {0x50b83f0, 0xc000c96330})
        /git/pkg/build/build_phase.go:300 +0x2c5
github.com/werf/werf/v2/pkg/build.(*Conveyor).runPhases.func1()
        /git/pkg/build/conveyor.go:586 +0x30
github.com/werf/logboek/internal/stream.(*LogProcess).DoError(0xc000fe2440, 0xc0015a53c0)
        /go/pkg/mod/github.com/werf/logboek@v0.6.1/internal/stream/process_types.go:195 +0xbb
github.com/werf/werf/v2/pkg/build.(*Conveyor).runPhases(0xc000cbe6e0?, {0x50b83f0?, 0xc000c96330}, {0xc001582610, 0x1, 0x1}, 0x0?)
        /git/pkg/build/conveyor.go:585 +0x47f
github.com/werf/werf/v2/pkg/build.(*Conveyor).Build(0xc000cbe6e0, {0x50b83f0, 0xc000c96330}, {{{0x0, 0x0}, 0x0, 0x0}, {{0x0, 0x0, 0x0}}, ...})
        /git/pkg/build/conveyor.go:539 +0x25e
github.com/werf/werf/v2/cmd/werf/export.run.func2(0xc000cbe6e0)
[...]

PR #6316, opened 2024-08-19
Work around null pointer exception on reports for multi-platform images

jarojasm95:
Tries to fix: https://github.com/werf/werf/issues/6315

Although it does not address the root cause (unknown to me atm) this at least prevents a secondary function (like reporting) to break the primary functionality of building images


jarojasm95:
closing in favor of https://github.com/werf/werf/pull/6317

PR #6317, opened 2024-08-20
fix(build/stages): Add RWMutex to ImagesTree

jarojasm95:
Add RWMutex to ImagesTree to synchronize concurrent access of the "multiplatformImages" array. This is done to prevent race conditions where an image is not stored in the array because it is overwritten by a parallel task.

Closes https://github.com/werf/werf/issues/6315

### DIFF
--- a/pkg/build/image/image_tree.go
+++ b/pkg/build/image/image_tree.go
@@ -7,6 +7,7 @@ import (
 	"path/filepath"
 	"reflect"
 	"sort"
+	"sync"
 
 	"github.com/werf/logboek"
 	"github.com/werf/logboek/pkg/types"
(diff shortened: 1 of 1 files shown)


pkg/build/image/image_tree.go
```diff
@@ -203,6 +208,8 @@ func (tree *ImagesTree) GetMultiplatformImage(name string) *MultiplatformImage {
 }
 
 func (tree *ImagesTree) SetMultiplatformImage(newImg *MultiplatformImage) {
```
jarojasm95, line 210: this appears to be called concurrently here: https://github.com/werf/werf/blob/main/pkg/build/build_phase.go#L268

which I believe is causing a race condition that causes images to be missing in the `multiplatformImages` - adding a mutex should synchronize access to the array and prevent the race

alexey-igrychev:
You are right. Thank you for your contribution! ❤️

PR #6365, opened 2024-10-23
fix(cleanup, aws): support sso

alexey-igrychev:


shurup:
Related GitHub discussion: #6357

PR #6367, opened 2024-10-24
feat(release): distribute werf images for linux/arm64 plarform

Fral738:


scripts/werf-in-image/werf.yaml
```diff
@@ -51,12 +24,17 @@ dockerfile: "./{{ $distro }}.Dockerfile"
...
-  add: "/usr/local/bin/werf-{{ $group }}-{{ $channel }}"
-  to: /usr/local/bin/werf
+  imports:
+    - type: ImageName
+      targetBuildArg: source_image
+
+args:
+  distro: "{{ $distro }}"
+  source: "/usr/local/bin/werf-{{ $group }}-{{ $channel }}"
+  dest: "/usr/local/bin/werf"
```
alexey-igrychev, line 37: Let's use the same logic for naming:
- source_image, distro_image
- src, dest or source, destination

scripts/werf-in-image/trdl.Dockerfile
```diff
@@ -0,0 +1,26 @@
...
+    wget https://github.com/mikefarah/yq/releases/download/v4.24.2/yq_linux_amd64 -O /usr/local/bin/yq && \
+    chmod +x /usr/local/bin/yq
+
+COPY trdl_channels.yaml /trdl_channels.yaml
+
+RUN for group in ${groups}; do \
+        for channel in ${channels}; do \
+            echo "Perform trdl update for werf $group $channel ..." &&\
+            trdl update werf $group $channel && \
+            . $(trdl use werf $group $channel) && \
```
alexey-igrychev, line 23: The current implementation needs to check that we get the expected werf version for the channel (based on the `trdl_channels.yaml` file from repo) as before.

Issue #3028, opened 2020-12-15
[v1.2, dockerfile] BuildKit is not supported


alexey-igrychev:
Do not use BuildKit until:
* The following issues with [BuildKit build](https://github.com/docker/cli/tree/40ec81a79afb993d3f77042194ad81a2a8b466a7/cli/command/image/build_buildkit.go) are not resolved:
  * BuildKit implementation uses standard streams instead of [docker client instance ones](https://github.com/docker/cli/tree/40ec81a79afb993d3f77042194ad81a2a8b466a7/cli/command/cli.go#L51-L53).
  * Reading context/dockerfile from the docker client stdin does not work (fix [here](https://github.com/docker/cli/pull/2885)).
  * The command output is written to the standard output and error streams, and we cannot process it.
* The use of the docker build command is not rewritten to the docker API.

distorhead:
https://github.com/werf/werf/pull/3367

PR #6368, opened 2024-10-24
fix(build,docker): remove quiet mode when using buildkit 

iapershin:


pkg/docker/image.go
```diff
@@ -213,11 +213,8 @@ func CliBuild_LiveOutputWithCustomIn(ctx context.Context, rc io.ReadCloser, args
 	if useBuildx {
 		buildOpts.EnableBuildx = true
 
-		// disable buildkit output in background tasks due to https://github.com/docker/cli/issues/2889
-		// there is no true way to get output, because buildkit uses the standard output and error streams instead of defined ones in the cli instance
 		if ctx.Value(parallelConstant.CtxBackgroundTaskIDKey) != nil {
-			logboek.Context(ctx).Warn().LogLn("WARNING: BuildKit output in background tasks is not supported (--quiet) due to https://github.com/docker/cli/issues/2889")
-			args = append(args, "--quiet")
+			args = append(args, "--progress", "plain") // for displaying logs correctly
```
alexey-igrychev, line 217: We no longer need this block since Buildx always uses plain progress.

We can pass stdout for the foreground task, but it is unnecessary. Let's keep the same behaviour for all tasks.

alexey-igrychev:
rel https://github.com/werf/werf/issues/3028

pkg/docker/image.go
iapershin: done

PR #6387, opened 2024-11-05
test(e2e, export): add export e2e test

Fral738:


test/e2e/export/simple_test.go
```diff
@@ -0,0 +1,60 @@
...
+				By("state1: preparing test repo")
+				SuiteData.InitTestRepo(repoDirname, fixtureRelPath)
+
+				By("state2: running export multiplatform image")
+				werfProject := werf.NewProject(SuiteData.WerfBinPath, SuiteData.GetTestRepoPath(repoDirname))
+				SuiteData.Stubs.SetEnv("WERF_REPO", SuiteData.WerfRepo)
+				exportOut := werfProject.Export(&werf.ExportOptions{
+					CommonOptions: werf.CommonOptions{
+						ExtraArgs: []string{
+							"--dev",
```
alexey-igrychev, line 51: The --dev option should not used in CI/CD pipelines and tests, only for local development.

test/e2e/export/simple_test.go
```diff
@@ -0,0 +1,60 @@
...
+							"--tag",
+							SuiteData.WerfRepo,
+						},
+					},
+				})
+				Expect(exportOut).To(ContainSubstring("Exporting image..."))
```
alexey-igrychev, line 57: Let's check that the final image is built for specified platforms by checking [index manifest](https://docs.docker.com/build/building/multi-platform/):

Example:
```
$ crane manifest localhost:5001/test/test | jq
{
  "schemaVersion": 2,
  "mediaType": "application/vnd.docker.distribution.manifest.list.v2+json",
  "manifests": [
    {
      "mediaType": "application/vnd.oci.image.manifest.v1+json",
      "size": 412,
      "digest": "sha256:601cd229e281c1637f64902ffec9f7ed3bd1a42514a4dab85e0850731655de8d",
      "platform": {
        "architecture": "arm64",
        "os": "linux"
      }
    },
    {
      "mediaType": "application/vnd.oci.image.manifest.v1+json",
      "size": 412,
      "digest": "sha256:b2bd66ea29c45f1789d46a9a89f9452f15c5415235c689fb2f5cc47dcea8d82b",
      "platform": {
        "architecture": "arm64",
        "os": "linux"
      }
    }
  ]
}
```

You can see [here](https://github.com/werf/werf/blob/49a1540031ed3733b809a348722ccef631553c98/pkg/docker_registry/api.go#L147-L179) how to work with index manifest.

test/e2e/export/simple_test.go
```diff
@@ -0,0 +1,60 @@
...
+	"strings"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+
+	"github.com/werf/werf/v2/test/pkg/werf"
+)
+
+var _ = Describe("Simple export", Label("e2e", "export", "simple"), func() {
+	It("should succeed and export images",
```
alexey-igrychev, line 13: You need to write two separate tests instead of one:
- It should succeed and export image;
- It should succeed and export image (multi-platform).

Two-step fixtures (state0, state1) aren’t necessary in this test; instead, a single directory setup like [this example](https://github.com/werf/werf/blob/aa63823a83c29cf3439db413703345210883bf26/test/e2e/build/_fixtures/cmd_entrypoint) will be sufficient. For the multi-platform test, simply add the `--platform=linux/amd64,linux/arm64` option to achieve the required behavior.

test/e2e/export/simple_test.go
alexey-igrychev: + check absence of werf labels

test/e2e/export/simple_test.go
```diff
@@ -0,0 +1,104 @@
...
+)
+
+type simpleTestOptions struct {
+	ExtraArgs []string
+}
+
+func ManifestGet(imageName string) *v1.IndexManifest {
+	bytes, err := crane.Manifest(imageName)
+	if err != nil {
+		fmt.Errorf("Error %s", err)
```
alexey-igrychev, line 24: It is not handled. Nothing happens if an error occurs.

test/e2e/export/simple_test.go
```diff
@@ -0,0 +1,104 @@
...
+	err = json.Unmarshal(bytes, &data)
+	if err != nil {
+		fmt.Errorf("Error %s", err)
+	}
+	if data.Config.Labels == nil {
+		result = false
+	} else {
+		result = true
+	}
+	return
```
alexey-igrychev, line 49: return len(data.Config.Labels) == 0

test/e2e/export/simple_test.go
```diff
@@ -0,0 +1,104 @@
...
+				manifest := ManifestGet(imageName)
+				for i := range manifest.Manifests {
+					if manifest.Manifests[i].Platform.Architecture == "amd64" {
+						Expect(manifest.Manifests[i].Platform.Architecture).To(Equal("amd64"))
+					} else {
+						if manifest.Manifests[i].Platform.Architecture != "unknown" {
+							Expect(manifest.Manifests[i].Platform.Architecture).To(Equal("arm64"))
+						}
+					}
+				}
```
alexey-igrychev, line 89: This code does not check explicit platforms.

test/e2e/export/simple_test.go
```diff
@@ -0,0 +1,104 @@
...
+	var data v1.IndexManifest
+	err = json.Unmarshal(bytes, &data)
+	if err != nil {
+		fmt.Errorf("Error %s", err)
+	}
+	return &data
+}
+
+func LabelsExist(imageName string) (result bool) {
+	bytes, err := crane.Config(imageName)
```
alexey-igrychev, line 35: This way, you check the image config only for one platform.

Issue #6319, opened 2024-08-22
Feature enabling verbose for `werf kubectl get` cli command


smilingDima:
### Before proceeding

- [X] I didn't find a similar [issue](https://github.com/werf/werf/issues)

### Problem

I have problem connecting gitlab to k8s:
`werf coverage`

> Error: release deploy: construct kube client factory: Kubernetes cluster unreachable: error getting kubernetes server version: Get "http://localhost:8080/version": dial tcp [::1]:8080: connect: connection refused 

I'm trying to view more information and enable verbose output.
But `$ werf kubectl get nodes -v=10` returns:

> Error: unknown shorthand flag: 'v' in -v=10
> See 'werf kubectl get --help' for usage.

### Solution (if you have one)

Adding permission to pass -v flag.

ilya-lesikov:
Fixed by https://github.com/werf/werf/pull/6415

PR #6415, opened 2024-11-14
fix(kubectl): add klog arbitrary flags

Drey:
`cmd.NewDefaultKubectlCommandWithArgs()` does not inject log arbitrary flags required by args like in the `werf kubectl get nodes -v=10`.

I had a look on  how `kubectl` is originally run and found `logs.AddFlags(kubectlCmd.PersistentFlags())` call here - https://github.com/kubernetes/component-base/blob/55c45bc78189d14353588befc15715b156914ce7/cli/run.go#L85C1-L118C1

Fix bug #6319

### DIFF
--- a/cmd/werf/kubectl/kubectl.go
+++ b/cmd/werf/kubectl/kubectl.go
@@ -8,6 +8,7 @@ import (
 
 	"github.com/spf13/cobra"
 	"k8s.io/cli-runtime/pkg/genericclioptions"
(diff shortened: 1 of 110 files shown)


alexey-igrychev:
Thank you for your attention. Fix the failed job, and I'll merge your contribution.
You can do that by running `task doc:gen` and committing changes.

Drey:
@alexey-igrychev good catch! Done 👍

Issue #4957, opened 2022-09-22
Custom docker config does not work with buildah container backend


alexey-igrychev:
### Before proceeding

- [X] I didn't find a similar [issue](https://github.com/werf/werf/issues)

### Version

v1.2.174

### How to reproduce

```
werf cr login -u $USERNAME -p $PASSWORD $WERF_REPO --docker-config=/tmp/test
werf render --docker-config=/tmp/test
```

### Result

```
Error: UNAUTHORIZED: unauthorized to access repository
```

### Expected result

Command runs without error.

alexey-igrychev:
fixed by https://github.com/werf/werf/issues/4957

PR #6421, opened 2024-11-15
fix(buildah): set docker config for buildah backend

drey:
Fix bug #4957


alexey-igrychev:
Three in a row! Your contributions are really making a difference, and we appreciate it so much. 

Thank you! ❤️

drey:
Thank you for your feedback! It's a win-win case. The project becomes even better and I have a chance to improve my golang skills :))

PR #6549, opened 2025-01-17
feat(build): improve 429 error handling

nervgh:


pkg/docker_registry/auth/auth.go
```diff
@@ -0,0 +1,65 @@
...
+		authConfig.IdentityToken = options.Password
+	} else {
+		authConfig.Password = options.Password
+	}
+
+	operation := func() (string, error) {
+		_, token, err := remote.Auth(ctx, authConfig, options.UserAgent)
+		if err != nil && strings.Contains(err.Error(), "failed with status: 429") {
+			return "", err
+		}
```
alexey-igrychev, line 52: This code ignores non-429 errors. Is it expected or a typo?

pkg/docker_registry/auth/auth.go
```diff
@@ -0,0 +1,65 @@
...
+		return token, nil
```
alexey-igrychev, line 53: This code ignores non-429 errors. Is it expected or a typo?

PR #6590, opened 2025-02-03
feat(config): add .Files.Exists and .Files.IsDir Go template functions

nervgh:


pkg/config/parser.go
alexey-igrychev, line 1: Look at how `doGlob` is implemented and cover this detail: https://werf.io/docs/v2/reference/werf_giterminism_yaml.html#L8.

docs/pages_en/reference/werf_yaml_template_engine.md
```diff
@@ -223,6 +223,38 @@ __Syntax__:
...
+__Syntax__:
+{% raw %}
+```yaml
+{{ .Files.Exists "<FILE_PATH>" }}
+```
+{% endraw %}
+
+> By default, the use of files that have non-committed changes is not allowed by giterminism (read more about it [here]({{ "usage/project_configuration/giterminism.html" | true_relative_url }}))
+
+##### Example: how to create a dir in stapel image depending on existence of a file
```
alexey-igrychev, line 239: We should think about something more realistic or add it without an example.

alexey-igrychev (changes requested):

alexey-igrychev (commented):
A user can use these functions with committed files without extra actions. If the files are not committed, the user must add their paths to `werf-giterminism.yaml`.

The current version doesn’t work this way:
```
➜ werf config render           

Error: template: werfConfig:39:9: executing "werfConfig" at <.Files.Exists>: error calling Exists: {{ .Files.Exists "werf.yaml" }}: the file "werf.yaml" not accepted by giterminism config

➜ git status
On branch docs/installation/update-deckhouse-cli-version
Your branch is ahead of 'origin/docs/installation/update-deckhouse-cli-version' by 2 commits.
  (use "git push" to publish your local commits)

nothing to commit, working tree clean

➜ ls -la
total 88
drwxr-xr-x  17 aigrychev  staff   544 Feb  5 22:04 .
drwxr-xr-x  55 aigrychev  staff  1760 Feb  4 13:55 ..
-rw-r--r--@  1 aigrychev  staff  6148 Jan 11 18:06 .DS_Store
drwxr-xr-x  16 aigrychev  staff   512 Feb  5 22:04 .git
-rw-r--r--   1 aigrychev  staff     5 Nov 14 12:49 .gitignore
-rw-r--r--   1 aigrychev  staff  1245 Dec 20 14:10 .gitlab-ci.yml
-rw-r--r--   1 aigrychev  staff    91 Nov 14 12:49 .helmignore
drwxr-xr-x@  5 aigrychev  staff   160 Jan 22 13:15 .idea
drwxr-xr-x   7 aigrychev  staff   224 Feb  1 02:56 .werf
-rw-r--r--@  1 aigrychev  staff    34 Dec 18 15:08 Chart.yaml
-rw-r--r--   1 aigrychev  staff    30 Feb  1 03:00 Dockerfile
-rw-r--r--   1 aigrychev  staff    50 Dec 20 14:24 README.md
drwxr-xr-x  11 aigrychev  staff   352 Jan 22 13:19 docs
-rwxr-xr-x   1 aigrychev  staff   181 Nov 14 12:49 enabled
drwxr-xr-x   4 aigrychev  staff   128 Nov 14 12:49 openapi
-rw-r--r--   1 aigrychev  staff   184 Feb  5 22:04 werf-giterminism.yaml
-rw-r--r--   1 aigrychev  staff   606 Feb  5 21:57 werf.yaml
```

alexey-igrychev (changes requested):
Looks good, but the parameter name `isFileExisted` should be changed to something like `isFileExist`.

alexey-igrychev (changes requested):
`.Files.IsDir` currently returns `true` for regular files, e.g., `{{ .Files.IsDir "werf.yaml" }}`.

I think the behavior should be as follows:

- **.Files.Exists**:  
  - Returns `false` if the file does not exist in the Git repository (either missing or not committed).  
  - Returns `true` if the file exists in the Git repository or is accepted by Giterminism.  
- **.Files.IsDir**:  
  - Should follow the same logic as `.Files.Exists`, but additionally verify whether the file is a directory.

PR #6598, opened 2025-02-03
refactor(build): refactor secrets validation error handling

iapershin:


pkg/config/raw_secrets.go
```diff
@@ -4,40 +4,48 @@ import (
 	"fmt"
 )
 
+// TODO (iapershin)
+// rawOrigin is not suitable here since image stack contains `doc` property and coudn't be used along with doc()
```
alexey-igrychev, line 8:
```suggestion
// rawOrigin is not suitable here since image stack contains `doc` property and couldn't be used along with doc()
```

pkg/config/raw_secrets.go
```diff
@@ -51,6 +59,6 @@ func (s *rawSecret) toDirective() (Secret, error) {
 	case s.PlainValue != "":
 		return newSecretFromPlainValue(s)
 	default:
-		return nil, newDetailedConfigError("secret type is not supported", s, s.doc)
+		return nil, newDetailedConfigError("unknown secret type. only `env`, `src` and `value` is supported", s, s.parent.getDoc())
```
alexey-igrychev, line 62:
```suggestion
		return nil, newDetailedConfigError("secret should be defined as `env`, `src` or `value`", s, s.parent.getDoc())
```

pkg/config/secrets.go
```diff
@@ -140,14 +140,14 @@ func GetValidatedSecrets(rawSecrets []*rawSecret, giterminismManager giterminism
...
-			return nil, err
+			return nil, newDetailedConfigError(fmt.Sprintf("unable to load build secrets: %s", err.Error()), s, s.parent.getDoc())
 		}
 
 		secretId := secret.GetSecretId()
 		if _, ok := secretIds[secretId]; !ok {
 			secretIds[secretId] = struct{}{}
 		} else {
-			return nil, newDetailedConfigError("duplicated secret %s", secretId, s.doc)
+			return nil, newDetailedConfigError("duplicated secret %s", secretId, s.parent.getDoc())
```
alexey-igrychev, line 150:
```suggestion
			return nil, newDetailedConfigError(fmt.Sprintf("duplicated secret %q", secretId), nil, s.parent.getDoc())
```

alexey-igrychev (changes requested):

pkg/config/raw_secrets.go
iapershin: done

pkg/config/raw_secrets.go
iapershin: done

pkg/config/secrets.go
iapershin: done

PR #6620, opened 2025-02-07
fix(build): improve http retry after header handling logic

nervgh:


pkg/docker_registry/transport/backoff_http_retry_after_handler.go
```diff
@@ -0,0 +1,65 @@
...
+	// ------------------
+	// IMPORTANT: This is constraint with uint8 (bitSize).
+	// It means max=255, what is 255/60 = 4.25 min or 4 min 15 sec
+	seconds, err := strconv.ParseUint(headerValue, 10, 8)
+	// Consider str_conv err as permanent one
+	if err != nil {
+		return nil, backoff.Permanent(err)
+	}
+
+	return nil, backoff.RetryAfter(int(seconds))
```
alexey-igrychev, line 64:
```suggestion
// ------------------
// Try parsing as <http-date> (RFC1123 format).
// Example: 'Tue, 29 Oct 2024 16:56:32 GMT'.
// ------------------
if retryAfterTime, err := time.Parse(time.RFC1123, headerValue); err == nil {
	seconds := time.Until(retryAfterTime).Seconds()
	return nil, backoff.RetryAfter(int(math.Max(0, seconds)))
}

// ------------------
// Try parsing as <delay-seconds>.
// IMPORTANT: Constraint with uint8 (bitSize) → max=255 (4 min 15 sec).
seconds, err := strconv.ParseUint(headerValue, 10, 8)
// ------------------
if err != nil {
	return nil, backoff.Permanent(err) // Consider parsing error as permanent.
}

return nil, backoff.RetryAfter(int(seconds))
```

alexey-igrychev: fyi code uses `math.Max(0, seconds)` to prevent negative retry times

pkg/docker_registry/transport/backoff_http_retry_after_handler_test.go
```diff
@@ -0,0 +1,69 @@
...
+	"net/http/httptest"
+	"strconv"
+	"time"
+
+	"github.com/cenkalti/backoff/v5"
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+)
+
+var _ = Describe("backoff_http_retry_after_handler", func() {
```
alexey-igrychev, line 15: What do you think about using `DescribeTable` instead of this series of `It`?

alexey-igrychev (changes requested):

PR #6643, opened 2025-02-17
feat(build): extend werf.yaml with `build.cacheVersion` and `<image>.cacheVersion` to help manage the cache

nervgh:


pkg/build/stage/from.go
```diff
@@ -42,20 +42,24 @@ func newFromStage(fromImageOrArtifactImageName, baseImageRepoIdOrNone, cacheVers
...
+	fromImageOrArtifactImageName string
 }
 
 func (s *FromStage) HasPrevStage() bool {
 	return false
 }
 
 func (s *FromStage) GetDependencies(ctx context.Context, c Conveyor, cb container_backend.ContainerBackend, prevImage, prevBuiltImage *StageImage, buildContextArchive container_backend.BuildContextArchiver) (string, error) {
-	var args []string
+	args := make([]string, 0, len(s.configMounts)+4)
```
alexey-igrychev, line 55: Should we explicitly set the capacity in this case if 4 is a magic number that won’t adapt to code changes?

pkg/build/stage/instruction/add.go
alexey-igrychev, line 1: For staged Dockerfile, you should add this dependency only to the FROM instruction (similar to the Stapel from stage), as that should be sufficient.

alexey-igrychev (changes requested):
✊🏼

pkg/build/stage/from.go
nervgh: Oops... I'd done the change thinking about code optimization but I forget to roll it back. I agree with you. I think we should use previous code version.

docs/_data/werf_yaml.yml
```diff
@@ -23,6 +23,11 @@ sections:
         collapsible: true
         isCollapsedByDefault: false
         directives:
+          - name: cacheVersion
+            description:
+              en: A static parameter used when forming tags for all image configurations.
+              ru: Статический параметр используемый при формировании тегов для всех конфигураций образов.
+            value: "string"
```
alexey-igrychev, line 30:
```suggestion
          - &image-section-cache-version 
            name: cacheVersion
            description:
              en: A static parameter used when forming tags for all image configurations.
              ru: Статический параметр используемый при формировании тегов для всех конфигураций образов.
            value: "string"
```

docs/_data/werf_yaml.yml
```diff
@@ -268,6 +273,12 @@ sections:
         description:
           en: "Enable layer-by-layer caching of Dockerfile instructions in container registry"
           ru: "Включить послойное кеширование Dockerfile-инструкций в container registry"
+      - &image-section-cache-version
+        name: cacheVersion
+        description:
+          en: A static parameter used when forming tags for all image configurations. Overrides the global build.cacheVersion.
+          ru: Статический параметр используемый при формировании тегов для всех конфигураций образов. Переопределяет глобальный build.cacheVersion.
+          value: "string"
```
alexey-igrychev, line 281:
```suggestion
      - << : *image-section-cache-version
```

docs/_data/werf_yaml.yml
```diff
@@ -505,6 +516,7 @@ sections:
         detailsArticle:
           all: "/usage/build/stapel/imports.html"
       - << : *dockerfile-image-section-final
+      - << : *image-section-cache-version
```
alexey-igrychev, line 519:
```suggestion
      - << : *meta-section-build-cache-version
```

alexey-igrychev (approved):

PR #6644, opened 2025-02-17
fix(build, buildah, dockerfile, staged): fix <bin>: no found (invalid PATH)

iapershin:


alexey-igrychev (changes requested):
Looks good! Let’s do a small refactoring by removing unused code.

Let’s remove this method from the interface along with its implementation (https://github.com/werf/werf/blob/36e5c37c1dc1bf51e95c9160ad7d493cc72f52b8/pkg/storage/manager/storage_manager.go#L71), since this code is never executed:
https://github.com/werf/werf/blob/36e5c37c1dc1bf51e95c9160ad7d493cc72f52b8/pkg/storage/manager/storage_manager.go#L369.

pkg/storage/manager/storage_manager.go
```diff
@@ -10,7 +10,7 @@ import (
 
 	"gopkg.in/yaml.v2"
 
-	"github.com/werf/common-go/pkg/lock"
+	chart "github.com/werf/common-go/pkg/lock"
```
alexey-igrychev, line 13: chart?

PR #6708, opened 2025-03-17
feat(cleanup): add concurrency to git history based cleanup

iapershin:


pkg/cleaning/stage_manager/stages.go
```diff
@@ -1,12 +1,15 @@
...
 import (
+	"sync"
+
 	"github.com/werf/werf/v2/pkg/image"
 )
 
 type managedStageDescSet struct {
 	stageDescSet     image.StageDescSet
 	stageDescMetaMap map[*image.StageDesc]*stageMeta
+	mu               sync.Mutex
```
nervgh, line 12: Official Go's documentation [says](https://pkg.go.dev/sync#Mutex): 

> A Mutex must not be copied after first use.

However we have the type `managedStageDescSet` and the constructor [newManagedStageDescSet](https://github.com/werf/werf/blob/8e4867095a2a37063ccbdc5ae5e1f817279028ed/pkg/cleaning/stage_manager/stages.go#L46) which returns the value (not a pointer). So we have the risk of copying of mutex. To mitigate the risk we could use the pointer to mutex.

Maybe we have any linters which warn us about coping of mutxes. I'm not sure about this type of protection.

iapershin: There are no `go vet` warnings because `managedStageDescSet` methods have pointer receivers then there's no copying of mutex. 
Anyway I see no problem make  [newManagedStageDescSet](https://github.com/werf/werf/blob/8e4867095a2a37063ccbdc5ae5e1f817279028ed/pkg/cleaning/stage_manager/stages.go#L46) to return pointers

pkg/cleaning/git_history_based_cleanup/repo_with_cache.go
```diff
@@ -0,0 +1,67 @@
...
+// It doesn't actually protect any data and works like a semaphore.
+// Locking particular commit only is not safe because of inner go-git logic.
+// We need this workaround because go-git is not thread-safe. It should be fixed in go-git v6.
+// Ref: https://github.com/go-git/go-git/issues/773
+func (g *GitRepositoryWithCache) CommitObject(commitHash plumbing.Hash) (*object.Commit, error) {
+	if c, ok := g.getFromCache(commitHash); ok {
+		return c, nil
+	}
+	g.mu.Lock()
+	defer g.mu.Unlock()
```
alexey-igrychev, line 43:
```suggestion
	if c, ok := g.getFromCache(commitHash); ok {
		return c, nil
	}
	
	g.mu.Lock()
	defer g.mu.Unlock()
	
	if c, ok := g.getFromCache(commitHash); ok {
		return c, nil
	}
```

alexey-igrychev: Double-check the cache after acquiring the lock to avoid redundant Git operations.

pkg/cleaning/git_history_based_cleanup/repo_with_cache.go
```diff
@@ -0,0 +1,72 @@
...
+	
+	g.mu.Lock()
+	defer g.mu.Unlock()
+	
+	if c, ok := g.getFromCache(commitHash); ok {
+		return c, nil
+	}
+	c, err := g.GitRepo.CommitObject(commitHash)
+	if err != nil {
+		return nil, fmt.Errorf("unable to get commit object for %s: %s", commitHash.String(), err)
```
alexey-igrychev, line 51:
```suggestion
		return nil, fmt.Errorf("unable to get commit object for %s: %w", commitHash.String(), err)
```

alexey-igrychev (approved):

Issue #6667, opened 2025-02-27
werf cr login does not work with the docker registry image


chasinglogic:
### Before proceeding

- [x] I didn't find a similar [issue](https://github.com/werf/werf/issues)

### Version

2.29.0

### How to reproduce

1. On my local laptops I setup the docker registry with this command:
```
docker run -d   -p 5000:5000   --name registry   -v registry-data:/var/lib/registry   --restart always   registry:2
```
2. I am able to push images to this registry using docker push:
```
(kube: docker-desktop) ~/tmp/werf-playground main λ docker pull alpine
Using default tag: latest
latest: Pulling from library/alpine
6e771e15690e: Download complete
Digest: sha256:a8560b36e8b8210634f77d9f7f9efd7ffa463e380b75e2e74aff4511df3ef88c
Status: Downloaded newer image for alpine:latest
docker.io/library/alpine:latest

What's next:
    View a summary of image vulnerabilities and recommendations → docker scout quickview alpine
(kube: docker-desktop) ~/tmp/werf-playground main λ docker tag alpine localhost:5000/alpine
(kube: docker-desktop) ~/tmp/werf-playground main λ docker push localhost:5000/alpine
Using default tag: latest
The push refers to repository [localhost:5000/alpine]
6e771e15690e: Pushed
latest: digest: sha256:757d680068d77be46fd1ea20fb21db16f150468c5e7079a08a2e4705aec096ac size: 1025

i Info → Not all multiplatform-content is present and only the available single-platform image was pushed
         sha256:a8560b36e8b8210634f77d9f7f9efd7ffa463e380b75e2e74aff4511df3ef88c -> sha256:757d680068d77be46fd1ea20fb21db16f150468c5e7079a08a2e4705aec096ac
(kube: docker-desktop) ~/tmp/werf-playground main λ
```
3. When I run `werf cr login --help` I see this example for insecure registries: 
```
# Login into insecure registry (over http)
werf cr login --insecure-registry registry.example.com
```
4. Run `werf cr login --insecure-registry localhost:5000`

### Result

When I run that I get the error: `Error: provide --username`

If I try running `--insecure-registry` with a bogus `--username` and `--password` flag (`--password` is also required if you simply provide `--username` alone) I get this error:

```
(kube: docker-desktop) ~/tmp/werf-playground main λ werf cr login -u username -p token --insecure-registry localhost:5000
[...]

PR #6711, opened 2025-03-17
docs(cr-login): update example

alexey-igrychev:


alexey-igrychev:
rel https://github.com/werf/werf/issues/6667

PR #6736, opened 2025-03-27
fix(host-cleanup): ensure no parallel processes of auto host cleanup

nervgh:


iapershin (approved):

pkg/background/background.go
```diff
@@ -0,0 +1,27 @@
...
+	chart "github.com/werf/common-go/pkg/lock"
+	"github.com/werf/lockgate"
+)
+
+func IsBackgroundModeEnabled() bool {
+	return os.Getenv("_WERF_BACKGROUND_MODE_ENABLED") == "1"
+}
+
+// TryLock tries background host-lock or panics if error.
+func TryLock() bool {
```
alexey-igrychev, line 15: Why do you use panic instead of handling the error?

cmd/werf/main.go
```diff
@@ -14,10 +14,18 @@ import (
...
 	"github.com/werf/werf/v2/cmd/werf/root"
+	"github.com/werf/werf/v2/pkg/background"
 	"github.com/werf/werf/v2/pkg/process_exterminator"
 )
 
 func main() {
+	// IMPORTANT. In background mode we MUST take host-lock to prevent parallel "werf host cleanup" processes.
+	// The processes write data to the same log files that causes "data race".
+	// We don't need to release the lock manually, because it does automatically when the background process will end.
+	if background.IsBackgroundModeEnabled() && !background.TryLock() {
```
alexey-igrychev, line 25: Was something like that implied?
`!background.IsBackgroundModeEnabled() || !background.TryLock()`

pkg/background/background.go
nervgh: We use this function in background mode only and we are forced to use it BEFORE logger initialization. If we return the error from `TryLock` how should we handle that error?

cmd/werf/main.go
nervgh: No. We do "early exit" only when both conditions are met:

1. werf is running in background mode
2. werf didn't acquire the background mode lock

If we do `!background.Is BackgroundModeEnabled()` werf will be doing "early exit" in foreground mode.

alexey-igrychev (approved):

PR #6765, opened 2025-04-03
refactor(giterminism, build, host-cleanup): update host locker

nervgh:


iapershin (approved):

pkg/background/background.go
```diff
@@ -3,23 +3,27 @@ package background
...
 
 func IsBackgroundModeEnabled() bool {
 	return os.Getenv("_WERF_BACKGROUND_MODE_ENABLED") == "1"
 }
 
 // TryLock tries background host-lock or panics if error.
+//
+// The logger is not available inside of this function
+// because we use the function before initialization of the logger.
 func TryLock() bool {
```
alexey-igrychev, line 18: Why can't we try this lock inside the host cleanup command when the service dir is initialized correctly?

nervgh: I see two relevant comments:

1. We will have correct werf's initialization at this code line after merging #6796
2. We are forced to `TryLock()` here to prevent multiple concurrent writes into the [same static log files](https://github.com/werf/werf/pull/6783/files#diff-11c706d50ab85189c4f54b109452e31984ef860e5ce6c071c54c30ccb76e49b5R39) from detached `werf host cleanup` processes. #6736

PR #6775, opened 2025-04-08
fix(build): use ctx for graceful shutdown

nervgh:


iapershin (changes requested):

cmd/werf/compose/main.go
```diff
@@ -22,6 +22,7 @@ import (
 	"github.com/werf/werf/v2/pkg/giterminism_manager"
 	"github.com/werf/werf/v2/pkg/tmp_manager"
 	"github.com/werf/werf/v2/pkg/true_git"
+	exec2 "github.com/werf/werf/v2/pkg/werf/exec"
```
iapershin, line 25: Needs to be renamed

iapershin (changes requested):

iapershin (approved):

alexey-igrychev (approved):

PR #6776, opened 2025-04-08
test(build, imageSpec): add complex test

iapershin:


nervgh (approved):

integration/suites/build/spec_mutation/suite_test.go
```diff
@@ -51,6 +52,9 @@ var (
 	})
 
 	_ = AfterEach(func() {
+		for _, i := range SuiteData.ImagesToClean {
+			utils.RunSucceedCommand("", "docker", "rmi", i)
+		}
```
alexey-igrychev, line 57: Why do you need this step if the werf host purge command deletes all images?

iapershin: In case of deleting all labels I want to be sure all images created by test will be deleted

alexey-igrychev (approved):

PR #6907, opened 2025-06-17
refactor(build): unify code organization of mutate stages

alexey-igrychev:
Keep mutate logic in stage (MutateImage method).


pkg/docker_registry/api.go
```diff
@@ -263,35 +263,109 @@ func (api *api) deleteImageByReference(ctx context.Context, reference string) er
...
+		newI := i
+		if options.mutateImageLayersFunc != nil {
+			layers, err := i.Layers()
+			if err != nil {
+				return nil, err
+			}
+
+			newLayers, err := options.mutateImageLayersFunc(ctx, layers)
+			if err != nil {
+
```
alexey-igrychev:
```suggestion
return nil, err
```

iapershin (commented):
@alexey-igrychev take a look

PR #6922, opened 2025-06-25
docs(includes): add extended example

iapershin:


docs/examples/includes/app/werf.yaml
```diff
@@ -0,0 +1,15 @@
...
+build:
+  platform:
+    - linux/amd64
+  imageSpec:
+{{ .Files.Get ".werf/meta/imagespec.yaml" | indent 4 }}
+cleanup:
+{{ .Files.Get ".werf/meta/cleanup.yaml" | indent 2 }}
+
+---
+{{ range $path, $content := .Files.Glob "**/werf-partial.yaml" }}
```
ilya-lesikov: I don't think non-templatable partials are that useful. How about we have .werf/_imagespec.tpl, .werf/_cleanup.tpl and .werf/_image.tpl, and in it we have something like:
```
{{ define "cleanup" }}
{{ define "imagespec" }}
{{ define "backend_node" }}
{{ define "frontend_nginx" }}
```
And all of these probably should have dict passed as a context, so they can be parametrized, e. g. `{{ include "cleanup" (dict "main_branch_name" "something") }}`.

And when we implement werf-values.yaml, users can switch to this: `{{ include "cleanup" $.Values.cleanup }}`

docs/examples/includes/app/frontend/index.html
```diff
@@ -0,0 +1,17 @@
...
+<html>
+<head><title>Frontend</title></head>
+<body>
+  <h1>Hello from Frontend</h1>
+  <button onclick="callApi()">Call Backend</button>
+  <p id="output"></p>
+
+  <script>
+    async function callApi() {
+      const res = await fetch('http://backend.typical-app.svc.cluster.local/api/hello');
```
ilya-lesikov: This has to be dynamic somehow. `envsubst` during frontend building will suffice I guess, unless there is a more native way.

docs/examples/includes/common/.werf/meta/cleanup.yaml
```diff
@@ -0,0 +1,16 @@
...
+      limit:
+        last: 20
+        in: 168h
+        operator: And
+    imagesPerReference:
+      last: 1
+      in: 168h
+      operator: And
+  - references:
+      branch: main
```
ilya-lesikov: Templating candidate

docs/examples/includes/common/.werf/meta/imagespec.yaml
```diff
@@ -0,0 +1,4 @@
+clearHistory: true
+config:
+  labels:
+    example.io/app: typical-app
```
ilya-lesikov: Templating candidate

docs/examples/includes/common_js/.helm/charts/backend/templates/deployment.yaml
```diff
@@ -0,0 +1,27 @@
+apiVersion: apps/v1
```
ilya-lesikov: Maybe we just have a single Deployment with two containers: backend and frontend? This will greatly simplify the example. Then we could have only the `docs/examples/includes/common_js/.helm/templates/deployment.yaml` file.

docs/pages_en/usage/project_configuration/includes.md
```diff
@@ -38,64 +38,58 @@ If the same file exists in multiple places, the following rules apply:
...
 
-* In `examples`: `.helm/Chart.yaml`, `.helm/values.yaml`
-* In `helm_examples`: `.helm/Chart.yaml`
+```bash
+.
+├── a      # from the local repository
+├── b      # from repo2
+├── x      # from repo1
+├── y      # from the local repository
+└── z      # from repo2
```
ilya-lesikov: A bit hard to follow, maybe this?
```suggestion
├── a      # from repo1
├── b      # from repo2
├── c      # from local repo
```
Where repo1 has a+b+c, repo2 has b+c, local repo has c.

alexey-igrychev (approved):

PR #6935, opened 2025-06-30
feat(build): add glob patterns support for image names

Fral738:


pkg/config/images_to_process_test.go
```diff
@@ -0,0 +1,180 @@
+package config
+
+import (
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+)
+
+type mockImage struct {
```
nervgh: Please use auto mock generation. For example:

1. [Write generation command as Go comment](https://github.com/werf/werf/pull/6906/files#diff-a7056bbd30d09a760100c35830a6a88f870972a9764aecfdf8c6f0e571e82d82R13)
2. Run `mock:generate` to generate mocks.
3. [Use generated mocks in tests](https://github.com/werf/werf/pull/6906/files#diff-9bf0609f6200f27bb1ed0b2ea4d8dddab48b58ed18fe79cdf3d27be73614c7a4R22).

pkg/config/images_to_process.go
```diff
@@ -10,42 +12,97 @@ type ImagesToProcess struct {
 	WithoutImages      bool
 }
 
+func parsePattern(pattern string) (include, exclude string, isExclusion bool) {
```
nervgh: Could we use external Go library for parsing and matching? For example: https://pkg.go.dev/github.com/go-git/go-git/v5/plumbing/format/gitignore

pkg/config/images_to_process_test.go
Fral738: thx

pkg/config/images_to_process.go
Fral738: I appreciate the suggestion! Let's keep this in mind if we need more advanced pattern matching later. For now, the stdlib solution gives us all needed functionality with better control over allowed syntax.

pkg/config/images_to_process.go
```diff
@@ -10,42 +12,111 @@ type ImagesToProcess struct {
...
+			match, _ := filepath.Match(exclPattern, name)
+			if match {
+				excluded = true
+				break
+			}
+		}
+		if !excluded {
+			finalImages[name] = true
+		}
+	}
```
iapershin: Duplicated code

iapershin (dismissed):

pkg/config/images_to_process.go
Fral738: Thanks for spotting this

pkg/config/images_to_process.go
```diff
@@ -10,42 +13,114 @@ type ImagesToProcess struct {
...
-		var finalImageNameList []string
-		for _, name := range imageNameList {
-			image := werfConfig.GetImage(name)
-			if image == nil {
-				return ImagesToProcess{}, fmt.Errorf("specified images cannot be used: image %q not defined in werf.yaml", name)
+		return ImagesToProcess{WithoutImages: true}, nil
+	}
+
+	allImageNames := werfConfig.GetImageNameList(false)
+	includePatterns := make([]string, 0, len(imageNameList))
```
nervgh: @Fral738 looks fine! Let's improve `NewImagesToProcess` code readability moving some code into functions. Assume `NewImagesToProcess` is the place for main (business) logic. So we want to have something simple like this:

```golang
func NewImagesToProcess(werfConfig *WerfConfig, imageNameList []string, onlyFinal, withoutImages bool) (ImagesToProcess, error) {
   includePatterns, excludePatterns := parsePatterns(...)

    filteredImageNameList := filterUsingIncludePatterns(imageNameList, includePatterns)
    filteredImageNameList = filterUsingExcludePatterns(filteredImageNameList, excludePatterns)

    finalImageNameList := filterUsingFinalAttribute(filteredImageNameList, ...)

    return ImagesToProcess{
		ImageNameList:      filteredImageNameList,
		FinalImageNameList: finalImageNameList,
		WithoutImages:      len(filteredImageNameList) == 0,
	}
}

func filterUsingIncludePatterns () {
 // implement me
}

func filterUsingExcludePatterns () {
 // implement me
}

func filterUsingFinalAttribute () {
   // implement me 
}
```

nervgh (changes requested):

nervgh (approved):

pkg/config/images_to_process.go
```diff
@@ -10,42 +13,148 @@ type ImagesToProcess struct {
...
+func matchPattern(name, pattern string) (bool, error) {
+	match, err := filepath.Match(pattern, name)
+	if err != nil {
+		return false, fmt.Errorf("invalid pattern %q: %v", pattern, err)
+	}
+	return match, nil
+}
+
+func filterUsingIncludePatterns(allImageNames, includePatterns []string) (map[string]bool, error) {
+	finalImages := make(map[string]bool)
```
nervgh, line 61: @Fral738 just a question. Why do we need `finalImages map`? 

For me the pipeline looks like this `all_image_names | include | exclude`. If `include` pattern is empty we pass `all_image_names` to `exclude`.

From it follows that `filterUsingExcludePatterns` should not accept `includePatterns` arg.

Is my vision correct?

Fral738: **Why The Code Works This Way**

**finalImages map**
1. Tracks selected images efficiently

2. Prevents duplicates if multiple patterns match the same image

**Include patterns in exclude step**

1. Ensures explicitly included images (like exact name "app1") aren't removed by broader exclude rules (like "!app*")

PR #6960, opened 2025-07-09
fix(build): avoid inefficient tag listing and cache usage in Docker registry

iapershin:


pkg/container_backend/legacy_base_image.go
```diff
@@ -57,11 +60,15 @@ func (i *legacyBaseImage) UnsetInfo() {
 }
 
 func (i *legacyBaseImage) SetStageDesc(stageDesc *image.StageDesc) {
+	i.mu.Lock()
+	defer i.mu.Unlock()
```
iapershin, line 64: Found data race

pkg/storage/repo_stages_storage.go
```diff
@@ -98,37 +98,38 @@ func (storage *RepoStagesStorage) GetStagesIDs(ctx context.Context, _ string, op
 	var res []image.StageID
 
 	o := makeOptions(opts...)
-	if tags, err := storage.DockerRegistry.Tags(ctx, storage.RepoAddress, o.dockerRegistryOptions...); err != nil {
```
iapershin, line 101: Logic is not changed, only condition statement to simplify the code

pkg/storage/repo_stages_storage.go
```diff
@@ -201,63 +202,64 @@ func (storage *RepoStagesStorage) GetStagesIDsByDigest(ctx context.Context, _, d
...
+	tags, err := storage.DockerRegistry.Tags(ctx, storage.RepoAddress, o.dockerRegistryOptions...)
```
iapershin, line 205: Logic is not changed, only condition statement to simplify the code

PR #6970, opened 2025-07-11
refactor(sbom): log image name while processing

nervgh:


pkg/build/build_phase.go
```diff
@@ -268,7 +268,7 @@ func (phase *BuildPhase) AfterImages(ctx context.Context) error {
 			}
 
 			if img.UseSbom() {
-				if err = phase.sbomStep.Converge(ctx, img.GetStageDesc(), scanner.DefaultSyftScanOptions()); err != nil {
+				if err = phase.sbomStep.Converge(ctx, "", img.GetStageDesc(), scanner.DefaultSyftScanOptions()); err != nil {
```
alexey-igrychev: mmm?

```suggestion
				if err = phase.sbomStep.Converge(ctx, img.Name, img.GetStageDesc(), scanner.DefaultSyftScanOptions()); err != nil {
```

PR #6983, opened 2025-07-16
refactor(host-cleanup): simplify tmp manager

nervgh:


pkg/tmp_manager/gc.go
```diff
@@ -15,306 +17,148 @@ import (
...
+		info, err := dirEntry.Info()
 		if err != nil {
-			return nil, fmt.Errorf("unable to list files in %s: %w", dir, err)
+			return nil, fmt.Errorf("file info for %s: %w", dirEntry.Name(), err)
 		}
 
-		for _, info := range infos {
-			res = append(res, &LinkDesc{FileInfo: info, LinkPath: filepath.Join(dir, info.Name())})
+		// filter out recent files
+		if timeSince(info.ModTime()) < 2*time.Hour {
```
Fral738: Сould we extract the magic number `2 * time.Hour` into a named constant (e.g., `const minFileAge = 2 * time.Hour`)?

pkg/tmp_manager/gc.go
```diff
@@ -15,306 +17,148 @@ import (
...
-			return false, fmt.Errorf("unable to list released projects tmp dirs in %s: %w", releasedProjectsDir, err)
-		}
-
-		if len(releasedProjectsDirs) > 50 {
-			return true, nil
-		}
-	}
-
-	now := time.Now()
+var timeSince = time.Since // for stubbing in tests
```
Fral738, line 24: What about using interface instead of variable `timeSince`?

nervgh: I don't understand clearly. Could you explain a bit more?

alexey-igrychev (approved):

PR #7006, opened 2025-07-28
feat: support basic auth for remote git in includes and stapel git directive

iapershin:


pkg/includes/helpers.go
```diff
@@ -0,0 +1,25 @@
+package includes
+
+import (
+	"path"
+	"path/filepath"
+	"strings"
+)
+
+func sliceContainsSubstring(s string, substrings []string) bool {
```
nervgh, line 9: I think we could replace these function with https://pkg.go.dev/slices#Contains

pkg/git_repo/helpers_test.go
```diff
@@ -0,0 +1,106 @@
+package git_repo
```
nervgh, line 1: The code mostly uses Ginkgo + Gomega for testing. What is the reason to use another setup?

pkg/git_repo/helpers.go
```diff
@@ -17,3 +21,76 @@ func newHash(s string) (plumbing.Hash, error) {
...
+		return nil, fmt.Errorf("no secret specified")
+	}
+
+	return &BasicAuth{
+		Username: cfg.Username,
+		Password: password,
+	}, nil
+}
+
+func oneOrNone(conditions []bool) bool {
```
nervgh, line 80: I think we could use https://github.com/samber/lo?tab=readme-ov-file#everyby to simplify the code.

nervgh (changes requested):

pkg/includes/helpers.go
iapershin: It checks if any element in the slice contains the given **substring**, while slices.Contains() is looking for exact match.

pkg/git_repo/helpers_test.go
iapershin: The reason is to create as simplest unit test as possible for a simple function. G+G is a whole BDD framework which looks like an overkill for such tasks

nervgh: I think the common test tools is the preferred way. This case looks suitable for table testing.

But I'm not sure that blocking the PR because of this is the correct decision.

nervgh (approved):

PR #7008, opened 2025-07-29
feat(build): add worktree pool for git patch

iapershin:


pkg/git_repo/base.go
```diff
@@ -158,18 +156,19 @@ func (repo *Base) GetName() string {
 }
 
 func (repo *Base) getOrCreatePatch(ctx context.Context, repoPath, gitDir, repoID, workTreeCacheDir string, opts PatchOptions) (Patch, error) {
```
alexey-igrychev, line 159: The code is not identical: in the original version, the lock is held until after writing, preventing concurrent patch creation. In the new version, concurrent patch creation is possible for the same patchID, but only one will be stored.

iapershin: https://github.com/werf/werf/blob/6d476ffafbc1698a155af02900bc04897a8bd964/pkg/git_repo/base.go#L162-L164 should do the job

alexey-igrychev (approved):

PR #7021, opened 2025-08-06
fix(build,cache): invalidate local manifest cache on unexpected storage state

Fral738:


pkg/storage/manager/storage_manager.go
```diff
@@ -165,10 +168,10 @@ func (stages *StagesList) AddStageID(stageID image.StageID) {
 }
 
 type StorageManager struct {
-	parallel           bool
-	parallelTasksLimit int
-
-	ProjectName string
+	parallel                  bool
+	parallelTasksLimit        int
+	disableLocalManifestCache atomic.Bool
```
iapershin: I'm not sure about real concurrent access for that field. Anyway I suggest to use mutex to protect StorageManager's state consistency

pkg/storage/manager/storage_manager.go
```diff
@@ -165,10 +167,11 @@ func (stages *StagesList) AddStageID(stageID image.StageID) {
 }
 
 type StorageManager struct {
-	parallel           bool
-	parallelTasksLimit int
-
-	ProjectName string
+	Mux                       sync.Mutex
```
iapershin: don't export

iapershin (changes requested):

PR #7022, opened 2025-08-06
fix: include context.Cause in error handling

Fral738:


pkg/werf/exec/cancel.go
```diff
@@ -27,6 +29,12 @@ func CommandContextCancellation(ctx context.Context, name string, arg ...string)
 	return PrepareGracefulCancellation(exec.CommandContext(ctx, name, arg...))
 }
 
+func TerminateIfCanceled(ctx context.Context, err error, exitCode int) {
```
nervgh: Looks nice! Let's simplify code a little bit: place deriving of causation and exit_code to under the hood of this function.

nervgh: Is it a typo: `canceled` -> `cancelled`?

nervgh (changes requested):

pkg/werf/exec/cancel.go
Fral738: No, it’s not a typo. `Canceled` is the correct American English spelling, while ‘cancelled’ is the British variant. Both are correct, but the American version uses a single `l`.

nervgh (approved):

PR #7039, opened 2025-08-19
fix(build,stapel): fix fromArtifact error

iapershin:


pkg/config/stapel_image_base.go
```diff
@@ -64,6 +64,9 @@ func (c *StapelImageBase) Platform() []string {
 }
 
 func (c *StapelImageBase) GetFrom() string {
+	if c.FromArtifactName != "" {
```
nervgh, line 67: Here we prefer FromArtifactName over From. 

But [there](https://github.com/werf/werf/blob/4824773bada7efca49eeebe23011cea5cfac2e87/pkg/build/stage/from.go#L24) we perfer From over FromArtifactName. 

Is it correct that we have different priorities for those cases?

iapershin: they exclude each other earlier in config validation. this is already deprecated directive and legacy logic that should be completely removed in v3

PR #7045, opened 2025-08-21
fix(host-cleanup): respect host locking for dry-run mode

nervgh:


pkg/host_cleaning/cleanup_report.go
```diff
@@ -7,35 +7,26 @@ import (
 
 type cleanupReport prune.Report
 
-func newCleanupReport() cleanupReport {
+func newCleanupReport(size int) cleanupReport {
```
iapershin: Remove the size parameter, since it’s not known in advance in most cases. Along with it, remove the preallocation of the slice.

pkg/host_cleaning/cleanup_report.go
```diff
@@ -7,35 +7,26 @@ import (
...
 	return cleanupReport{
-		ItemsDeleted:   []string{},
+		ItemsDeleted:   make([]string, 0, size),
 		SpaceReclaimed: 0,
 	}
 }
 
 func mapPruneReportToCleanupReport(report prune.Report) cleanupReport {
-	cr := newCleanupReport()
+	cr := newCleanupReport(0)
```
iapershin: This code is really strange. I think constructor method is unnecessary here. This function is a constructor itself. Just remap fields values to new struct to make logic more clear

pkg/host_cleaning/cleanup_report.go
```diff
@@ -7,35 +7,26 @@ import (
...
 	cr.SpaceReclaimed = report.SpaceReclaimed
 	cr.ItemsDeleted = append(cr.ItemsDeleted, report.ItemsDeleted...)
 	return cr
 }
 
 func mapImageListToCleanupReport(list image.ImagesList) cleanupReport {
-	report := newCleanupReport()
+	report := newCleanupReport(0)
```
iapershin: Same here. Do all necessary logic then just map values to new structure

iapershin (changes requested):

pkg/host_cleaning/cleanup_report.go
```diff
@@ -7,22 +7,22 @@ import (
...
 	return cleanupReport{
-		ItemsDeleted:   make([]string, 0, size),
+		ItemsDeleted:   []string{},
 		SpaceReclaimed: 0,
 	}
 }
 
 func mapPruneReportToCleanupReport(report prune.Report) cleanupReport {
-	cr := newCleanupReport(0)
+	cr := newCleanupReport()
```
iapershin:
```
func mapPruneReportToCleanupReport(report prune.Report) cleanupReport {
	return cleanupReport{
		ItemsDeleted:   report.ItemsDeleted,
		SpaceReclaimed: report.SpaceReclaimed,
	}
}
```

pkg/host_cleaning/cleanup_report.go
```diff
@@ -7,22 +7,22 @@ import (
...
 	cr.SpaceReclaimed = report.SpaceReclaimed
 	cr.ItemsDeleted = append(cr.ItemsDeleted, report.ItemsDeleted...)
 	return cr
 }
 
 func mapImageListToCleanupReport(list image.ImagesList) cleanupReport {
-	report := newCleanupReport(0)
+	report := newCleanupReport()
```
iapershin:
```
func mapImageListToCleanupReport(list image.ImagesList) cleanupReport {
	itemsDeleted := make([]string, 0, len(list))
	var spaceReclaimed uint64

	for _, img := range list {
		itemsDeleted = append(itemsDeleted, img.ID)
		spaceReclaimed += uint64(img.Size)
	}

	return cleanupReport{
		ItemsDeleted:   itemsDeleted,
		SpaceReclaimed: spaceReclaimed,
	}
}
```

pkg/host_cleaning/cleanup_report.go
```diff
@@ -7,22 +7,22 @@ import (
 
 type cleanupReport prune.Report
 
-func newCleanupReport(size int) cleanupReport {
+func newCleanupReport() cleanupReport {
```
iapershin: Why do we even need this constructor? It always return default values

pkg/host_cleaning/local_backend_cleaner.go
```diff
@@ -475,7 +475,7 @@ func (cleaner *LocalBackendCleaner) pruneImages(ctx context.Context, options Run
 	if options.DryRun {
 		list, err := cleaner.backend.Images(ctx, buildImagesOptions(filters.ToPairs()...))
 		if err != nil {
-			return newCleanupReport(0), err
+			return newCleanupReport(), err
```
iapershin: The purpose is to return default value? Maybe just return nil or empty struct? Avoiding unnecessary intermediates makes the code clearer and simpler.

pkg/host_cleaning/cleanup_report.go
nervgh: `newCleanupReport()` constructor is a compromise between efficiency and workability.

Under "efficiency" in our case I talk about optimization via memory allocation when we create a slice of `ItemsDeleted`.

Under "workability" in our case I talk about tests. We have actual and expected reports and we need to compare them. The logic of `cleanupWerfContainers()` is not determined. For example, we could allocate memory for the slice but skip it filling for [some reasons](https://github.com/werf/werf/blob/60a5fdcf85a4b8fdc5597732908d8e0a2ad568f3/pkg/host_cleaning/local_backend_cleaner.go#L531). In the result our comparison will end with the error because `empty slice != nil slice`.

I've handled it using post processing (normalization). I think now it is clearer then it was before.

Thank you!

PR #7049, opened 2025-08-22
fix(host-cleanup, docker): protect "tempID:latest" docker tag from parallel "werf host cleanup" process

nervgh:


pkg/container_backend/docker_server_backend.go
```diff
@@ -24,10 +25,14 @@ import (
 	"github.com/werf/werf/v2/pkg/ssh_agent"
 )
 
-type DockerServerBackend struct{}
+type DockerServerBackend struct {
+	locker lockgate.Locker
+}
```
iapershin, line 30: 1. I don't see any reason why locker is a part of DockerServerBackend. As I see from the code it could be called independently in a single implementation of DockerServerBackend method. You can get host locker from anywhere in the code by werf.HostLocker().Locker(), so I don't think it should be a part of initialization of docker backend.
2. By the way, lockgate.Locker is a very weak abstraction. Ideally, you should declare locker interface on the consumer's side instead importing.

iapershin (changes requested):

pkg/container_backend/docker_server_backend.go
nervgh: I pass the locker as the argument to constructor for these reasons: 

1. Prefer explicit over implicit. If we want to use `werf.HostLocker().Locker()` we are required to call [werf.Init()](https://github.com/werf/werf/blob/main/pkg/werf/main.go#L85) before. In this case it is implicit dependency. 
2. Simplification of testing. If we have the locker as implicit dependency then assertion of `locker.Acquire()` is not so easy as [mocking](https://github.com/werf/werf/pull/7049/files#diff-3e1b0a6f5d8e56cc109fa92ed1df54422f7fc69f9a0aeec5f6c4d8e68b858421R69) using standard approach.

iapershin: 1. It's not about explicit over implicit, but dependency injection.
Ok, left as is, I don't see better approach for this for now
2. It's interface segregation principle - "Clients should not be forced to depend on methods they do not use". 
It makes the code more coupled and rigid. I don't understand why it's not easy for mocking? 
Your mock object implements all the methods and your interface requires only one, so it satisfies the interface.
I'm not blocking the PR but encourage you to review this to make code flexible, easily extensible and minimizing dependencies between modules.

iapershin (approved):

PR #7094, opened 2025-09-24
fix(stapel, git): fix file ownership setup for scratch (rel sudo: unknown uid 0)

Fral738:


pkg/build/stage/git_mapping.go
```diff
@@ -249,6 +250,16 @@ func (gm *GitMapping) applyPatchCommand(patchFile *ContainerFileDescriptor, arch
...
+	if gm.Owner != "" || gm.Group != "" {
+		chownCommand := fmt.Sprintf(
+			"%s -R %s:%s %s",
+			stapel.ChownBinPath(),
+			gm.Owner,
+			gm.Group,
+			applyPatchDirectory,
+		)
+		commands = append(commands, strings.TrimLeft(chownCommand, " "))
+	}
```
alexey-igrychev, line 269: Let's implement the same with chown https://github.com/werf/werf/blob/f37a2bbbb1ce1b9a90c5d768779ea80c6e5b955f/pkg/build/stage/git_mapping.go#L428-L440.

PR #7098, opened 2025-09-28
fix bundle render

loliklol:


ilya-lesikov:
Hello, just saw this. If this was the fix for `bundle apply`/`bundle render` requiring git, then we [fixed](https://github.com/werf/werf/commit/682827485e2302a9b6cf3a466f9c6497fc393429) this in 2.47.10.

Thanks for the effort though. I guess I close this now.

PR #7110, opened 2025-10-01
fix(build): cross validate images' platforms

nervgh:


pkg/config/validator_image_platform.go
```diff
@@ -0,0 +1,62 @@
...
+	// Check dependencies for dockerfile images
+	for _, img := range rawImagesFromDockerfile {
+		for _, dep := range img.RawDependencies {
+			_, rightDiff := lo.Difference(allImagesPlatforms, v.crossJoinImagesPlatforms([]string{dep.Image}, img.Platform))
+
+			if len(rightDiff) > 0 {
+				return v.newImagePlatformError(img.Images[0], rightDiff[0].A, rightDiff[0].B)
+			}
+		}
+	}
```
alexey-igrychev: It also applies to `from`, `fromImage`, `fromArtifact`, and `import.[from|image]`.

PR #7112, opened 2025-10-01
fix(includes): respect flag and env for werf includes update

Fral738:


Fral738:
There are already CLI flags for custom configuration files, e.g., `--config` for `werf.yaml` and `--giterminism-config` for `werf-giterminism.yaml`.
Should we also add a similar flag for specifying a custom path to `werf-includes.yaml`?

PR #7166, opened 2025-10-24
fix(build, stapel): exit with error when import nothing

nervgh:


pkg/build/stage/dependencies.go
```diff
@@ -77,6 +81,8 @@ func (s *DependenciesStage) GetDependencies(ctx context.Context, c Conveyor, cb
 				sourceChecksum, err := s.getImportSourceChecksum(ctx, c, cb, elm)
 				if err != nil {
 					return fmt.Errorf("unable to get import %d source checksum: %w", ind, err)
+				} else if sourceChecksum == nothingChecksum {
+					return ErrNothingToImport
```
alexey-igrychev: To improve UX let's add information about import (https://github.com/werf/werf/pull/7166/files#diff-870966f61da8074b1f9670667b9eba207bd784e500fdf775404a7f192f86e2a5R100)

PR #7176, opened 2025-10-31
feat(build): ability to save the render along a custom path

reyreavman:


cmd/werf/common/common.go
```diff
@@ -99,6 +99,11 @@ func SetupConfigTemplatesDir(cmdData *CmdData, cmd *cobra.Command) {
 	cmd.Flags().StringVarP(cmdData.ConfigTemplatesDir, "config-templates-dir", "", os.Getenv("WERF_CONFIG_TEMPLATES_DIR"), `Custom configuration templates directory (default $WERF_CONFIG_TEMPLATES_DIR or .werf in working directory)`)
 }
 
+func SetupConfigRenderDir(cmdData *CmdData, cmd *cobra.Command) {
+	cmdData.ConfigRenderDir = new(string)
+	cmd.Flags().StringVarP(cmdData.ConfigRenderDir, "config-render-dir", "", os.Getenv("WERF_CONFIG_RENDER_DIR"), `Custom directory for storing rendered configuration files (default $WERF_CONFIG_RENDER_DIR or .werf in working directory)`)
```
alexey-igrychev: - We should give the user the option to set a custom path (not a directory).
- `.werf` is a default folder for user templates ([rel](https://werf.io/docs/v2/usage/project_configuration/werf_yaml_template_engine.html#template-directory)).
- By default rendered werf.yaml stored somewhere in the tmp directory (`--tmp-dir`), but it may not be described here.

pkg/config/parser.go
```diff
@@ -92,7 +92,7 @@ func GetWerfConfig(ctx context.Context, customWerfConfigRelPath, customWerfConfi
 			return fmt.Errorf("unable to render werf config: %w", err)
 		}
 
-		werfConfigRenderPath, err := tmp_manager.CreateWerfConfigRender(ctx)
+		werfConfigRenderPath, err := tmp_manager.CreateWerfConfigRender(customWerfConfigRenderDir)
```
alexey-igrychev: You should not use `tmp_manager` if the user sets a custom render path.

pkg/file_manager/common.go
```diff
@@ -0,0 +1,29 @@
...
+	"os"
+)
+
+func newFileWithPath(path string) (string, error) {
+	var newFile *os.File
+	var err error
+
+	if path == "" {
+		return "", fmt.Errorf("empty path")
+	} else {
```
nervgh: `else` block looks as redundant one because `if` block above uses "early exit" pattern

alexey-igrychev (approved):

PR #7204, opened 2025-11-11
refactor(build): use files for output in parallel mode

nervgh:


pkg/util/parallel/worker.go
```diff
@@ -1,32 +1,109 @@
...
+	return nil
 }
 
-func newBufferedWorker(id int) *bufferedWorker {
-	return &bufferedWorker{
-		ID:     id,
-		buf:    util.NewGoroutineSafeBuffer(),
-		doneCh: make(chan struct{}, 1),
+func newParallelWorker(id int) (*parallelWorker, error) {
+	file, err := os.CreateTemp("", fmt.Sprintf("parallel-worker-%d-%d-*.log", os.Getpid(), id))
```
alexey-igrychev: I think we should create tmp with this function https://github.com/werf/werf/blob/d970341aa32cd0da4ee6e2fc600a7ee1402adff3/pkg/tmp_manager/common.go#L41

PR #7209, opened 2025-11-12
feat(build):  command to import/export stages (for current build or all)

reyreavman:


cmd/werf/stages/copy/copy.go
```diff
@@ -0,0 +1,236 @@
...
+		InitManifestCache:           true,
+		InitTrueGitWithOptions: &common.InitTrueGitOptions{
+			Options: true_git.Options{LiveGitOutput: *commonCmdData.LogDebug},
+		},
+	})
+	if err != nil {
+		return fmt.Errorf("component init error: %w", err)
+	}
+
+	if cmdData.From == "" {
```
nervgh: Please, decompose validation and getting values (lines 111-134) of options to separate function. For example: https://github.com/deckhouse/delivery-kit/blob/main/cmd/werf/common/signature.go#L222

pkg/build/conveyor.go
```diff
@@ -439,15 +439,26 @@ func (c *Conveyor) GetFullImageName(ctx context.Context, imageName string) (stri
...
+	for _, desc := range c.imagesTree.GetImagesByName(opts.OnlyFinal) {
 		name, images := desc.Unpair()
 		platforms := util.MapFuncToSlice(images, func(img *image.Image) string { return img.TargetPlatform })
 
 		if len(platforms) == 1 {
 			img := images[0]
-			getter := c.StorageManager.GetImageInfoGetter(img.Name, img.GetLastNonEmptyStage().GetStageImage().Image.GetStageDesc(), opts)
-			imagesGetters = append(imagesGetters, getter)
-		} else {
+			if opts.OnlyFinal {
```
nervgh: Please, decompose logic of "getting a getter per image" into separate private method.

pkg/build/stages/archive_storage_reader.go
```diff
@@ -0,0 +1,96 @@
...
+
+	if err != nil {
+		return nil, fmt.Errorf("unable to open bundle archive: %w", err)
+	}
+
+	b := bytes.NewBuffer(nil)
+
+	for {
+		header, err := treader.Next()
+		if err == io.EOF {
```
nervgh, line 74: Are you sure that this logic correct? tar.Reader always [returns](https://github.com/deckhouse/delivery-kit/blob/a5b71d16ddc83d48038f23c4da584a04ea203939/pkg/build/signing/elf_signing.go#L126) EOF in the end of reading.

pkg/build/stages/archive_storage_reader.go
```diff
@@ -0,0 +1,96 @@
...
+			return nil, fmt.Errorf("no chart archive found in the bundle archive %q", reader.Path)
+		}
+		if err != nil {
+			return nil, fmt.Errorf("error reading tar archive: %w", err)
+		}
+
+		if header.Typeflag != tar.TypeReg {
```
nervgh, line 81: Directories will be skipped. Is it right?

pkg/build/stages/archive_storage_writer.go
```diff
@@ -0,0 +1,118 @@
...
+
+func (writer *ArchiveStorageFileWriter) Open() error {
+	p := fmt.Sprintf("%s.%s.tmp", writer.Path, uuid.New().String())
+
+	f, err := os.Create(p)
+	if err != nil {
+		return fmt.Errorf("unable to open tmp archive file %q: %w", p, err)
+	}
+
+	zipper := gzip.NewWriter(f)
```
nervgh, line 45: Do we need to change composition order to reach "archive.tar.gz" (tar -> gz) ?

pkg/build/stages/archive_storage_writer.go
```diff
@@ -0,0 +1,118 @@
...
+	zipper.Header.Comment = "stage-archive"
+	twriter := tar.NewWriter(zipper)
+
+	writer.tmpArchivePath = p
+	writer.tmpArchiveWriter = twriter
+	writer.tmpArchiveCloser = func() error {
+		if err := twriter.Close(); err != nil {
```
nervgh: I guess it should have reverse order: gz.close() -> tar.close() -> f.close() ... Or not?

pkg/build/stages/archive_storage_writer.go
```diff
@@ -0,0 +1,118 @@
...
+	"github.com/google/uuid"
+)
+
+type ArchiveStorageWriter interface {
+	Open() error
+	WriteStageArchive(tag string, data []byte) error
+	Save() error
+}
+
+type ArchiveStorageFileWriter struct {
```
nervgh, line 23: Just my opinion (optional): `ArchiveStorageFileWriter` should be `ArchiveStorageFileWriteCloser`. It clarifies that we are responsible to close it anyway (even without calling Save()).

nervgh (changes requested):

pkg/build/stages/archive_storage_writer.go
reyreavman: Thank you for raising this important question about the composition order!

When I implemented this, I actually researched this specific topic and referenced [Arthur Koziel's article on writing tar.gz files in Go](https://www.arthurkoziel.com/writing-tar-gz-files-in-go/), which explains the standard approach.

According to the article and common practice, the correct data flow should be:
file → gzip.Writer → tar.Writer

In my current implementation, I'm following this approach.

If you see any specific places in the code where this order might be incorrect, or if you have additional insights based on your experience, I'd be happy to discuss and make adjustments!

pkg/build/stages/archive_storage_writer.go
reyreavman: You're absolutely right about the closing order! Thank you for catching this important detail.

Yes, the closing order should indeed be the reverse of the creation order. Looking back at the same [Arthur Koziel article I referenced earlier](https://www.arthurkoziel.com/writing-tar-gz-files-in-go/), I made my own implementation

cmd/werf/stages/copy/copy.go
```diff
@@ -0,0 +1,272 @@
...
+		InitLRUImagesCache:          true,
+		InitTrueGitWithOptions: &common.InitTrueGitOptions{
+			Options: true_git.Options{LiveGitOutput: *commonCmdData.LogDebug},
+		},
+	})
+	if err != nil {
+		return fmt.Errorf("component init error: %w", err)
+	}
+
+	opts, err := getCopyOptions()
```
nervgh: Please, pass [cmdData](https://github.com/werf/werf/pull/7209/files#diff-b6ffb79c53a3630b60db5e425fa941c73af8f6cc8d7fa4c4dc84d6141b05632dR25) as the argument ino getCopyOptions() and into validateRawCopyOptions() to simplify reading and testing of these functions.

cmd/werf/stages/copy/copy.go
```diff
@@ -0,0 +1,272 @@
...
+	if err != nil {
+		return err
+	}
+
+	if opts.From.RegistryAddress != nil {
+		commonCmdData.Repo.Address = &opts.From.RegistryAddress.Repo // FIXME выдумать что-нить симпатичнее
+	} else if opts.To.RegistryAddress != nil {
+		commonCmdData.Repo.Address = &opts.To.RegistryAddress.Repo
+	} else {
+		return fmt.Errorf("--from or --to address must be container registry address")
```
nervgh: I think we should move this check into validation function.

pkg/build/conveyor.go
```diff
@@ -583,6 +602,14 @@ func (c *Conveyor) determineStages(ctx context.Context) error {
 		})
 }
 
+func (c *Conveyor) Calculate(ctx context.Context) error {
```
nervgh: It is unused method

pkg/build/stages/archive_storage.go
```diff
@@ -0,0 +1,128 @@
...
+	return NewStageArchiveOpener(s, stageTag)
+}
+
+func (s *ArchiveStorage) copyAllFromRemote(ctx context.Context, fromRemote *RemoteStorage, opts copyToOptions) error {
+	stageIds, err := fromRemote.StorageManager.StagesStorage.GetStagesIDs(ctx, opts.ProjectName)
+	if err != nil {
+		return fmt.Errorf("unable to get stages: %w", err)
+	}
+
+	if err := s.Writer.Open(); err != nil {
```
nervgh: There is a chance to open the writer and do not call Save(). Then we miss closing of the file, the tar and the gzip writers. 

Because of that, we need to rethink the design to prevent this case.

pkg/build/stages/archive_storage_reader.go
```diff
@@ -0,0 +1,122 @@
...
+		if err == io.EOF {
+			break
+		}
+		if err != nil {
+			return nil, fmt.Errorf("error reading tar archive: %v", err)
+		}
+
+		if header.Typeflag == tar.TypeReg {
+			filename := filepath.Base(header.Name)
+			if strings.HasSuffix(filename, ".tar.gz") {
```
nervgh: Please, declare this extension (".tar.gz") as the string constant.

pkg/build/stages/archive_storage_reader.go
```diff
@@ -0,0 +1,122 @@
...
+		}
+	}
+
+	return tags, nil
+}
+
+func (reader *ArchiveStorageFileReader) ReadArchiveStage(stageTag string) (*StageArchiveReadCloser, error) {
+	treader, closer, err := reader.openForReading()
+	if err != nil {
+		defer closer()
```
nervgh: It looks strange: we name it "closer" and call it if an error happened. 

For example:

```go
f, err := os.Open(name)
if err != nil {
  return err
}
defer f.close() # <-- I mean we should call closer only when "no error"
 ```

File: test/e2e/stages/copy/simple_test.go [hunk 14]
```diff
@@ -0,0 +1,97 @@
+package e2e_stages_copy_test
+
+//
```
nervgh: Why did you disable "simple" test? If you don't need it just remove the file.

nervgh (changes requested):

test/e2e/stages/copy/complex_test.go
```diff
@@ -0,0 +1,158 @@
...
+				By("state0: check stages copy output")
+				Expect(stagesCopyOut).To(ContainSubstring(fmt.Sprintf("Copy stages")))
+				Expect(stagesCopyOut).To(ContainSubstring(fmt.Sprintf("From: %s", SuiteData.WerfFromAddr)))
+				Expect(stagesCopyOut).To(ContainSubstring(fmt.Sprintf("To: %s", SuiteData.WerfToAddr)))
+
+				By("state0: check that images were built successfully")
+				werfProject.Build(ctx, &werf.BuildOptions{
+					CommonOptions: werf.CommonOptions{
+						ShouldFail: false,
+						ExtraArgs:  []string{"--require-built-images", "--repo", SuiteData.WerfToAddr},
```
nervgh, line 62: I think we should expect output "Use previously built image" because of --require-built-images option

test/e2e/stages/copy/complex_test.go
```diff
@@ -0,0 +1,158 @@
...
+				if opts.All {
+					shouldFail = false
+				} else {
+					shouldFail = true
+				}
+
+				werfProject.Build(ctx, &werf.BuildOptions{
+					CommonOptions: werf.CommonOptions{
+						ShouldFail: shouldFail,
+						ExtraArgs:  []string{"--require-built-images", "--repo", SuiteData.WerfToAddr},
```
nervgh, line 119: I think we should expect output "Use previously built image" because of --require-built-images option

pkg/build/stages/archive_storage.go
```diff
@@ -0,0 +1,120 @@
...
+
+			stageDesc, err := fromRemote.StorageManager.StagesStorage.GetStageDesc(ctx, opts.ProjectName, stageId)
+			if err != nil {
+				return err
+			}
+
+			stageRef := stageDesc.Info.Name
+			tag := stageDesc.Info.Tag
+
+			stageBytes := bytes.NewBuffer(nil)
```
nervgh, line 77: Just my opinion for future: stageBytes could have very large size (more than free RAM), so it's better to use [io.Pipe() ](https://pkg.go.dev/io#Pipe) or something like that.

pkg/build/stages/archive_storage.go
```diff
@@ -0,0 +1,120 @@
...
+
+			infoGetters, err := c.GetImageInfoGettersWithOpts(image.InfoGetterOptions{OnlyFinal: false})
+			if err != nil {
+				return err
+			}
+
+			for _, infoGetter := range infoGetters {
+				logboek.Context(ctx).Default().LogFDetails("Copying stage: %s\n", infoGetter.Tag)
+
+				stageBytes := bytes.NewBuffer(nil)
```
nervgh, line 106: Same here

pkg/build/stages/archive_storage_reader.go
```diff
@@ -0,0 +1,126 @@
...
+func (reader *ArchiveStorageFileReader) openForReading() (*tar.Reader, func() error, error) {
+	f, err := os.Open(reader.Path)
+	if err != nil {
+		return nil, func() error { return nil }, err
+	}
+
+	unzipper, err := gzip.NewReader(f)
+	if err != nil {
+		if closeErr := f.Close(); closeErr != nil {
+			logboek.Warn().LogF("Warning: error closing file after gzip error: %v\n", closeErr)
```
nervgh: Hmmm... pass ctx or don't use logging like that

pkg/build/stages/archive_storage_writer.go
```diff
@@ -0,0 +1,168 @@
...
+func (writer *ArchiveStorageFileWriter) WithTask(task func(ArchiveStorageWriter) error) error {
+	if err := writer.open(); err != nil {
+		return fmt.Errorf("unable to open target stages archive: %w", err)
+	}
+
+	var err error
+	defer func() {
+		if err != nil {
+			if closeErr := writer.Close(); closeErr != nil {
+				fmt.Printf("Warning: error closing archive after task failure: %v\n", closeErr)
```
nervgh: Please, use logger with ctx or handle the error in other way (e.g. returning from func)

PR #7226, opened 2025-11-17
feat(build): add toYaml func to config parser

eshutov:
This PR adds the `toYaml` templating function to the werf.yaml renderer context.
The code was intentionally taken from the [Helm repository](https://github.com/helm/helm/blob/99cd1964357c793351be481d55abbe21c6b2f4ec/pkg/engine/funcs.go#L86) to preserve user experience with helm tool.

The code has been tested:
werf.yaml:
```
shell:
  install:
{{- .WerfValues.backend.builder.installPost | toYaml | nindent 4 }}
```
werf-values.yaml:
```
backend.builder.installPost:
  - echo 1
  - echo 2
```
Results the following output in `werf config render`:
```
shell:
  install:
    - echo 1
    - echo 2
```
Don’t forget to remove the corresponding ticket (for this PR) created in your private backlog.


pkg/config/parser.go
```diff
@@ -339,6 +339,14 @@ func funcMap(ctx context.Context, tmpl *template.Template, giterminismManager gi
 
 		return m, nil
 	}
+	funcMap["toYaml"] = func(v interface{}) (string, error) {
+		data, err := yaml.Marshal(v)
+		if err != nil {
+			return "", err
```
ilya-lesikov: For consistency let's do this exactly like Helm does, ignoring errors of marshaling: https://github.com/helm/helm/blob/a75fcc29484730e14ba203f4593f81e6788b0152/pkg/engine/funcs.go#L90

eshutov: Initially, I also considered adding `mustToYAML` here to provide two behaviors like Helm does: with error suppression and without. However:
- In practice, I haven't been able to make the values structure that couldn't be marshaled into valid YAML.
- On the other hand, suppressing errors even for an artificially created invalid structure (defined directly in the code) doesn't work:
  ```
  parser.go:
  funcMap["toYaml"] = func(v interface{}) string {
      data, err := yaml.Marshal(v)
      if err != nil {
          return ""
      }

      return strings.TrimSuffix(string(data), "\n")
  }
  funcMap["invalid"] = func() interface{} {
      return map[string]interface{}{
          "channel": make(chan int),
      }
  }
  ```
  ```
  werf.yml:
  {{ invalid | toYaml }}
  ```
  This leads to:
  ```
  Error: unable to render werf config: template: werfConfig:50:18: executing "werfConfig" at <toYaml>: error calling toYaml: cannot marshal type: chan int
  Use --debug-templates or $WERF_DEBUG_TEMPLATES to get more details about this error.
  ```
  I suspect this might be a kind of panic. This issue goes far beyond the scope of this feature commit.

Therefore, `mustToYAML` doesn't provide practical value.

Changes made:
- Error suppression implemented as requested
- Documentation updated
- A minor typo with `fromYaml` has been fixed

ilya-lesikov: Thanks. The closer it is to Helm functions the better. We will merge werf.yaml and Helm templating engines at some point.

ilya-lesikov:
Thanks!

PR #7249, opened 2025-11-28
fix(config): align .Files.Glob behavior with helm

reyreavman:


pkg/config/parser.go
```diff
@@ -582,8 +582,8 @@ func (f files) doGlob(ctx context.Context, pattern string) (map[string]interface
 		return nil, err
 	}
 
-	if len(res) == 0 {
```
nervgh, line 585: As we know there are two zero cases for slices:

1. nil as zero
2. empty slice (`[]string{}`) as zero

Why don't  we treat "an empty slice case" as zero case? For our user they both are equal I guess.

reyreavman: Good catch, mate!

I will make the length check equal to zero and normalize the result to an empty map — so the zero‑case will be uniform and we will not have to rely on the implementation details of **ConfigGoTemplateFilesGlob**.

I think that it is impossible to get rid of this branch completely - the calling code may get into an unpleasant situation =)

PR #7263, opened 2025-12-04
chore(ci): add trigger website deploy job

odiora:


.github/workflows/release_trdl-publish.yml
```diff
@@ -218,13 +218,28 @@ jobs:
 
           process_releases
 
+  run_website_deploy:
+    if: success()
+    runs-on: ubuntu-22.04
+    needs:
+      - publish
+      - deploy_channels_cm_prod
+      - publish_werf_images
```
Fral738: Leave only `update_release` in `needs`

.github/workflows/release_trdl-publish.yml
```diff
@@ -218,13 +218,28 @@ jobs:
...
+      - update_release
+    steps:
+      - name: Trigger Deploy workflow in werf/website
+        run: gh workflow run Deploy --repo werf/website --ref main
+        env:
+          GH_TOKEN: ${{ secrets.RELEASE_PLEASE_TOKEN }}
```
Fral738, line 230: I don't think this token has enough permissions, but we can try

PR #7273, opened 2025-12-08
test(build, import): should not create empty directories and account symlinks in import checksum

reyreavman:


pkg/build/stage/dependencies.go
```diff
@@ -424,7 +424,16 @@ func generateChecksumScript(from string, includePaths, excludePaths []string, re
 
 	// We have an old rsync version, so we can't use --out-format and other options to parse file paths.'
 	// Example line: "-rw-r--r--    1 root     root             0 Nov 17 22:57 test-file.a".
```
nervgh:
```suggestion
// Example lines: 
// "-rw-r--r--    1 root     root             0 Nov 17 22:57 test-file.a" 
// "lrw-r--r--    1 root     root             0 Dec 10 10:07 path/to/link -> target/path"
```

integration/suites/build/stapel_image/imports/imports_test.go
```diff
@@ -144,6 +146,17 @@ var _ = Describe("Stapel imports", func() {
 
 			Expect(output).To(ContainSubstring("11111:11111"))
 		})
+
+		It("should remove empty directories in final image when importing stages", func(ctx SpecContext) {
+			SuiteData.CommitProjectWorktree(ctx, SuiteData.ProjectName, utils.FixturePath("import_app_empty_dirs", "001"), "initial commit")
+
+			Expect(werfBuild(ctx, SuiteData.GetProjectWorktree(SuiteData.ProjectName), liveexec.ExecCommandOptions{})).To(Succeed())
+
+			output := werfRunOutputWithSpecificImage(ctx, SuiteData.GetProjectWorktree(SuiteData.ProjectName),
```
nervgh: Current assertion is not robust. For example, I do 

```bash
find /tmp -type d -empty | wc -l | grep 0
find: ‘/tmp/snap-private-tmp-100’: Permission denied
find: ‘/tmp/snap-private-tmp-101’: Permission denied
... (many same lines)
```

and I have the result "True".

Please use more straightforward way to assert each item. [There is a good example with `test`](https://github.com/werf/werf/blob/68b1a987042b54664bc263fa0226041d1fee6176/test/e2e/build/complex_test.go#L84).

Pseudocode:
```bash
test -d /dir/exist
! test -d /dir/not/exist
```

nervgh (approved):

PR #7297, opened 2025-12-23
feat(build): use build report in commands require build

reyreavman:


pkg/build/build_report.go
```diff
@@ -250,3 +254,125 @@ func getStagesReport(img *image.Image, multiplatform bool) []ReportStageRecord {
...
+
+	if err := validateBuildReport(report); err != nil {
+		return nil, fmt.Errorf("invalid build report file %q: %w", path, err)
+	}
+
+	return report, nil
+}
+
+func parseBuildReport(data []byte) (*ImagesReport, error) {
+	decoder := json.NewDecoder(bytes.NewReader(data))
```
nervgh: (Optional) `json.NewDecoder()` takes `io.Reader` as its argument. So we could simplify our code (I) and take advantage of buffered reading (II):

```go
func LoadBuildReportFromFile(path string) (*ImagesReport, error) {
  file, err := os.OpenFile(path) // (1)
  if err != nil {
  	  return nil, fmt.Errorf("unable to open build report file %q: %w", path, err)
  }
  defer file.Close()

  report, err := parseBuildReport(file) // (2)
  if err != nil {
  	  return nil, fmt.Errorf("unable to parse build report file %q: %w", path, err)
  }

    // ...
}

func parseBuildReport(reader io.Reader) (*ImagesReport, error) {
	decoder := json.NewDecoder(reader) // (3)
    // ...
```

pkg/build/export_phase.go
```diff
@@ -75,6 +77,125 @@ func (e *Exporter) Run(ctx context.Context) error {
...
+	if err != nil {
+		return fmt.Errorf("unable to load build report: %w", err)
+	}
+
+	imagesToExport := e.filterImagesFromReport(report)
+	if len(imagesToExport) == 0 {
+		return nil
+	}
+
+	if err := parallel.DoTasks(ctx, len(imagesToExport), parallel.DoTasksOptions{
```
nervgh, line 101: (Optional) What if `--parallel=false` ?

pkg/build/export_phase.go
```diff
@@ -131,3 +252,49 @@ func (e *Exporter) exportImage(ctx context.Context, img *build_image.Image) erro
...
+			Tag:               record.DockerTag,
+			RepoDigest:        fmt.Sprintf("%s@%s", record.DockerRepo, record.DockerImageDigest),
+			ID:                record.DockerImageID,
+			Size:              record.Size,
+			CreatedAtUnixNano: lastStage.CreatedAt,
+		},
+	}
+}
+
+func parseStageTag(tag string) (digest string, creationTs int64) {
```
nervgh: Let's reuse existing code: https://github.com/werf/werf/blob/c8085df2db4b46402f96dd8ecde5c9bf0f7c7266/pkg/image/summary.go#L48

pkg/build/export_phase.go
```diff
@@ -75,6 +77,125 @@ func (e *Exporter) Run(ctx context.Context) error {
...
+		}).
+		DoError(func() error {
+			stageDesc := stageDescFromReportRecord(record)
+
+			for _, tagFunc := range e.ExportTagFuncList {
+				stageID := extractStageIDFromReport(record)
+				tag := tagFunc(record.WerfImageName, stageID)
+				if err := logboek.Context(ctx).Default().LogProcess("tag %s", tag).
+					DoError(func() error {
+						if err := e.Conveyor.StorageManager.GetStagesStorage().ExportStage(ctx, stageDesc, tag, e.MutateConfigFunc); err != nil {
```
nervgh: If we don't handle the error we could just forward returned value.

pkg/build/export_phase.go
```diff
@@ -75,6 +77,125 @@ func (e *Exporter) Run(ctx context.Context) error {
...
+		DoError(func() error {
+			stageDesc := stageDescFromReportRecord(record)
+			stageDesc.Info.IsIndex = true
+
+			for _, tagFunc := range e.ExportTagFuncList {
+				stageID := extractStageIDFromReport(record)
+				tag := tagFunc(record.WerfImageName, stageID)
+				if err := logboek.Context(ctx).Default().LogProcess("tag %s", tag).
+					DoError(func() error {
+						if err := e.Conveyor.StorageManager.GetStagesStorage().ExportStage(ctx, stageDesc, tag, e.MutateConfigFunc); err != nil {
```
nervgh, line 166: If we don't handle the error we could just forward returned value.

pkg/build/export_phase.go
```diff
@@ -75,6 +77,125 @@ func (e *Exporter) Run(ctx context.Context) error {
...
+					}); err != nil {
+					return err
+				}
+			}
+
+			return nil
+		})
+}
+
+func (e *Exporter) exportMultiplatformImageFromReport(ctx context.Context, record ReportImageRecord) error {
```
nervgh: Could we reduce code duplication from `exportImageFromReport` and `exportMultiplatformImageFromReport` using single method with param like `IsMultiplatform`?

pkg/build/export_phase.go
```diff
@@ -131,3 +252,49 @@ func (e *Exporter) exportImage(ctx context.Context, img *build_image.Image) erro
 			return nil
 		})
 }
+
+func stageDescFromReportRecord(record ReportImageRecord) *image.StageDesc {
+	var lastStage ReportStageRecord
+	if len(record.Stages) > 0 {
+		lastStage = record.Stages[len(record.Stages)-1]
```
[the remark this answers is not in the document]
reyreavman: CreatedAtUnixNano is just a number, not a ptr or something like that
I made the code more explicit and clean, hope this is what needed

pkg/build/export_phase.go
nervgh: The question still exists. Could we use `exportImageFromReport` for single and multi-platform cases? If we can then, please, remove `exportMultiplatformImageFromReport`

pkg/build/export_phase.go
nervgh: I moved this question into separate issue.

PR #7299, opened 2025-12-24
fix(build): sanitize docker credentials from buildkit and docker errors

Fral738:


pkg/container_backend/errors.go
```diff
@@ -25,6 +26,10 @@ Possible solutions:
 
   - If these files are not needed, exclude them using the includePaths or excludePaths options under the git directive.`)
 
+var dockerRateLimitCredsRe = regexp.MustCompile(
```
reyreavman: this regex and the message ("You have re...hidden).") that we are replacing the original one are duplicated, let's put it in a separate place and reuse it.

pkg/container_backend/docker_server_backend.go
```diff
@@ -246,6 +246,7 @@ func (backend *DockerServerBackend) RemoveImage(ctx context.Context, img LegacyI
 
 func (backend *DockerServerBackend) PullImageFromRegistry(ctx context.Context, img LegacyImageInterface) error {
 	if err := img.Pull(ctx); err != nil {
+		err = SanitizeError(err)
```
reyreavman, line 249: It`s only one implementation of ContainerBackend, i suggest we bring this logic [here](https://github.com/werf/werf/blob/d30a3d0e8b77e30ba6988259a2cb7f7b72229e0c/pkg/container_backend/interface.go#L127)

pkg/container_backend/errors.go
reyreavman: this regex and the message ("You have re...hidden).") that we are replacing the original one are duplicated, let's put it in a separate place and reuse it.

pkg/logging/sanitize.go
```diff
@@ -0,0 +1,30 @@
...
+
+func (s *SanitizeWriter) Write(p []byte) (int, error) {
+	out := s.re.ReplaceAllString(
+		string(p),
+		"You have reached your pull rate limit (credentials hidden).",
+	)
+	return s.w.Write([]byte(out))
+}
+
+func WrapIfNeeded(w io.Writer) io.Writer {
```
reyreavman: i would suggest something like "WrapInSanitizeWriter", it's much clearer what we wrap and why

```suggestion
func WrapInSanitizeWriter(w io.Writer) io.Writer {
```

reyreavman (changes requested):

pkg/container_backend/docker_server_backend.go
reyreavman: lets get `SanitizeError(err)` call out of here,  we're already handling it at a higher level, there's no need to call it here

Fral738: If I remove error handling from here, then the final error will include creds

PR #7300, opened 2025-12-25
fix(build, dockerfile, staged): fix staged dockerfile dependencies for COPY --from

Fral738:


pkg/build/image/dockerfile.go
```diff
@@ -248,6 +248,15 @@ func mapDockerfileToImagesSets(ctx context.Context, cfg *dockerfile.Dockerfile,
 
 			img.stages = append(img.stages, stg)
 
+			if copyInstr, ok := instr.(*dockerfile.DockerfileStageInstruction[*instructions.CopyCommand]); ok {
```
reyreavman: can't we get rid of this block? there is a feeling that it duplicates the logic that runs in the next "for ... range"

reyreavman (approved):

PR #7303, opened 2025-12-26
chore(ci): unify test_main, test_pr into single workflow

odiora:


.github/workflows/_build.yml
```diff
@@ -0,0 +1,43 @@
...
+    strategy:
+      fail-fast: false
+      matrix:
+        osPlatform:
+          - build:dev:linux:amd64:cgo
+          - build:dev:linux:amd64:go
+          - build:dev:linux:arm64
+          - build:dev:darwin:amd64
+          - build:dev:darwin:arm64
+          - build:dev:windows:amd64
```
Fral738: `build:dev:all`?

.github/workflows/tests.yml
```diff
@@ -0,0 +1,174 @@
+name: tests:all
+
+on:
+  pull_request:
+  push:
+    branches: [main]
```
Fral738: Should this be here if we want to run tests in pull requests too?

PR #7306, opened 2025-12-30
fix(build, import): add empty dirs if they correspond to globs

reyreavman:


pkg/build/import_server/rsync_server_test.go
```diff
@@ -71,16 +150,492 @@ func Test_globToRsyncFilterPaths(t *testing.T) {
...
+			want: []string{"/app/src/*.go", "/app/src/**/*.go"},
+		},
 	}
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
 			got := globToRsyncFilterPaths(tt.args.glob, tt.args.finalSegmentOnly)
 			sort.Strings(got)
 			sort.Strings(tt.want)
+			if len(got) == 0 && len(tt.want) == 0 {
+				return
```
nervgh: "continue" ?

pkg/build/import_server/rsync_server_test.go
```diff
@@ -71,16 +150,492 @@ func Test_globToRsyncFilterPaths(t *testing.T) {
...
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			got := PrepareRsyncExcludeFiltersForGlobs(tt.add, tt.excludeGlobs)
+			if len(tt.excludeGlobs) == 0 || len(tt.wantContains) == 0 {
+				if len(tt.wantContains) == 0 && got != "" {
+					t.Errorf("PrepareRsyncExcludeFiltersForGlobs() = %v, want empty string", got)
+				}
+				return
```
nervgh: "continue" ?

nervgh (changes requested):

pkg/build/import_server/rsync_server.go
```diff
@@ -325,7 +498,12 @@ func globToRsyncFilterPaths(glob string, finalSegmentOnly bool) []string {
 
 	out := make([]string, 0, len(results))
 	for v := range results {
-		out = append(out, v)
+		// restore leading slash if original glob had one
+		if hasLeadingSlash {
+			out = append(out, "/"+v)
```
nervgh, line 491: If we have the leading slash (435 line) this code adds prefix "/" for **each** result item, not just for first one. Should we move this logic outside of the loop?

reyreavman: i guess adding `/` prefix to each result item is the correct and expected behavior, each element in results is an independent rsync filter path, not parts of a single path
when the original glob is absolute (starts with `/`), all generated paths must also be absolute

for example, glob like `/app/**/*.txt` produces:
`/app/`
`/app/**/`
`/app/**/*.txt`

each of these is a separate rsync filter rule, and each must start with `/` to correctly match absolute paths in rsync

moving this logic outside the loop wouldn't change the behavior — we'd still need to add / to every element

nervgh (approved):

pkg/build/import_server/rsync_server.go
```diff
@@ -158,6 +186,76 @@ func (srv *RsyncServer) GetCopyCommand(ctx context.Context, importConfig *config
 	return command
 }
 
+// hasDirectoryGlobs returns true if any of the globs explicitly targets a directory
+// directory globs are patterns that should preserve empty directories when matched
+// examples of directory globs: "cache", "cache/", "logs/**", "app/data"
+// examples of file globs: "*.txt", "**/*.go", "file?.log"
+func hasDirectoryGlobs(globs []string) bool {
```
alexey-igrychev: In the current changes, globs are distinguished as "file globs" (e.g., "*.txt", "**/*.go", "file?.log") and "directory globs" (e.g., "cache", "cache/", "logs/**"). However, in reality, there are no restrictions on naming directories—directory names can contain dots, look like file extensions, or use wildcards. There is no reliable way to differentiate between file and directory globs based on the pattern alone. Globs targeting files are not fundamentally different from globs targeting directories; the distinction does not actually exist. It is recommended to use a generic approach for processing globs, treating all globs in a uniform way.

Reason:
On filesystems and in rsync, objects are only defined as directories or files by their actual type, not by their name or pattern. Therefore, filter and matching logic should not rely on assumptions from the glob string itself, but instead handle patterns in a way that works generically for both files and directories.

nervgh (approved):

PR #7315, opened 2026-01-16
chore(telemetry): add build, image and stage duration metrics

odiora:


Fral738 (approved):

pkg/build/build_report.go
```diff
@@ -210,6 +248,8 @@ func createBuildReport(ctx context.Context, phase *BuildPhase, imagePairs []util
 		}
 	}
 
+	phase.ImagesReport.sendTelemetry(ctx)
```
nervgh: Let's place telemetry sending before writing a report on disk (from 251 line to 188 line).

PR #7329, opened 2026-01-22
fix(build): use registry-mirrors from docker daemon.json

odiora:


pkg/docker/system.go
```diff
@@ -9,3 +9,31 @@ import (
...
+// GetRegistryMirrors returns registry mirrors from Docker daemon.
+func GetRegistryMirrors(ctx context.Context) ([]string, error) {
+	if !IsEnabled() {
+		return nil, nil
+	}
+
+	var info types.Info
+	var err error
+
+	if IsContext(ctx) {
```
nervgh, line 22: I think this condition is redundant because `apiCli` does the same check under the hood.

cmd/werf/common/components_manager.go
```diff
@@ -70,6 +71,21 @@ func InitCommonComponents(ctx context.Context, opts InitCommonComponentsOptions)
...
 	}
 
+	// Initialize Docker early to read registry mirrors from daemon.json before GetContainerRegistryMirror call
+	if opts.InitProcessContainerBackend {
+		buildahMode, _, err := GetBuildahMode()
+		if err != nil {
+			return nil, ctx, fmt.Errorf("unable to determine buildah mode: %w", err)
+		}
+		if *buildahMode == buildah.ModeDisabled {
+			newCtx, err := InitProcessDocker(ctx, opts.Cmd)
```
nervgh, line 81: We have several problems with this approach:

1. Docker initialization duplication:

    1. first init here
     2. second init [there](https://github.com/werf/werf/blob/6a4078be246c737f60b39ef46e63175f29baa438/cmd/werf/common/container_backend.go#L132)

2. (Optional) Buildah [does not require](https://github.com/werf/werf/blob/6a4078be246c737f60b39ef46e63175f29baa438/cmd/werf/common/container_backend.go#L116) early initialization for registry mirrors.

odiora: 1. You're right about the duplication. Fixed in `container_backend.go` by adding `docker.IsContext(ctx)` check before calling `InitProcessDocker`. This way, if Docker was already initialized early (for reading registry mirrors), it won't be initialized again. The early init in `components_manager.go` is needed because `GetContainerRegistryMirror` requires Docker to read mirrors from daemon.json.
2. This is already handled — early Docker initialization only happens when `buildahMode == buildah.ModeDisabled`. When Buildah is used, this block is skipped, so there's no unnecessary initialization.

pkg/docker/system.go
odiora: The `IsContext(ctx)` check is necessary to prevent panic. The `cli(ctx)` function (called by `apiCli`) panics when context doesn't contain docker cli and is not `context.Background()`:
```
func cli(ctx context.Context) command.Cli {
    cliInterf := ctx.Value(ctxDockerCliKey)
    switch {
    case cliInterf != nil:
        return cliInterf.(command.Cli)
    case ctx == context.Background():
        return defaultCLI
    default:
        panic("context is not bound with docker cli")
    }
}
```
Our check allows safe fallback to `defaultCLI` without panic.

cmd/werf/common/container_backend.go
```diff
@@ -129,11 +129,13 @@ func InitProcessContainerBackend(ctx context.Context, cmdData *CmdData, registry
 		return wrapContainerBackend(container_backend.NewBuildahBackend(b, container_backend.BuildahBackendOptions{TmpDir: filepath.Join(werf.GetServiceDir(), "tmp", "buildah")})), ctx, nil
 	}
 
-	newCtx, err := InitProcessDocker(ctx, cmdData)
-	if err != nil {
-		return nil, ctx, fmt.Errorf("unable to init process docker for docker-server container backend: %w", err)
+	if !docker.IsContext(ctx) {
```
nervgh: Let's move this check inside of  `InitProcessDocker` to get more robust preventing duplication of initialization. 

If `docker.IsContext(ctx)` returns true we just return passed context and do nothing.

PR #7377, opened 2026-02-27
v3

### LINKED ISSUE #6127: Lock files accumulation after prolonged use of werf on host
### Before proceeding

### Version

2.0.0

### How to reproduce

Use werf on a runner for an extended period.

### Result

After prolonged use of werf on a runner, lock files accumulate. Some are over a year old, and the situation is similar on other runners.

```
# ls /home/ubuntu/.werf/service/locks/ | wc -l
1715600

# df -ih
Filesystem     Inodes IUsed IFree IUse% Mounted on
/dev/sdb         3.2M  2.8M  351K   90% /home/ubuntu

# ls -lah /tmp/werf-host-cleanup.log 
-rw-rw-r-- 1 ubuntu ubuntu 4.7K Apr  4 09:00 /tmp/werf-host-cleanup.log
```

### Expected result

Automatic removal of old lock files and prevention of their accumulation.

### Additional information

Workaround:
```
while pgrep "werf" > /dev/null; do sleep 1; done && rm -rf ~/.werf/service/locks
```

alexey-igrychev:
## Summary

Umbrella PR for **werf v3**: removes deprecated and legacy subsystems (Ansible builder, `artifact` type, virtual merge, synchronization service), unifies the `from`/`import.from` configuration surface, makes builds self-contained (no external sync service, stapel image embedded into the binary), introduces content-anchor-based image reuse and multiplatform builds for Docker/Stapel backends, extends the deploy side with `werf release get/list/history`, plan artifacts and `--meta-repo`, switches the native Buildah backend's network stack to `netavark`/`pasta`, and upgrades the core toolchain (Go 1.26, docker/buildx v0.37.1, buildkit v0.33, buildah v1.44.1, Kubernetes client 0.37 — the `3p-buildah` fork dropped, official images on Ubuntu 24.04).

### Configuration (`werf.yaml`)
- **Remove Ansible builder** from Stapel images — only the Shell builder remains.
- **Remove `artifact` image type** — use `image` with `final: false` instead.
- **Remove `docker` directive** from Stapel images — use `imageSpec.config` instead.
- **Deprecate `fromImage` and `import.image`** — they now work as aliases for `from` with a deprecation warning; specifying both old and new keys together is a hard error.
- **Remove `import.stage` and `disableGitAfterPatch` directives.**
- **Remove nameless image support** — all images must have explicit names.
- **Require explicit tag or digest for external images** in `from` and imports: `:TAG` or `@sha256:...` is now mandatory. Internal `werf.yaml` image names need no tag.
- **Self-reference in `from`** is now a validation error instead of silent `:latest` substitution.
- **Trailing slash in export/import `to:` path** is now a hard config error instead of a warning (previously silently reinterpreted, could produce wrong image contents).
- **`stageDependencies` defaults to `**/*`** when not specified; referencing unused stages is now flagged instead of silently ignored.
- **Strict `werf-giterminism.yaml` schema validation** — unknown/misspelled keys now fail validation instead of being silently ignored.

### CLI / behavior
- **Remove synchronization subsystem**: `--synchronization` / `-S` / `WERF_SYNCHRONIZATION`, the `werf synchronization` daemon command, and the dependency on `synchronization.werf.io` are gone with no deprecated no-op. Builds are self-contained; stage tags are content-addressable, so no distributed lock manager is needed.
- **`.global.env` is no longer exposed to helm values** unless `WERF_LEGACY_VALUES_GLOBAL_ENV` is set.
- **Remove deprecated deploy flags**: `--exit-code`, `--set-runtime-json`, `--show-verbose-diffs`, `--status-progress-period`, `--hooks-status-progress-period`, `--insecure-helm-dependencies`, `-t/--timeout`, plus deploy feature gates (`FeatGateResourceValidation`, `FeatGateMoreDetailedExitCodeForPlan`, ...).
- **Remove support for all `HELM_*` env vars** in deploy.
- **Remove `WERF_EXPERIMENT_NO_GLOBAL_SERVICE_VALUES`** and the deprecated no-op `--docker-compose-bin-path`.
- **`--final-images-only` defaults to `true` for `werf build`** — orphan non-final images (not referenced by any final image) are no longer built by default.
- **`converge`/`plan` positional `[IMAGE_NAME ...]` args are always-on** (previously behind `WERF_CONVERGE_ENABLE_IMAGES_PARAMS`); scripts passing positional args will now get image filtering.
- **Build report schema**: `StagesReused` renamed to `StagesSkipped`; per-image `AnchorReused` flag added. Consumers of `.werf-build-report.json` must adapt (no compatibility alias).
- **Remove git commit ancestry check on stage reuse** (and its `WERF_DISABLE_GIT_COMMIT_ANCESTRY_CHECK` toggle).
- **Remove legacy `DOCKER_BUILDKIT=0` path** — buildx is unconditional.
- **Remove `VirtualMerge`.**
- **Remove `werf.io/base-image-id` label** and `image.Info.ParentID`.
- **Cache identity changes**: import metadata tracking replaced with source image content digest; git file mode now included in the ls-tree checksum. First v3 build against v2 stages storage may partially rebuild.
- **Binary size**: stapel image (linux/amd64 + linux/arm64) is embedded into release binaries, adding ~35–40MB with no opt-out.

### Host / runtime requirements
- **Native Buildah backend switches network stack to `netavark`/`pasta`**, replacing CNI and slirp4netns. Every host running werf in Buildah mode must have the `netavark` binary in a Podman helper binaries directory (`/usr/local/libexec/podman`, `/usr/local/lib/podman`, `/usr/libexec/podman` or `/usr/lib/podman`; `$PATH` is not searched) — its absence fails backend initialization even in chroot mode with no network-using instructions. Rootless mode (`native-rootless`) additionally requires `pasta` from the `passt` package (found via `$PATH`). CNI support is compiled out and cannot be restored; slirp4netns can be brought back per host via `CONTAINERS_CONF_OVERRIDE` (`default_rootless_network_cmd="slirp4netns"` in `[network]`). Existing local Buildah storages migrate to netavark automatically.
- **Official Ubuntu werf images move to Ubuntu 24.04** (the first LTS to package `netavark`); build user UIDs (1000/1001) are preserved despite Ubuntu 24.04 shipping its own UID-1000 user.

## New features

- **Content-anchor stage**: the trailing stage of every image acts as its content anchor — its digest is derived holistically from the target platform and all upstream stage inputs, decoupling published image identity from the build process; on cache hit the whole per-stage build chain is skipped as ordinary stage reuse.
- **Multiplatform builds** for Docker and Stapel backends; stapel arm64 enabled by default (experimental gate removed).
- **Embedded stapel image**: `docker load` from the binary instead of registry pull — first build works air-gapped on amd64/arm64; `WERF_STAPEL_IMAGE_NAME`/`VERSION` overrides still pull.
- **`--meta-repo`** (`WERF_META_REPO`): dedicated OCI repository for werf metadata (image-metadata, managed images, custom-tag metadata, cleanup records), decoupled from the stages repo; unset means metadata stays in `--repo` as before.
- **`werf release get` / `list` / `history`** commands (with `--revisions-limit`).
- **Plan artifact** support for deploy and `bundle` commands.
- **`bundle publish` no longer requires a secret key** unless secret values are actually present.
- **Multiple imports into the same target directory** for Stapel.
- **Container backend optional for `converge`/`render`** — helm-chart-only projects run without Docker/Buildah installed.
- **`werf ci-env` auto-detects `DOCKER_AUTH_CONFIG`** to enable `--use-docker-auth-config` (explicit flag/env always wins).
- **Scratch base image**: dedicated type with proper validation (`fromLatest` rejected, import-from-scratch rejected, `"scratch"` reserved as image name).
- **Host cleanup GC for stale lock files** — prevents inode exhaustion from unbounded `~/.werf/service/locks/` growth (#6127); lock dir moved to a versioned path.
- **`werf-in-image` images ship `netavark`** so the native Buildah backend initializes out of the box in all four flavors (alpine, centos, fedora, ubuntu); `passt` is not needed there since rootless mode is rejected inside containers.
- Fixes shipped alongside: `--check-built-images`/`--require-built-images` were silently ignored by `werf build`; build-time env vars (`SSH_AUTH_SOCK`, `WERF_COMMIT_*`, `COLUMNS`) leaked into final image `ENV`; symlink import destinations resolved under root mount; base64 kubeconfig prioritized for kubectl; UTF-8 runes no longer split across parallel build log reads; git patches unified through filtered archive instead of `git apply`.

## Why

The v2 config surface accumulated redundant directives (`fromImage` vs `from`, `import.image` vs `import.from`, nameless images, implicit `:latest`) and two Stapel builders that had drifted apart. v3 collapses those to one shape per concept and makes every external image reference explicit, so the build graph is reproducible from `werf.yaml` alone.

Operationally, v3 removes werf's two external runtime dependencies for the default path: the synchronization service (content-addressable stage tags make distributed locking unnecessary) and the registry pull of the stapel image (now embedded). The fork chain (`werf/3p-docker-buildx`, `werf/3p-buildah`) existed to patch behaviors upstream now supports; dropping it removes cobra CLI emulation and the burden of keeping forks in sync.

## Review focus / risks

- **Synchronization removal** is the hardest breaking change: no deprecated no-op, `-S` simply disappears. The safety argument rests on stage tags being content-addressable — concurrent identical publishes are registry-safe. Worth confirming there is no remaining flow that relied on the distributed lock beyond stage publishing.
- **Anchor digest domain**: the content-anchor digest must cover every input that meaningfully affects image content — anything missed silently causes a false cache hit and full-image reuse of a stale image.
- **Cache identity migration**: several independent changes affect stage reuse (import content digest, anchor digest, git file mode in checksum, ancestry check removal). Each alone is safe; together the first v3 build on an existing v2 storage will partially rebuild. Release notes should set expectations.
- **Ancestry check removal**: cached git stages are now reused without verifying the cached commit is an ancestor of the current one. Confirm this cannot resurrect stale git content after a force-push/history rewrite.
- **`.global.env` removal** breaks charts referencing it; the escape hatch is an env var (`WERF_LEGACY_VALUES_GLOBAL_ENV`), not a flag — worth documenting prominently.
- **`--meta-repo` routing completeness**: every metadata call must hit meta storage, every stage call must stay on stages storage; no auto-migration of pre-existing metadata from `--repo`.
- **Giterminism strict validation + trailing-slash error** silently break previously-"working" configs. Error messages should name the offending key/path and the fix.
- **`--final-images-only=true`** skips orphan non-final images; CI relying on the old default must pass `--final-images-only=false`.
- **Fork removal (buildx/buildah)**: any behavior the forks carried that upstream lacks is gone. Highest-risk: buildah rootless UID mapping, secrets mounting.
- **buildx provenance disabled by default** — otherwise the index manifest breaks werf's label reads; users depending on provenance attestations lose them silently.
- **Embedded stapel** build-tag split (`embedstapel` vs stub): any future build entrypoint that forgets `buildGoTags` silently ships a pull-only binary.
- **`fromImage`/`import.image` are deprecation aliases, not removals** — migration docs should say "deprecated in v3", not "removed".
- **podman stack pinned below its latest release**: `go.podman.io/common` 0.69 hard-requires `pasta` >= 20240814 (it dropped the retry that used to strip `--map-guest-addr` for older binaries), which Ubuntu 24.04 does not ship, so rootless builds fail there. buildah stays at 1.44.1 with common 0.68.1, image 5.40.0 and storage 1.63.0 until the supported host baseline moves.
- **`docker/buildx` is still replaced by `werf/3p-buildx`** — a single patch (docker/buildx#4007, concurrent log pause) rebased onto upstream v0.37.1. Only `3p-buildah` and `3p-docker-buildx` are gone; this replace must be dropped once the patch lands upstream.
- **netavark/pasta host requirement**: unlike CNI/slirp4netns, netavark isn't bundled with werf and isn't universally preinstalled outside `werf-in-image` — hosts running the native Buildah backend directly (not via `werf-in-image`) need it added to their provisioning. Ubuntu 24.04 is the first LTS to package it; older distros/images need a manual install or backport.


alexey-igrychev:
## Review focus retained from the PR description
- **Synchronization removal** is the hardest breaking change: no deprecated no-op, `-S` simply disappears. The safety argument rests on stage tags being content-addressable — concurrent identical publishes are registry-safe. Worth confirming there is no remaining flow that relied on the distributed lock beyond stage publishing.
- **Anchor digest domain**: the content-anchor digest must cover every input that meaningfully affects image content — anything missed silently causes a false cache hit and full-image reuse of a stale image.
- **Cache identity migration**: several independent changes affect stage reuse (import content digest, anchor digest, git file mode in checksum, ancestry check removal). Each alone is safe; together the first v3 build on an existing v2 storage will partially rebuild. Release notes should set expectations.
- **Ancestry check removal**: cached git stages are now reused without verifying the cached commit is an ancestor of the current one. Confirm this cannot resurrect stale git content after a force-push/history rewrite.
- **`.global.env` removal** breaks charts referencing it; the escape hatch is an env var (`WERF_LEGACY_VALUES_GLOBAL_ENV`), not a flag — worth documenting prominently.
- **`--meta-repo` routing completeness**: every metadata call must hit meta storage, every stage call must stay on stages storage; no auto-migration of pre-existing metadata from `--repo`.
- **Giterminism strict validation + trailing-slash error** silently break previously-"working" configs. Error messages should name the offending key/path and the fix.
- **`--final-images-only=true`** skips orphan non-final images; CI relying on the old default must pass `--final-images-only=false`.
- **Fork removal (buildx/buildah)**: any behavior the forks carried that upstream lacks is gone. Highest-risk: buildah rootless UID mapping, secrets mounting.
- **buildx provenance disabled by default** — otherwise the index manifest breaks werf's label reads; users depending on provenance attestations lose them silently.
- **Embedded stapel** build-tag split (`embedstapel` vs stub): any future build entrypoint that forgets `buildGoTags` silently ships a pull-only binary.
- **`fromImage`/`import.image` are deprecation aliases, not removals** — migration docs should say "deprecated in v3", not "removed".
- **podman stack pinned below its latest release**: `go.podman.io/common` 0.69 hard-requires `pasta` >= 20240814 (it dropped the retry that used to strip `--map-guest-addr` for older binaries), which Ubuntu 24.04 does not ship, so rootless builds fail there. buildah stays at 1.44.1 with common 0.68.1, image 5.40.0 and storage 1.63.0 until the supported host baseline moves.
- **`docker/buildx` is still replaced by `werf/3p-buildx`** — a single patch (docker/buildx#4007, concurrent log pause) rebased onto upstream v0.37.1. Only `3p-buildah` and `3p-docker-buildx` are gone; this replace must be dropped once the patch lands upstream.
- **netavark/pasta host requirement**: unlike CNI/slirp4netns, netavark isn't bundled with werf and isn't universally preinstalled outside `werf-in-image` — hosts running the native Buildah backend directly (not via `werf-in-image`) need it added to their provisioning. Ubuntu 24.04 is the first LTS to package it; older distros/images need a manual install or backport.

The description was corrected against the current source: `--timeout` remains; only `-t` is removed, Helm path variables remain, `AnchorReused` is not a report field, and neither full offline builds nor automatic migration of existing CNI networks are guaranteed. Review notes remain here rather than in the user-facing PR body.

alexey-igrychev:
## Review → fix → re-review status

Reviewed the cumulative `main` → `3` change, including deleted tests. The following corrections are in separate PRs targeting `3`; none is merged by this review session.

| PR | Result |
| --- | --- |
| #7897 | Restores unavailable-stage error contracts and broken-image repair, removes redundant metadata tag listing, enables dormant config tests, fixes the history limit environment default and catches actual internal `from` cycles. Independent review accepted after fixing its copy-on-miss finding. Required CI is green. |
| #7898 | Keeps embedded Stapel tags independent of build-host runtime overrides; runs embedded Stapel/Deno integrity checks in CI. Independent review accepted; CI green. |
| #7899 | EN/RU migration warnings for Ansible, `docker:`, `.Values.global.env` and the removed `HELM_DRIVER` alias. Independent review required narrowing an overbroad `HELM_*` claim; corrected while retaining Helm path variables. Docs CI green. |
| #7900 | Restores `--kube-scan-namespaces` and its environment inputs, multiple-context namespace union and fail-closed handling of namespace scan errors. Independent review accepted; CI green. |
| #7901 | Restores real-stage anchor and custom-tag checks, secret CLI round trips/wrong-key/rotation coverage, and the converge → release list/get/history → rollback → plan → dismiss lifecycle. Independent review accepted. A new test's CI-only ANSI/JSON failure was reproduced locally and corrected by explicitly disabling JSON coloring, without weakening assertions. CI green. |

### Evidence and rejected findings

- All five patch branches were cherry-picked together without conflicts in a local integration worktree at `700825ce7`. The combined tree passed build, full lint, all 53 unit suites, embedded-asset tests, the content-tag/secret/release e2e selections and config integration. The tree remained clean after formatting and verification. Mutations exercised stage iteration, propagation of read errors, custom-tag presence/content, wrong-key errors, rollback/dismiss, revision selection and plan exit status.
- Existing CI scenarios pass on the fixes, including Linux CGO, e2e and integration jobs. The intermittent registry-export failure on #7897 passed when failed jobs were rerun on the same head; it was not papered over with a code change.
- External HTTP/secret-dependent inputs are longstanding cache-key limitations, not newly proven anchor regressions. Deliberate v3 breaking changes were not reverted merely because a reviewer assigned them a severity.
- The alleged `ChartFileExists` byte-read bypass has no call site in the pinned nelm dependency: it declares the interface method but does not call it. Chart reads still go through the file reader. The inspected symlink traversal implementation was unchanged from the base; a security regression was not established by that report.
- Deletion of obsolete Ansible/artifact/synchronization tests is intentional; coverage for still-supported secret/release/cache behavior was restored instead.
- `task mock:check` was executed. Its differences were generated import ordering, not missing interface methods. Attempts to normalize this without changing generated files were discarded; the existing disabled mock gate remains an explicitly known verification limitation.

### Remaining release gates — umbrella is NOT approved for main

- **Deferred by owner:** cleanup safety when adopting `--meta-repo`. A build-written routing marker does not prove old metadata was migrated. No change to this behavior or its marker format is included in these patches.
- **Release handoff:** `.release-please/main/manifest.json` is still `2.79.0`, while `.release-please/3/manifest.json` is `3.5.0`. The workflow chooses by branch. Release Please v17.6.0 matches releases against the manifest version; it does not simply choose the newest v3 tag. Before promotion, reconcile release state through the maintained release process so main does not restart from the v2 baseline or propose an already-used v3 version. Generated release metadata was not edited ad hoc.
- **Supply-chain follow-up:** embedded Stapel is still fetched from the mutable `0.7.2` tag. Artifact checksum tests verify internal consistency, not the provenance of that tag. Pinning the pull to the index digest needs a separate change/review; no unreviewed pinning mechanism was added after the advisor provider returned HTTP 429.
- Linux runtime/platform-variant and existing CNI network migration claims from the initial reports were not reproduced. Do not interpret green compilation or a partial platform check as exhaustive runtime certification. Unsupported automatic CNI-migration and fully air-gapped-build claims were removed from the PR description.

The corrective PRs can be reviewed and merged independently; this report does not authorize or perform the merge of #7377.

PR #7391, opened 2026-03-12
refactor(build): refactor method GetStageDesc and its callers

reyreavman:


pkg/storage/stages_storage.go
```diff
@@ -20,13 +20,29 @@ const (
...
+	ErrStageRejected = errors.New("stage rejected")
 )
 
 func IsErrBrokenImage(err error) bool {
 	return err != nil && strings.HasSuffix(err.Error(), ErrBrokenImage.Error())
 }
 
+func IsErrStageNotFound(err error) bool {
+	if err != nil {
+		return strings.HasSuffix(err.Error(), ErrStageNotFound.Error())
```
nervgh: Let's prefer usage of `errors.Is()` or `errors.As()` instead. So we don't need this function at all.

pkg/storage/stages_storage.go
```diff
@@ -20,13 +20,29 @@ const (
...
+	}
+	return false
+}
+
+func IsErrStageRejected(err error) bool {
+	return err != nil && strings.HasSuffix(err.Error(), ErrStageRejected.Error())
```
nervgh: Let's prefer usage of `errors.Is()` or `errors.As()` instead. So we don't need this function at all.

pkg/storage/stages_storage.go
```diff
@@ -20,13 +20,29 @@ const (
...
+}
+
+func IsErrStageUnavailable(err error) bool {
+	return IsErrStageNotFound(err) || IsErrBrokenImage(err) || IsErrStageRejected(err)
```
nervgh: For this composite function we could use combination calls of `errors.Is()` or `errors.As()`.

pkg/storage/repo_stages_storage.go
```diff
@@ -968,7 +968,7 @@ func (storage *RepoStagesStorage) PostMultiplatformImage(ctx context.Context, pr
 
 func (storage *RepoStagesStorage) CopyFromStorage(ctx context.Context, src StagesStorage, projectName string, stageID image.StageID, opts CopyFromStorageOptions) (*image.StageDesc, error) {
 	desc, err := storage.GetStageDesc(ctx, projectName, stageID)
-	if err != nil {
+	if err != nil && !IsErrStageUnavailable(err) {
 		return nil, fmt.Errorf("unable to get stage %s description: %w", stageID, err)
 	}
 	if desc != nil {
```
nervgh: Could we simplify reading and understanding of this logic?

PR #7394, opened 2026-03-12
fix(build): handle broken import metadata images in container registry

reyreavman:


pkg/build/stage/dependencies.go
```diff
@@ -342,9 +346,29 @@ func (s *DependenciesStage) getImportSourceChecksum(ctx context.Context, c Conve
 		}
 	}
 
+	if s.resolvedImportMetadata == nil {
+		s.resolvedImportMetadata = make(map[string]*storage.ImportMetadata)
```
nervgh: Should we move initialization into [constructor](https://github.com/werf/werf/blob/50faf1927a8dd3b9a058390a17ffa559ef58938f/pkg/build/stage/dependencies.go#L63) ?

pkg/build/stage/dependencies.go
```diff
@@ -342,9 +346,29 @@ func (s *DependenciesStage) getImportSourceChecksum(ctx context.Context, c Conve
...
+	}
+	s.resolvedImportMetadata[importSourceID] = importMetadata
+
 	return importMetadata.Checksum, nil
 }
 
+func (s *DependenciesStage) getResolvedImportMetadata(ctx context.Context, c Conveyor, importSourceID string) (*storage.ImportMetadata, error) {
+	if importMetadata := s.resolvedImportMetadata[importSourceID]; importMetadata != nil {
```
nervgh, line 357: What risks do you see using caching?

reyreavman: i'll try to speculate

in theory, there are several factors: 
- the first thing that comes to mind is the race condition, but each stage is a separate instance that is processed sequentially, so there should be no problems with this. I'm not sure if there's a big need to make a thread-safe map, but if you think it's worth doing, then I can implement it
- the next point is the stale cache, there are no risks, `ImportSourceID` is deterministic, related to content digest
- next, a cache miss is theoretically possible, but for such a case, a fallback is provided with a request to the registry with a warning message

I don't see any more risks, maybe I'm missing something?

pkg/build/stage/dependencies.go
reyreavman: it makes sense, in this case the code will become more explicit and the nil checks will go away

PR #7411, opened 2026-03-24
feat(cleanup): support volume usage units for host cleanup

odiora:


pkg/cleaning/threshold/threshold_test.go
```diff
@@ -0,0 +1,59 @@
+package threshold
```
nervgh, line 1: Our Taskfile [assumes](https://github.com/werf/werf/blob/99be2f470f20f31c56d05e948f8d0ad0d245b0e3/Taskfile.dist.yaml#L300) using of Ginkgo / Gomega for testing.  So, let's use Ginkgo / Gomega everywhere.

pkg/host_cleaning/local_backend_cleaner.go
```diff
@@ -30,16 +30,76 @@ var ErrUnsupportedContainerBackend = errors.New("unsupported container backend")
...
+		panic(fmt.Sprintf("mixed volume usage threshold types: threshold=%q, margin=%q; both must be the same type", threshold.Type, margin.Type))
+	}
+
+	switch threshold.Type {
+	case VolumeUsageThresholdTypePercentage:
+		return vu.BytesToFree(max(threshold.PercentageValue()-margin.PercentageValue(), 0))
+	case VolumeUsageThresholdTypeBytes:
+		return vu.BytesToFreeForTargetFreeBytes(threshold.Value + margin.Value)
+	default:
+		panic(fmt.Sprintf("unexpected volume usage threshold type %q", threshold.Type))
```
nervgh, line 67: Let's avoid runtime panics when it possible. I suggest we could derive the type while initialization.

cmd/werf/common/host_cleanup.go
```diff
@@ -22,18 +24,89 @@ func RunAutoHostCleanup(ctx context.Context, cmdData *CmdData, containerBackend
...
+		f.afterSet(f.value)
+	}
+	return nil
+}
+
+func (f *volumeUsageThresholdFlag) Type() string {
+	return "volume-usage"
+}
+
+func initVolumeUsageThresholdTarget(target **thresholdpkg.Threshold, defaultValue string) {
```
nervgh: Why do you use double pointers (**) ? What is the reason?

docs/_includes/reference/cli/werf_cleanup.md
```diff
@@ -27,22 +27,24 @@ werf cleanup [options]
...
-            level (default 5% or $WERF_ALLOWED_BACKEND_STORAGE_VOLUME_USAGE_MARGIN)
+            Set the cleanup margin for backend (Docker or Buildah) storage. In percentage mode the  
+            margin is subtracted from the usage threshold, while in bytes mode it is added to the   
+            minimum required free space threshold, e.g. 5 or 2GB (default 5 or                      
+            $WERF_ALLOWED_BACKEND_STORAGE_VOLUME_USAGE_MARGIN)
       --allowed-local-cache-volume-usage=70
-            Set allowed percentage of local cache (~/.werf/local_cache by default) volume usage     
-            which will cause cleanup of least recently used data from the local cache (default 70%  
-            or $WERF_ALLOWED_LOCAL_CACHE_VOLUME_USAGE)
+            Set the cleanup threshold for local cache (~/.werf/local_cache by default). Plain
```
nervgh, line 42: Let's keep clear our docs. It can confuse our user if a param turns from  "the maximum allowed usage percentage" to "minimum required free space threshold".

nervgh (changes requested):

pkg/cleaning/threshold/threshold.go
```diff
@@ -0,0 +1,107 @@
...
+		return NewPercentage(number), nil
+	}
+
+	bytesValue, err := humanize.ParseBytes(value)
+	if err != nil {
+		return Threshold{}, fmt.Errorf("parse volume usage threshold %q: %w", value, err)
+	}
+	if bytesValue == 0 {
+		return Threshold{}, fmt.Errorf("bytes threshold %q should be greater than 0", value)
+	}
```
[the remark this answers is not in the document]
odiora: This is partially valid, but the strict parser is intentional. 0B is currently used as an implicit effective margin for bytes mode when the margin is omitted. We intentionally do not allow explicit 0B from CLI/env yet, to keep the generic parser strict. If explicit 0B should be supported later, it would be better to add that intentionally for margin handling rather than silently loosening the generic threshold parser.

pkg/host_cleaning/local_backend_cleaner.go
odiora: This is a fair defensive-programming concern. At the moment these panics only guard internal invariant paths: after successful parse/resolve and type-consistency checks, those branches should not be reachable from user input. So for this PR I kept them as internal invariant guards rather than expanding explicit error plumbing across multiple helper layers. If the team wants to remove such panics consistently from internal helpers, that can be done as a separate small refactor.

cmd/werf/common/host_cleanup.go
odiora: The double pointer is needed not for mutating the value itself, but for initializing the pointer field in CmdData. The helper must be able to: (1) detect that the target is not yet allocated, (2) allocate a Threshold, and (3) write that pointer back into the caller field. It also lets the main flag and aliases share the same target and avoids the pointer overwrite bug. A short code comment was also added near initVolumeUsageThresholdTarget(...) to make this less surprising in review.

docs/_includes/reference/cli/werf_cleanup.md
odiora: This is a valid UX concern. We intentionally did not change the runtime semantics, because bytes mode as minimum required free space is operationally much more useful for cleanup. Instead, we improved the wording in help and docs so that the dual semantics are stated explicitly: plain numbers define the maximum allowed usage percentage, while values with units define the minimum required free space.

alexey-igrychev:
resolved by https://github.com/werf/werf/pull/7467

PR #7459, opened 2026-04-09
refactor(host-cleanup): switch internal cleanup logic to bytes

nervgh:
Transition the host cleaning (GC) engine from percentage-based logic to absolute byte values (uint64). This unified approach simplifies space calculations across different storage backends and prepares the codebase for upcoming absolute unit support (e.g., GB, MB) in CLI flags.

- Replace percentage fields with byte-based limits in RunGCOptions.
- Update LocalBackendCleaner and gitdata manager to operate on bytes.
- Implement percentage-to-bytes conversion at the entry point in RunHostCleanup to maintain backward compatibility.
- Simplify images cleanup loop and recalculate disk usage only when necessary.
- Add PercentageToBytes helper to VolumeUsage utility.


pkg/host_cleaning/local_backend_cleaner.go
```diff
@@ -315,12 +316,12 @@ func (cleaner *LocalBackendCleaner) RunGC(ctx context.Context, options RunGCOpti
...
 	}
 
-	if vu.Percentage() <= options.AllowedStorageVolumeUsagePercentage {
+	if vu.UsedBytes <= options.AllowedStorageVolumeUsageBytes {
 		logboek.Context(ctx).LogBlock("Check storage").Do(func() {
 			logboek.Context(ctx).LogF("Total freed space: %s\n", utils.RedF("%s", humanize.Bytes(vuBefore.UsedBytes-vu.UsedBytes)))
 			logboek.Context(ctx).LogF("Volume usage: %s / %s\n", humanize.Bytes(vu.UsedBytes), humanize.Bytes(vu.TotalBytes))
-			logboek.Context(ctx).LogF("Allowed percentage level exceeded: %s > %s — %s\n", utils.RedF("%0.2f%%", vu.Percentage()), utils.YellowF("%0.2f%%", options.AllowedStorageVolumeUsagePercentage), utils.RedF("HIGH VOLUME USAGE"))
-			logboek.Context(ctx).LogF("Target percentage level after cleanup: %0.2f%% - %0.2f%% (margin) = %s\n", options.AllowedStorageVolumeUsagePercentage, options.AllowedStorageVolumeUsageMarginPercentage, utils.BlueF("%0.2f%%", targetVolumeUsagePercentage))
+			logboek.Context(ctx).LogF("Allowed level exceeded: %s > %s — %s\n", utils.RedF("%s", humanize.Bytes(vu.UsedBytes)), utils.YellowF("%s", humanize.Bytes(options.AllowedStorageVolumeUsageBytes)), utils.RedF("HIGH VOLUME USAGE"))
```
reyreavman, line 323: the log message says "Allowed level exceeded: ... — HIGH VOLUME USAGE" (in red), but it's inside the if vu.UsedBytes <= options.AllowedStorageVolumeUsageBytes block  — meaning usage is actually within the allowed limit 
the message should reflect that cleanup succeeded and usage is now OK, not that the limit was exceeded, nope?

reyreavman (changes requested):

pkg/host_cleaning/local_backend_cleaner.go
nervgh: We have two similar log messages:
1. [Allowed volume usage ...](https://github.com/werf/werf/pull/7459/changes#diff-2338ecf63db7d4134bed95a392b2d11601c820e51c8c4323bd7b2910cffcbea6R104) inside of `if vu.UsedBytes <= options.AllowedLocalCacheVolumeUsageBytes {` condition
2. [Allowed level exceeded ...](https://github.com/werf/werf/pull/7459/changes#diff-2338ecf63db7d4134bed95a392b2d11601c820e51c8c4323bd7b2910cffcbea6R116) beyond of that condition

So it 's fine.

reyreavman (approved):

alexey-igrychev (approved):

PR #7467, opened 2026-04-13
feat(host-cleanup): add support for absolute storage units

nervgh:
- Implement universal units parser supporting both percentages and absolute sizes (GB, MiB, etc.).
- Refactor CmdData to use units as a pflag.Value, eliminating redundant string/uint pointer fields.
- Add strict unit consistency validation within cleanup groups (backend storage and local cache).
- Update pkg/host_cleaning to handle thresholds in bytes.
- Document absolute units support and unit consistency rules in both English and Russian.
- Add table-driven Ginkgo tests for storage value parsing and threshold calculation.


pkg/host_cleaning/units/units.go
reyreavman, line 1: don't you think it's worth moving `units.UnitValue` to `pkg/` (for example, `pkg/units/` or directly to `pkg/volumeutils/`)?

reyreavman: i mean `pkg/host_cleaning imports cmd/werf/common/units` — this is an inversion of layers, packages in `pkg/` are library code, they should not depend on `cmd/`

reyreavman (approved):

pkg/host_cleaning/units/units.go
nervgh: I've moved to `pkg/host_cleaning/units` because `units` related with host cleanup.

reyreavman (approved):

alexey-igrychev (approved):

PR #7480, opened 2026-04-16
fix(build, docker, stapel): fix stage image cache for multi-platform builds

Fral738:


pkg/build/conveyor.go
reyreavman: lets make this method private and document the multiplatform build case - in this case `GetStageImage` will return `nil`

pkg/build/conveyor_ai_test.go
```diff
@@ -0,0 +1,40 @@
...
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	"github.com/werf/werf/v2/pkg/build/stage"
+	"github.com/werf/werf/v2/pkg/container_backend"
+)
+
+func TestAI_Conveyor_StageImageCacheSeparatesPlatforms(t *testing.T) {
```
reyreavman: let's migrate test suite to `ginkgo/gomega`

reyreavman: also, the current set of tests covers cache primitives but does not cover the script itself, which was broken, let's add it

reyreavman (changes requested):

reyreavman (approved):

PR #7492, opened 2026-04-22
fix(build): resolve dependency image refs in FROM stage before computing cache digest

reyreavman:


nervgh (approved):

pkg/build/build_phase.go
```diff
@@ -633,6 +633,16 @@ func (phase *BuildPhase) onImageStage(ctx context.Context, img *image.Image, stg
 		}
 	}
 
+	if img.IsDockerfileImage && img.DockerfileImageConfig.Staged {
```
nervgh, line 636: This code block looks like duplication of that https://github.com/werf/werf/blob/4927080cc11ffee7e928e9e45f094389518949ff/pkg/build/build_phase.go#L746

On one level above we have that https://github.com/werf/werf/blob/4927080cc11ffee7e928e9e45f094389518949ff/pkg/build/build_phase.go#L604 It means `afterImageStage` is already called inside of `onImageStage`. It looks strange.

reyreavman: two `if img.IsDockerfileImage && img.DockerfileImageConfig.Staged` blocks in `onImageStage` and `afterImageStage` look like duplication but serve different purposes at different points in the stage lifecycle:

     - `onImageStage`: expands dependency refs (e.g. `${SSR_IMAGE}` → `registry/repo:digest`) BEFORE `calculateStage`, so the cache digest is computed from the resolved image reference
     - `afterImageStage`: saves the base env from the built FROM image AFTER `calculateStage` + build, so subsequent stages can use it for their own expansion

these cant be merged because `calculateStage` and the build sit between them

to eliminate the visual duplication, `afterImageStage` can be inlined into the end of `onImageStage` — both blocks would then live in one function, making the before/after relationship explicit

nervgh (approved):

alexey-igrychev (approved):

PR #7493, opened 2026-04-22
fix(build, stapel): handle platform mismatch in stage base image resolution and avoid panic

Fral738:


pkg/stapel/container.go
```diff
@@ -15,21 +16,53 @@ type container struct {
...
+				if err != nil {
+					return err
+				}
+
+				actualPlatform := fmt.Sprintf("%s/%s", inspect.Os, inspect.Architecture)
+				if inspect.Variant != "" {
+					actualPlatform = fmt.Sprintf("%s/%s", actualPlatform, inspect.Variant)
+				}
+
+				platformMatches := targetPlatform == actualPlatform || strings.HasPrefix(targetPlatform, actualPlatform+"/") || strings.HasPrefix(actualPlatform, targetPlatform+"/")
```
reyreavman: this design looks extremely fragile

pkg/build/conveyor.go
```diff
@@ -1029,17 +1046,34 @@ func (c *Conveyor) GetOrCreateStageImage(name string, prevStageImage *stage.Stag
 }
 
 func (c *Conveyor) GetImage(targetPlatform, name string) *image.Image {
+	img, err := c.FindImage(targetPlatform, name)
+	if err != nil {
+		panic(err.Error())
```
reyreavman, line 1049: i think we don't need to panic, but return some kind of relevant error

reyreavman (changes requested):
in addition to the changes, i kindly ask you to add e2e and unit tests to cover the new logic

pkg/build/conveyor.go
Fral738: Agreed in principle. `GetImage` is a legacy wrapper over `FindImage` that already returns a proper error. The reason it still panics is that its signature `*Image` is part of the `stage.Conveyor` interface, which has ~15 call sites  — all of which don't expect an error return. Changing this requires updating the interface and all callers in sync, which is a non-trivial refactor beyond the scope of this PR. I'll leave a TODO and do it as a follow-up.

reyreavman (approved):

pkg/build/build_phase.go
```diff
@@ -651,6 +651,10 @@ func (phase *BuildPhase) onImageStage(ctx context.Context, img *image.Image, stg
 		DoError(func() error {
 			var err error
 			foundSuitableStage, calculateStageCleanupFunc, err = phase.calculateStage(ctx, img, stg)
+			if err != nil && calculateStageCleanupFunc != nil {
```
nervgh, line 654: We already have the cleanup call [there](https://github.com/werf/werf/pull/7493/changes#diff-ea9b317bcabdca97716079f76aa2a660f8097031e01d4798ee5bb6e7dd3cb231R666). Why do we need one more?

pkg/container_backend/legacy_stage_image_container.go
```diff
@@ -90,10 +90,15 @@ func (c *LegacyStageImageContainer) prepareRunArgs(ctx context.Context) ([]strin
 	setColumnsEnv := fmt.Sprintf("--env=COLUMNS=%d", logboek.Context(ctx).Streams().ContentWidth())
 	runArgs = append(runArgs, setColumnsEnv)
 
-	fromImageId := c.image.fromImage.GetID()
+	var fromImageRef string
```
nervgh: Let's reduce code duplication into a private function

pkg/stapel/container.go
```diff
@@ -15,21 +18,60 @@ type container struct {
 	Name      string
 	ImageName string
 	Volume    string
+	Platform  string
 }
 
 func (c *container) Create(ctx context.Context) error {
-	name := fmt.Sprintf("--name=%s", c.Name)
-	volume := fmt.Sprintf("--volume=%s", c.Volume)
+	imageLockName := fmt.Sprintf("stapel.image.%s", strings.NewReplacer("/", "_", ":", "_", "@", "_").Replace(c.ImageName))
```
nervgh: Let's move formatting of imageLockName into standalone private testable (!) helper.

pkg/stapel/container.go
```diff
@@ -15,21 +18,60 @@ type container struct {
...
+
+			if err := docker.CliPullWithRetries(ctx, pullArgs...); err != nil {
+				return err
+			}
 		}
-	}
 
-	return docker.CliCreate(ctx, name, volume, c.ImageName)
+		return docker.CliCreate(ctx, name, volume, c.ImageName)
+	})
```
nervgh, line 75: We already have `CreateIfNotExist` method. Should it support "multiple platforms logic" too?

pkg/stapel/stapel.go
```diff
@@ -57,11 +65,31 @@ func GetOrCreateContainer(ctx context.Context) (string, error) {
 }
 
 func Purge(ctx context.Context) error {
-	container := getContainer()
-	if err := container.RmIfExist(ctx); err != nil {
+	baseContainerName := fmt.Sprintf("%s%s", image.AssemblingContainerNamePrefix, getVersion())
+	containers, err := docker.Containers(ctx, types.ContainerListOptions{All: true})
```
nervgh: `types.ContainerListOptions` allows to specify `Filters` to filter containers by names. Example https://github.com/werf/werf/blob/9d0e8accb465e5566da7092ec77771359a4dc019/pkg/host_cleaning/containers.go#L15 Could we filter that way?

test/pkg/suite_init/werf_init_data.go
```diff
@@ -0,0 +1,25 @@
...
+)
+
+type WerfInitData struct{}
+
+func NewWerfInitData(tmpDirData *TmpDirData) *WerfInitData {
+	data := &WerfInitData{}
+	BeforeEach(func() {
+		homeDir, err := os.MkdirTemp("", "werf-test-home-")
+		Expect(err).To(Succeed())
+		DeferCleanup(func() {
```
nervgh: Ginkgo creates temp  directory per suite and removes it after execution. 

Could use this knowledge to delegate cleanup to Ginkgo?

nervgh (changes requested):

pkg/stapel/container.go
odiora: CreateIfNotExist() already applies the new platform-aware logic on the creation path because it delegates to Create(). If we also want to validate and recreate already existing stapel containers when their image/platform is stale, that would be a broader behavior change. I would keep it as a separate follow-up if we want reconciliation semantics.

pkg/build/build_phase.go
odiora: The deferred cleanup below is only registered after DoError(...) returns successfully. On the error path we return before reaching that defer, so the cleanup returned from calculateStage() would never run. Here it is the stage lock release, so this extra call covers only the error path.

pkg/stapel/stapel.go
```diff
@@ -57,11 +66,34 @@ func GetOrCreateContainer(ctx context.Context) (string, error) {
...
+	})
+	if err != nil {
 		return err
 	}
 
+	processed := map[string]struct{}{}
+	for _, ctr := range containers {
+		for _, containerName := range ctr.Names {
+			containerName = strings.TrimPrefix(containerName, "/")
+			if containerName != baseContainerName && !strings.HasPrefix(containerName, baseContainerName+"_") {
```
nervgh, line 82: I think we need the function like `stapel.IsStapelContainerName(ctrName)` which will return boolean value. We would use that function here ^ and [there](https://github.com/werf/werf/blob/9d0e8accb465e5566da7092ec77771359a4dc019/pkg/host_cleaning/containers.go#L45).

pkg/stapel/stapel.go
```diff
@@ -57,11 +66,34 @@ func GetOrCreateContainer(ctx context.Context) (string, error) {
...
+				continue
+			}
+
+			if _, done := processed[containerName]; done {
```
nervgh, line 86: Why do we need the `processed` map?

pkg/stapel/stapel.go
```diff
@@ -57,11 +66,34 @@ func GetOrCreateContainer(ctx context.Context) (string, error) {
 }
 
 func Purge(ctx context.Context) error {
-	container := getContainer()
-	if err := container.RmIfExist(ctx); err != nil {
+	baseContainerName := fmt.Sprintf("%s%s", image.AssemblingContainerNamePrefix, getVersion())
```
nervgh, line 69: The `image.AssemblingContainerNamePrefix` is shared between stapel (here ^ ) and [host cleanup](https://github.com/werf/werf/blob/9d0e8accb465e5566da7092ec77771359a4dc019/pkg/host_cleaning/local_purger.go#L29).

I suggest to reduce code duplication and unify this logic with single function.

nervgh (changes requested):

nervgh (approved):

PR #7517, opened 2026-04-29
feat(cleanup): add --kube-scan-namespaces support for in-cluster scan

Fral738:


pkg/cleaning/cleanup.go
```diff
@@ -391,18 +391,44 @@ AppendNewImages:
 func (m *cleanupManager) deployedDockerImages(ctx context.Context) ([]*DeployedDockerImage, error) {
 	var deployedDockerImages []*DeployedDockerImage
 	for _, contextClient := range m.KubernetesContextClients {
-		if err := logboek.Context(ctx).LogProcessInline("Getting deployed docker images (context %s)", contextClient.ContextName).
-			DoError(func() error {
-				contextDeployedImages, err := allow_list.DeployedDockerImages(ctx, contextClient.Client, m.KubernetesNamespaceRestrictionByContext[contextClient.ContextName])
-				if err != nil {
-					return fmt.Errorf("cannot get deployed imagesStageList: %w", err)
-				}
+		scanNamespace := func(namespace string) error {
```
nervgh: Let's move the function outside of `for _, contextClient := range m.KubernetesContextClients` loop.

nervgh (changes requested):

nervgh (approved):

cmd/werf/common/cleanup_namespaces_scan.go
```diff
@@ -52,19 +58,37 @@ func GetKubernetesContextClients(configPath, configDataBase64 string, configPath
...
+			res[contextClient.ContextName] = slices.Clone(scanNamespaces)
+			continue
+		}
+
 		if *cmdData.ScanContextNamespaceOnly {
-			res[contextClient.ContextName] = contextClient.ContextNamespace
+			if contextClient.ContextNamespace != "" {
+				res[contextClient.ContextName] = []string{contextClient.ContextNamespace}
+			} else {
+				res[contextClient.ContextName] = nil
```
reyreavman, line 78: when no `--kube-scan-namespaces` and no `--scan-context-namespace-only` are set, this line still restricts scanning to the context's namespace only, rather than all namespaces and cleanup will miss images in other namespaces, is it ok?

cmd/werf/common/cleanup_namespaces_scan.go
```diff
@@ -15,6 +16,11 @@ func SetupScanContextNamespaceOnly(cmdData *CmdData, cmd *cobra.Command) {
 	cmd.Flags().BoolVarP(cmdData.ScanContextNamespaceOnly, "scan-context-namespace-only", "", util.GetBoolEnvironmentDefaultFalse("WERF_SCAN_CONTEXT_NAMESPACE_ONLY"), "Scan for used images only in namespace linked with context for each available context in kube-config (or only for the context specified with option --kube-context). When disabled will scan all namespaces in all contexts (or only for the context specified with option --kube-context). (Default $WERF_SCAN_CONTEXT_NAMESPACE_ONLY)")
```
reyreavman: help text for `--scan-context-namespace-only` says "When disabled will scan all namespaces" - but that is no longer true

reyreavman (changes requested):

reyreavman (approved):

alexey-igrychev (approved):

PR #7519, opened 2026-04-30
fix(buildah): panic in `host purge`

reyreavman:


pkg/host_cleaning/local_purger.go
```diff
@@ -98,6 +99,10 @@ func (purger *localPurger) PurgeStapelFiles(ctx context.Context, options CommonO
 		return nil
 	}
 
+	if !docker.IsEnabled() {
```
nervgh: I suggest:
1. To add `purger.containerBackednType` filed
2. To reuse [logic](https://github.com/werf/werf/blob/0f545d8bef7cb1cf8123f5c1248e3b75b31c9698/pkg/host_cleaning/local_backend_cleaner.go#L68) from cleaner (and move that logic into private function)

nervgh (changes requested):

nervgh (approved):

alexey-igrychev (approved):

PR #7530, opened 2026-05-11
fix(ci-env): ci-env ignores session docker config when WERF_DOCKER_CONFIG is set

reyreavman:


nervgh (approved):

test/e2e/ci-env/simple_test.go
```diff
@@ -68,4 +69,55 @@ var _ = Describe("Simple ci-env", Label("e2e", "ci-env", "simple"), func() {
...
+			}
+
+			Expect(envs).To(HaveKey("DOCKER_CONFIG"))
+			Expect(envs).To(HaveKey("WERF_DOCKER_CONFIG"))
+			Expect(envs["WERF_DOCKER_CONFIG"]).To(Equal(envs["DOCKER_CONFIG"]))
+		},
+		Entry(
+			"gitlab --as-env-file",
+			[]string{"gitlab", "--as-env-file"},
+			func(line string) (string, string, bool) {
```
nervgh: Code block 109-111 duplicates 117-120.

nervgh (changes requested):

test/e2e/ci-env/simple_test.go
nervgh: By the way, we already have `github.com/joho/godotenv` lib as a dependency. Could we use https://pkg.go.dev/github.com/joho/godotenv@v1.5.1#Parse from there for parsing? I mean instead of our custom function.

nervgh (changes requested):

nervgh (approved):

PR #7536, opened 2026-05-14
fix(deploy): use action.SetupColorLevel instead of force `ColorMode=on` from ci-env

dmmordvi:


cmd/werf/common/common.go
```diff
@@ -1499,11 +1500,8 @@ func GetNelmLogLevel(cmdData *CmdData) log.Level {
 
 func ProcessLogColorMode(cmdData *CmdData) error {
 	switch logColorMode := *cmdData.LogColorMode; logColorMode {
-	case "auto":
-	case "on":
-		logboek.Streams().EnableStyle()
-	case "off":
-		logboek.Streams().DisableStyle()
+	case "auto", "on", "off":
```
ilya-lesikov: Make sure those are the exact names. Also, use the constants from Nelm here

ilya-lesikov (approved):

PR #7543, opened 2026-05-19
feat(deploy): add plan artifact support for bundle command

drey:


docs/_data/sidebars/documentation.yml
```diff
@@ -698,9 +698,6 @@ entries:
               - title: Parametrize templates
                 url: /usage/deploy/values.html
 
-              - title: TypeScript templates
```
ilya-lesikov, line 701: I guess we forgot to update _documentation.yml. Can you please update _documentation.yml?

The documentation.yml file is a generated one.

ilya-lesikov (approved):

docs/_data/sidebars/documentation.yml
drey: Do you mean running `task doc:gen`? I did it before review request. The second run does not produce any diffs.

PR #7548, opened 2026-05-21
fix(build): recover from missing content blobs on push

reyreavman:


pkg/container_backend/docker_server_backend.go
```diff
@@ -302,7 +302,23 @@ func (backend *DockerServerBackend) Tag(ctx context.Context, ref, newRef string,
...
+func (backend *DockerServerBackend) pushWithRecovery(ctx context.Context, ref string) error {
+	if err := docker.CliPushWithRetries(ctx, ref); err != nil {
+		if docker.IsErrContentNotFound(err) {
+			logboek.Context(ctx).Warn().LogF("WARNING: push failed due to missing content blob %s, pulling %s to materialize layers\n", docker.ContentNotFoundDigest(err), ref)
+
+			if pullErr := docker.CliPullWithRetries(ctx, ref); pullErr != nil {
+				return fmt.Errorf("push failed (content not found), pull to recover also failed: %w (original push error: %s)", pullErr, err)
+			}
+
+			return docker.CliPushWithRetries(ctx, ref)
```
nervgh: Why do we use this approach: CliPush -> err -> pull -> CliPush ?

nervgh (changes requested):

pkg/container_backend/docker_server_backend.go
reyreavman: - the push -> fail -> pull -> push approach is a reactive recovery that treats the missing blob error as a signal of an incomplete local state
- instead of preemptively verifying layers or hardcoding environment-specific checks, we use a standard pull to force `containerd` to materialize missing compressed blobs only when a push fails
- this ensures minimal overhead in healthy scenarios while providing an automated, platform-agnostic fix for `containerd-snapshotter's` optimizations or discard policies

nervgh: If we need the missing blob locally, why do we need to push after pulling successfully?

reyreavman: because the user's original goal was to push the image, not just fix the local cache =)

the pull is only a corrective step to restore missing data, once the pull finishes, the image is finally "whole" on our fs, but it's still not in the remote registry, we must perform the second push to actually complete the task the user asked for uploading the image

pkg/container_backend/docker_server_backend.go
```diff
@@ -206,6 +206,20 @@ func (backend *DockerServerBackend) BuildDockerfileStage(ctx context.Context, ba
 	return "", fmt.Errorf("staged Dockerfile is not available for Docker Server backend")
 }
 
+func (backend *DockerServerBackend) EnsureImageContent(ctx context.Context, ref string, opts EnsureImageContentOpts) error {
+	if stream, err := backend.SaveImageToStream(ctx, ref); err != nil {
```
nervgh, line 210: `SaveImageToStream` is an expensive operation because it creates new tar archive in RAM on the fly. Is it OK for us?

nervgh (changes requested):

PR #7550, opened 2026-05-21
feat(giterminism): do not fetch submodules for local repo

drey:


drey:
Submodules require initialization after worktree creation, so this PR solves nothing.

PR #7551, opened 2026-05-22
refactor(build): set current commit label for dependency stages

odiora:
Set the project repo commit label when building dependency stages so freshly rebuilt stages report the current head commit. Before this change dependency stages inherited the label from their base stage, which made rebuilt images log the previous stage commit even after storing a new image.

### DIFF
--- a/pkg/build/stage/base.go
+++ b/pkg/build/stage/base.go
@@ -365,15 +365,13 @@ func (s *BaseStage) PrepareImage(ctx context.Context, c Conveyor, cb container_b
 	return nil
 }
 
-func (s *BaseStage) addProjectRepoCommitLabel(ctx context.Context, c Conveyor, cb container_backend.ContainerBackend, stageImage *StageImage) map[string]string {
+func (s *BaseStage) addProjectRepoCommitLabel(ctx context.Context, c Conveyor, cb container_backend.ContainerBackend, stageImage *StageImage) {
 	addLabels := map[string]string{image.WerfProjectRepoCommitLabel: c.GiterminismManager().HeadCommit(ctx)}
 	if c.UseLegacyStapelBuilder(cb) {
 		stageImage.Builder.LegacyStapelStageBuilder().Container().ServiceCommitChangeOptions().AddLabel(addLabels)
 	} else {
 		stageImage.Builder.StapelStageBuilder().AddLabels(addLabels)
 	}
-
-	return addLabels
 }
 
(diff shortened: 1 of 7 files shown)


reyreavman (changes requested):
current approach relies on manual propagation which is fragile and prone to regressions, i think we should move it from "manual intervention" to lifecycle automation
                                                                                    
the goal is to ensure that common metadata like commit labels is applied automatically by the base stage rather than requiring each implementation to remember the call, this eliminates the risk of human error in future stages, removes redundant boilerplate and keeps the stage logic focused strictly on its unique tasks

pkg/build/stage/base.go
```diff
@@ -365,15 +365,13 @@ func (s *BaseStage) PrepareImage(ctx context.Context, c Conveyor, cb container_b
 	return nil
 }
 
-func (s *BaseStage) addProjectRepoCommitLabel(ctx context.Context, c Conveyor, cb container_backend.ContainerBackend, stageImage *StageImage) map[string]string {
+func (s *BaseStage) addProjectRepoCommitLabel(ctx context.Context, c Conveyor, cb container_backend.ContainerBackend, stageImage *StageImage) {
 	addLabels := map[string]string{image.WerfProjectRepoCommitLabel: c.GiterminismManager().HeadCommit(ctx)}
```
reyreavman: lets add a check that `headCommit` is not empty, this prevents overwriting valid inherited labels with an empty string in case git metadata is missing or an error occurs during commit retrieval

pkg/build/stage/from.go
```diff
@@ -95,7 +95,8 @@ func (s *FromStage) GetDependencies(_ context.Context, c Conveyor, _ container_b
 }
 
 func (s *FromStage) PrepareImage(ctx context.Context, c Conveyor, cb container_backend.ContainerBackend, prevBuiltImage, stageImage *StageImage, _ container_backend.BuildContextArchiver) error {
-	addLabels := s.addProjectRepoCommitLabel(ctx, c, cb, stageImage)
+	addLabels := map[string]string{imagePkg.WerfProjectRepoCommitLabel: c.GiterminismManager().HeadCommit(ctx)}
```
reyreavman: same here

reyreavman (changes requested):

pkg/build/stage/base.go
reyreavman: i was referring to line 369...

reyreavman (approved):

alexey-igrychev (approved):

PR #7575, opened 2026-06-11
feat(cleaning): delete rejected stages and linked custom tags during cleanup/purge

alexey-igrychev:
## Summary

Add a new cleanup/purge phase that unconditionally deletes all `*-rejected` stage tags from the registry together with any custom tags pointing to the same stage ID. Also add a broken-image overwrite fallback so that corrupted tags no longer block cleanup.

## Key changes

- **`StagesStorage` interface**: two new methods — `GetRejectedStageIDs` (scans repo tags for `*-rejected` suffix) and `DeleteRejectedStage` (deletes the rejected marker, with broken-image fallback).
- **`RepoStagesStorage`**: implements both methods. New helper `deleteRepoImageWithBrokenFallback`: on `IsBrokenImageError` overwrites the tag with a manifest-only dummy via `PushImage`, then retries `DeleteRepoImage`. `DeleteStage` now also uses this fallback for both the main stage tag and its rejected marker.
- **`LocalStagesStorage`**: no-op stubs (mirrors existing `RejectStage` stub).
- **`StorageManager`**: new `ForEachDeleteRejectedStage` — parallel deletion mirroring `ForEachDeleteStage`.
- **`cleanupManager.run()`** (`pkg/cleaning/cleanup.go`): new `cleanupRejectedStages` phase runs before `cleanupUnusedStages`. Deletes linked custom tags first, then rejected markers; forgotten from stage manager state so subsequent phases skip them.
- **`purgeManager.run()`** (`pkg/cleaning/purge.go`): new `purgeRejectedStages` phase after `deleteCustomTags`; reuses the shared `cleanupRejectedStages` helper.
- **Mocks**: regenerated via `task mock:generate`.
- **Tests**: 11 new `TestAI_*` tests — 6 for storage layer (broken-image fallback scenarios, `GetRejectedStageIDs`) and 6 for orchestration (`cleanupRejectedStages` — dry-run, linked custom tags, error propagation).

## Why

Rejected-stage markers (`repo:digest-creationTs-rejected`) created by `RejectStage()` accumulated indefinitely. They were only cleaned up when the parent stage was deleted by `DeleteStage()`, leaving orphaned markers when the parent was already gone. Additionally, broken/corrupt tags (manifest errors, unknown blobs) caused `DeleteRepoImage` to fail with no recovery path, permanently blocking cleanup for affected repos.

## Review focus / risks

- **Phase ordering**: `cleanupRejectedStages` runs before `cleanupUnusedStages` so that the stage manager's custom-tag map is mutated (via `ForgetCustomTagsByStageID`) before the unused-stages pass runs. Changing this order would cause double-deletion attempts on custom tags.
- **Broken-image fallback**: the dummy overwrite uses `PushImage` which internally calls `NewManifestOnlyImage` — same mechanism as `RejectStage`. After push, info is re-fetched because the digest changes; if the tag disappears between push and re-fetch (some registries GC immediately) the function treats it as successfully deleted.
- **`LocalStagesStorage` stubs**: return `nil, nil` / `nil` — safe because local storage has no concept of rejected tags.
- **Large generated diff in `test/mock/stages_storage.go`**: pure codegen, no manual logic.


alexey-igrychev:
Superseded by new PR based on origin/main

PR #7589, opened 2026-06-18
fix(import): exclude system directories

Fral738:


alexey-igrychev:
rel https://github.com/werf/werf/pull/7590

PR #7635, opened 2026-07-08
feat(telemetry): add OpenTelemetry (OTLP) metrics push exporter

glatinone:
While evaluating werf's behavior within automated and ephemeral CI/CD environments, we identified that a push-based telemetry framework is useful for capturing basic run metrics before the short-lived CLI process terminates.

Architectural approach:
- Introduces an OTLP metrics push exporter using the standard `otlpmetrichttp` driver.
- Isolated in `pkg/telemetry`; falls back to a safe No-Op path when disabled. `InitMetrics` returns a `*MetricsHandle` (no package-level mutable state), so the pipeline can be unit tested directly.
- Wired into the existing `TelemetryPreRun`/`ShutdownTelemetry` hooks; a forced flush happens at shutdown (no periodic interval, since the process is short-lived).

Changes:
- `pkg/telemetry/otlp_metrics.go`: MeterProvider init + the two emitted instruments.
- Root CLI flags `--otlp-metrics-enabled` (bool, default false) and `--otlp-metrics-endpoint` (string), each with a `WERF_OTLP_METRICS_ENABLED` / `WERF_OTLP_METRICS_ENDPOINT` env fallback like other werf flags. Renamed from the original `--telemetry-enabled`/`--telemetry-otlp-endpoint` to avoid implying this controls werf's existing, separate `WERF_TELEMETRY` mechanism.
- Endpoint is validated (scheme must be `http`/`https`, host required) before the exporter is constructed.
- Tests covering the No-Op path, endpoint validation, and the full start/end/shutdown lifecycle.

What is actually emitted (scoped down from the original description):
- `werf_runs` (counter) and `werf_run_duration_seconds` (histogram), tagged with `command`, `os`, `arch`, `exit_code`.
- Only unauthenticated OTLP HTTP collectors are supported right now (e.g. an OpenTelemetry Collector in front of Prometheus/Grafana). Datadog/Honeycomb-style endpoints that require an API-key header are **not** supported yet — no `otlpmetrichttp.WithHeaders` wiring exists. Happy to add a headers option in a follow-up if there's interest.

This revision addresses the review feedback: fixed the otel v1.24 API mismatch (branch didn't compile before), promoted `otlpmetrichttp`/`sdk/metric` to direct `go.mod` dependencies, stopped discarding errors, and renamed the flags per the naming-collision concern.

### DIFF
--- a/cmd/werf/common/telemetry.go
+++ b/cmd/werf/common/telemetry.go
@@ -39,6 +39,11 @@ func InitTelemetry(ctx context.Context) {
 func ShutdownTelemetry(ctx context.Context, exitCode int) {
 	telemetry.GetTelemetryWerfIO().CommandExited(ctx, exitCode)
 
+	telemetry.MetricsEnd(ctx, exitCode)
+	if err := telemetry.MetricsShutdown(ctx); err != nil {
+		telemetry.LogF("unable to shutdown metrics: %s", err)
+	}
+
 	if err := telemetry.Shutdown(ctx); err != nil {
 		telemetry.LogF("unable to shutdown: %s", err)
 	}
@@ -56,7 +61,18 @@ func TelemetryPreRun(cmd *cobra.Command, args []string) error {
 	}
 
 	InitTelemetry(ctx)
+
+	// Initialize OTLP metrics push exporter if enabled via flags
+	if root := cmd.Root(); root != nil {
+		enabled, _ := root.PersistentFlags().GetBool("telemetry-enabled")
+		endpoint, _ := root.PersistentFlags().GetString("telemetry-otlp-endpoint")
+		if err := telemetry.InitMetrics(ctx, enabled, endpoint); err != nil {
+			telemetry.LogF("error: %s", err)
+		}
+	}
+
 	telemetry.GetTelemetryWerfIO().SetCommand(ctx, command)
+	telemetry.MetricsStart(ctx, command)
 
 	var commandOptions []telemetry.CommandOption
 	for _, fs := range []*flag.FlagSet{cmd.Flags(), cmd.PersistentFlags(), cmd.LocalFlags(), cmd.InheritedFlags()} {
--- a/cmd/werf/common/telemetry.go
+++ b/cmd/werf/common/telemetry.go
@@ -5,6 +5,7 @@ import (
 	"context"
 	"fmt"
 	"net"
+	"os"
 	"strings"
 
 	"github.com/go-git/go-git/v5/plumbing/transport"
@@ -21,6 +22,22 @@ var telemetryIgnoreCommands = []string{
 	"werf synchronization",
 	"werf completion",
 }
+
+// Flag and env var names for the OTLP metrics push exporter. Kept as
+// constants because they're referenced from both root.go (flag
+// registration) and here (flag lookup).
+const (
+	FlagOTLPMetricsEnabled  = "otlp-metrics-enabled"
+	FlagOTLPMetricsEndpoint = "otlp-metrics-endpoint"
+	EnvOTLPMetricsEnabled   = "WERF_OTLP_METRICS_ENABLED"
+	EnvOTLPMetricsEndpoint  = "WERF_OTLP_METRICS_ENDPOINT"
+)
+
+// metricsHandle bridges the PreRun (where the OTLP pipeline is initialized
+// from CLI flags) to ShutdownTelemetry (called once from main.go after the
+// command finishes). pkg/telemetry itself holds no package-level state; see
+// telemetry.MetricsHandle.
+var metricsHandle *telemetry.MetricsHandle
 
 func InitTelemetry(ctx context.Context) {
 	if err := telemetry.Init(ctx, telemetry.TelemetryOptions{
@@ -39,8 +56,8 @@ func InitTelemetry(ctx context.Context) {
 func ShutdownTelemetry(ctx context.Context, exitCode int) {
 	telemetry.GetTelemetryWerfIO().CommandExited(ctx, exitCode)
 
-	telemetry.MetricsEnd(ctx, exitCode)
-	if err := telemetry.MetricsShutdown(ctx); err != nil {
+	metricsHandle.MetricsEnd(ctx, exitCode)
+	if err := metricsHandle.MetricsShutdown(ctx); err != nil {
 		telemetry.LogF("unable to shutdown metrics: %s", err)
 	}
 
@@ -62,17 +79,32 @@ func TelemetryPreRun(cmd *cobra.Command, args []string) error {
 
 	InitTelemetry(ctx)
 
-	// Initialize OTLP metrics push exporter if enabled via flags
+	// Initialize OTLP metrics push exporter if enabled via flags/env.
 	if root := cmd.Root(); root != nil {
-		enabled, _ := root.PersistentFlags().GetBool("telemetry-enabled")
-		endpoint, _ := root.PersistentFlags().GetString("telemetry-otlp-endpoint")
-		if err := telemetry.InitMetrics(ctx, enabled, endpoint); err != nil {
+		enabled, err := root.PersistentFlags().GetBool(FlagOTLPMetricsEnabled)
+		if err != nil {
+			return fmt.Errorf("get --%s flag: %w", FlagOTLPMetricsEnabled, err)
+		}
+		endpoint, err := root.PersistentFlags().GetString(FlagOTLPMetricsEndpoint)
+		if err != nil {
+			return fmt.Errorf("get --%s flag: %w", FlagOTLPMetricsEndpoint, err)
+		}
+
+		handle, err := telemetry.InitMetrics(ctx, enabled, endpoint)
+		if err != nil {
+			if enabled {
+				// The user explicitly opted in: a misconfiguration must be
+				// visible instead of silently discarded (telemetry.LogF only
+				// writes when WERF_TELEMETRY_LOG_FILE is set).
+				fmt.Fprintf(os.Stderr, "WARNING: otlp metrics: %s\n", err)
+			}
 			telemetry.LogF("error: %s", err)
 		}
+		metricsHandle = handle
 	}
 
 	telemetry.GetTelemetryWerfIO().SetCommand(ctx, command)
-	telemetry.MetricsStart(ctx, command)
+	metricsHandle.MetricsStart(ctx, command)
 
 	var commandOptions []telemetry.CommandOption
 	for _, fs := range []*flag.FlagSet{cmd.Flags(), cmd.PersistentFlags(), cmd.LocalFlags(), cmd.InheritedFlags()} {
--- a/cmd/werf/root/root.go
+++ b/cmd/werf/root/root.go
@@ -140,6 +140,10 @@ func ConstructRootCmd(ctx context.Context) (*cobra.Command, error) {
 
 	templates.ActsAsRootCommand(rootCmd, *groups...)
 
+	// Global telemetry flags (OTLP metrics push)
+	rootCmd.PersistentFlags().Bool("telemetry-enabled", false, "Enable OTLP metrics exporter (push)")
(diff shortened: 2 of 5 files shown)


pkg/telemetry/otlp_metrics.go
```diff
@@ -0,0 +1,126 @@
...
+		otlpmetrichttp.WithURLPath(urlObj.Path),
+		otlpmetrichttp.WithTimeout(5*time.Second),
+		otlpmetrichttp.WithRetry(otlpmetrichttp.RetryConfig{Enabled: false}),
+	)
+
+	client := otlpmetrichttp.NewClient(opts...)
+	exporter, err := otlpmetric.New(ctx, client)
+	if err != nil {
+		return fmt.Errorf("create otlp metric exporter: %w", err)
+	}
```
Fral738: This doesn't compile: `otlpmetrichttp.NewClient` doesn't exist in `otlpmetrichttp v0.44.0` — this is the otel v0.20-era API (looks copied from the commented-out example in `pkg/telemetry/metric_exporter.go`). The current API returns the exporter directly:

```suggestion
    exporter, err := otlpmetrichttp.New(ctx, opts...)
    if err != nil {
        return fmt.Errorf("create otlp metric exporter: %w", err)
    }
```

The `go.opentelemetry.io/otel/exporters/otlp/otlpmetric` import (deprecated module, not in our dependency graph) should be removed as well.

pkg/telemetry/otlp_metrics.go
```diff
@@ -0,0 +1,126 @@
...
+	attrs := []attribute.KeyValue{
+		attribute.String("command", command),
+		attribute.String("os", runtime.GOOS),
+		attribute.String("arch", runtime.GOARCH),
+		attribute.Int("exit_code", exitCode),
+	}
+
+	runs.Add(ctx, 1, attrs...)
+	if !start.IsZero() {
+		dur.Record(ctx, time.Since(start).Seconds(), attrs...)
```
Fral738: Doesn't type-check against otel metric v1.24: `Add`/`Record` take `...metric.AddOption`/`...metric.RecordOption`, so attributes must be wrapped:

```suggestion
    runs.Add(ctx, 1, metric.WithAttributes(attrs...))
    if !start.IsZero() {
        dur.Record(ctx, time.Since(start).Seconds(), metric.WithAttributes(attrs...))
    }
```

pkg/telemetry/otlp_metrics.go
```diff
@@ -0,0 +1,126 @@
...
+	command := metricsState.command
+	start := metricsState.start
+	metricsState.mu.Unlock()
+
+	if !enabled {
+		return
+	}
+
+	runs, _ := meter.Int64Counter("werf_runs")
+	dur, _ := meter.Float64Histogram("werf_run_duration_seconds")
```
Fral738: Errors are discarded with `_` here — if instrument creation fails, `runs` is nil and `runs.Add` panics in the shutdown path of every werf command. Discarding errors with `_` is forbidden by our guidelines (AGENTS.md); please handle the errors and return early.

pkg/telemetry/otlp_metrics.go
```diff
@@ -0,0 +1,126 @@
...
+	metricsState.mu.Lock()
+	defer metricsState.mu.Unlock()
+
+	if !enabled {
+		metricsState.enabled = false
+		return nil
+	}
+	if endpoint == "" {
+		return fmt.Errorf("telemetry enabled but --telemetry-otlp-endpoint is empty")
+	}
```
Fral738, line 44: Two issues: (1) the error message references the CLI flag name `--telemetry-otlp-endpoint`, which leaks the CLI layer into `pkg/telemetry`; (2) per CODESTYLE, error messages should describe the action being performed. Consider something like `"otlp metrics enabled but endpoint is not set"`.

pkg/telemetry/otlp_metrics.go
```diff
@@ -0,0 +1,126 @@
...
+
+	urlObj, err := neturl.Parse(endpoint)
+	if err != nil {
+		return fmt.Errorf("bad otlp endpoint: %w", err)
+	}
+
+	var opts []otlpmetrichttp.Option
+	if urlObj.Scheme == "http" {
+		opts = append(opts, otlpmetrichttp.WithInsecure())
+	}
```
Fral738, line 62: `neturl.Parse` accepts almost any string, so a typo'd endpoint silently produces a broken exporter. Please validate the scheme (`http`/`https` whitelist) and a non-empty `Host`, returning an error otherwise ("validate early, validate a lot").

pkg/telemetry/otlp_metrics.go
```diff
@@ -0,0 +1,126 @@
...
+		metricsState.enabled = false
+		return nil
+	}
+	if endpoint == "" {
+		return fmt.Errorf("telemetry enabled but --telemetry-otlp-endpoint is empty")
+	}
+
+	urlObj, err := neturl.Parse(endpoint)
+	if err != nil {
+		return fmt.Errorf("bad otlp endpoint: %w", err)
```
Fral738: Also per CODESTYLE:
```suggestion
        return fmt.Errorf("parse otlp endpoint: %w", err)
```

pkg/telemetry/otlp_metrics.go
```diff
@@ -0,0 +1,126 @@
...
+		otlpmetrichttp.WithRetry(otlpmetrichttp.RetryConfig{Enabled: false}),
+	)
+
+	client := otlpmetrichttp.NewClient(opts...)
+	exporter, err := otlpmetric.New(ctx, client)
+	if err != nil {
+		return fmt.Errorf("create otlp metric exporter: %w", err)
+	}
+
+	reader := sdkmetric.NewPeriodicReader(exporter, sdkmetric.WithInterval(2*time.Second))
```
Fral738: A 2s periodic interval is pointless for a short-lived CLI process — the only guaranteed delivery point is the forced flush inside `provider.Shutdown`. Suggest dropping `WithInterval` and relying on shutdown flush; also consider enabling retry, otherwise a transient network hiccup in the 5s flush window silently drops the whole batch.

pkg/telemetry/otlp_metrics.go
```diff
@@ -0,0 +1,126 @@
...
+// Metrics push exporter (OTLP). NOP-safe when disabled.
+
+var metricsState struct {
+	mu       sync.Mutex
+	enabled  bool
+	provider *sdkmetric.MeterProvider
+	meter    metric.Meter
+	start    time.Time
+	command  string
+}
```
Fral738: Global mutable singleton makes this untestable (and CODESTYLE says to avoid global state). Consider having `InitMetrics` return a handle that's passed to the End/Shutdown calls — that would also let us add tests for the No-Op path and the lifecycle, which are currently missing.

pkg/telemetry/otlp_metrics.go
```diff
@@ -0,0 +1,126 @@
...
+	var opts []otlpmetrichttp.Option
+	if urlObj.Scheme == "http" {
+		opts = append(opts, otlpmetrichttp.WithInsecure())
+	}
+	opts = append(opts,
+		otlpmetrichttp.WithEndpoint(urlObj.Host),
+		otlpmetrichttp.WithURLPath(urlObj.Path),
+		otlpmetrichttp.WithTimeout(5*time.Second),
+		otlpmetrichttp.WithRetry(otlpmetrichttp.RetryConfig{Enabled: false}),
+	)
```
Fral738, line 67: The PR description mentions Datadog/Honeycomb, but those require auth headers (`DD-API-KEY`, `x-honeycomb-team`) and there's no `otlpmetrichttp.WithHeaders` support here. Either add a headers option or state explicitly in the docs that only unauthenticated collectors are supported.

cmd/werf/common/telemetry.go
```diff
@@ -56,7 +61,18 @@ func TelemetryPreRun(cmd *cobra.Command, args []string) error {
...
 	InitTelemetry(ctx)
+
+	// Initialize OTLP metrics push exporter if enabled via flags
+	if root := cmd.Root(); root != nil {
+		enabled, _ := root.PersistentFlags().GetBool("telemetry-enabled")
+		endpoint, _ := root.PersistentFlags().GetString("telemetry-otlp-endpoint")
+		if err := telemetry.InitMetrics(ctx, enabled, endpoint); err != nil {
+			telemetry.LogF("error: %s", err)
+		}
+	}
```
Fral738: `telemetry.LogF` writes only when `WERF_TELEMETRY_LOG_FILE` is set — otherwise this error (including "enabled but endpoint is empty") is silently discarded. The user explicitly opted in with `--telemetry-enabled`, so misconfiguration must be visible: either fail fast by returning the error, or print a warning to stderr.

Also: flag names are duplicated as magic strings between here and `root.go`, and the `GetBool`/`GetString` errors are discarded with `_` — please extract constants and handle the errors.

cmd/werf/root/root.go
```diff
@@ -140,6 +140,10 @@ func ConstructRootCmd(ctx context.Context) (*cobra.Command, error) {
 
 	templates.ActsAsRootCommand(rootCmd, *groups...)
 
+	// Global telemetry flags (OTLP metrics push)
+	rootCmd.PersistentFlags().Bool("telemetry-enabled", false, "Enable OTLP metrics exporter (push)")
+	rootCmd.PersistentFlags().String("telemetry-otlp-endpoint", "", "OTLP collector endpoint, e.g., http://collector:4318/v1/metrics")
+
```
Fral738: Two issues:

1. Every werf flag has a `WERF_*` env fallback (see the `util.GetBoolEnvironmentDefaultFalse` pattern in `cmd/werf/common/common.go`) and mentions it in the help text. These flags don't — and CI (the primary audience of this feature) would have to pass flags to every werf invocation instead of setting one env var.
2. `--telemetry-enabled` reads as if it controls werf's existing documented telemetry (`WERF_TELEMETRY`), but it doesn't. Please rename (e.g. `--otlp-metrics-enabled` / `--otlp-metrics-endpoint`) or integrate with the existing mechanism.

The `// Global telemetry flags (OTLP metrics push)` comment just restates the code — per our guidelines it should be dropped.

Fral738 (changes requested):
Thanks for the contribution — push-based OTLP metrics for ephemeral CI processes is a valid and welcome idea. However, I have to request changes: the branch currently does not compile (3 typecheck errors, see inline comments), and there are several design/product concerns that need to be resolved before merge:

1. **Dependencies**: the new code imports the deprecated `go.opentelemetry.io/otel/exporters/otlp/otlpmetric` module which is not in `go.mod`/`go.sum`, and the diff doesn't touch `go.mod` at all. Please promote `otlpmetrichttp` and `sdk/metric` to direct dependencies at versions consistent with `otel v1.24.0`, and verify with `task build` / `task lint:golangci-lint`. Note: new external dependencies must be explicitly called out in the PR description per our contribution rules.
2. **Naming collision**: werf already has documented telemetry controlled by `WERF_TELEMETRY` (traces to telemetry.werf.io). A root flag named `--telemetry-enabled` that does *not* control it will mislead users in the most trust-sensitive area (data collection). Please either rename the flags (e.g. `--otlp-metrics-enabled` / `--otlp-metrics-endpoint`) or integrate with the existing mechanism.
3. **PR description vs implementation**: there are no "build durations, deployment status, pipeline execution metrics" — only a `werf_runs` counter and a process-duration histogram recorded once at exit. Direct export to Datadog/Honeycomb also won't work without auth-header support. Please align the description with what's actually shipped.
4. **Missing before merge**: `WERF_*` env-var fallbacks for the new flags (project convention, and CI is the primary audience here); docs update (`docs/pages_en` **and** `docs/pages_ru` telemetry pages); CLI reference regeneration (`task doc:gen` — two new persistent root flags change the help of every command); tests (No-Op path when disabled, endpoint validation, start→end→shutdown lifecycle).

Happy to re-review once the branch compiles and the flag naming question is resolved.

glatinone:
Closing this PR to avoid adding unnecessary review overhead for maintainers. Thank you for your time!

PR #7645, opened 2026-07-09
refactor(config): drop deprecated fromImage / dependencies.image / import.image / clearWerfLabels

alexey-igrychev:
Part of v3 breaking-change cleanup.

## What

Removes the following long-deprecated werf.yaml directives (each already emitted a \`GlobalDeprecationWarningLn\` promising removal in a future version):

- \`fromImage:\` (stapel image) → use \`from:\`
- \`dependencies[].image:\` → use \`dependencies[].from:\`
- \`import[].image:\` → use \`import[].from:\`
- \`imageSpec.config.clearWerfLabels\` / \`meta.build.imageSpec.config.clearWerfLabels\` (already no-op)

Also removes the matching Go fields, the \`ImageSpec.ClearWerfLabels\` output field, and updates the affected tests / fixtures.

## Breaking changes

Users still relying on the deprecated names will get \`unknown attribute\` config errors. Rename to \`from:\` / drop \`clearWerfLabels\`.


alexey-igrychev:
rejected

PR #7648, opened 2026-07-09
fix(slug): strip trailing hyphen instead of leading

alexey-igrychev:
Part of v3 breaking-change cleanup.

## What

Replaces \`strings.HasPrefix\` with \`strings.HasSuffix\` in \`pkg/slug/slug.go\`. The intent was to strip a trailing \`-\` from the cropped slug data so the join with the separator does not produce \`something---hash\`; the code accidentally checked the prefix instead, which never matched (slugified data never starts with \`-\`) and left the double-hyphen bug in place.

The block was flagged in-code:

> \`\`\`\n> // legacy: this check cannot be fixed without breaking reproducibility.\n> // Fix by replacing HasPrefix on HasSuffix in the following major releases.\n> \`\`\`

## Breaking changes

For inputs whose cropped slugged form would end with \`-\`, the resulting slug now has one hyphen instead of two before the hash suffix. Slugs feed into k8s resource names, image tags, etc., so any resource named with the buggy \`--\` slug gets a new name after upgrade.

## Tests

Existing test cases \`legacyCaseWithTwoHyphen\` / \`legacyCaseWithTwoHyphen_2\` renamed and updated to reflect the fixed output.

### DIFF
--- a/pkg/slug/slug.go
+++ b/pkg/slug/slug.go
@@ -179,9 +179,7 @@ func slug(data string, maxSize int) string {
 	if sluggedData != "" {
 		croppedSluggedData := cropSluggedData(sluggedData, murmurHash, maxSize)
 
-		// legacy: this check cannot be fixed without breaking reproducibility.
-		// Fix by replacing HasPrefix on HasSuffix in the following major releases.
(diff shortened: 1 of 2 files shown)


alexey-igrychev:
rejected

PR #7659, opened 2026-07-14
fix(true_git): forward include.path to submodule sync/update commands

alexey-igrychev:
## Problem

On GitLab Runner 18.6+ git auth is provided via an external git config referenced through `include.path` in the repo's `.git/config` instead of tokenized submodule URLs. werf's cached-worktree submodule preparation (`git submodule sync --recursive` / `git submodule update --checkout --force --init --recursive`) failed with:

```
fatal: could not read Username for 'https://...': No such device or address
```

The child `git fetch`/`clone` processes spawned by those commands run against each submodule's own git dir and do not inherit the parent repo's file-level `include.path`, so the auth chain (credential config, `url.*.insteadOf`) never reaches them.

## Solution

Discover `include.path` values from the source repo (`git config --show-origin --get-all include.path`) and pass them as `-c include.path=<abs>` to the internal `submodule sync`/`submodule update` commands. `-c` is exported to git child processes via `GIT_CONFIG_PARAMETERS`, restoring the auth chain for submodule fetches. Mirrors GitLab Runner MR gitlab-org/gitlab-runner!5962.

- Relative values are resolved against the declaring config file's directory (matching git's own include resolution); `~`-prefixed values pass through (git expands them).
- Only file paths are forwarded — never credential values — so nothing secret lands on argv or in logs.
- Non-submodule cached-worktree flow, common git options, and locking are unchanged.

## Testing

- `TestAI_UpdateSubmodulesForwardsIncludePath`: behavioural regression test — an include stub rewriting `https://` to a bogus transport surfaces in the child clone error only when forwarding works; verified it fails with the fix reverted.
- `TestAI_GetIncludePathOptions`: helper contract (absent key, absolute/relative/tilde values).
- `TestAI_SwitchWorkTreeNonSubmodulesUnaffectedByIncludePath`: non-submodule path unaffected.
- `task build`, `task lint:golangci-lint`, ginkgo suite 32/32 green.

### DIFF
--- a/pkg/true_git/common.go
+++ b/pkg/true_git/common.go
@@ -1,11 +1,49 @@
 package true_git
 
 import (
+	"context"
+	"errors"
+	"fmt"
 	"os"
+	"os/exec"
+	"path/filepath"
+	"strings"
 )
 
 func getCommonGitOptions() []string {
 	return []string{"-c", "core.autocrlf=false", "-c", "gc.auto=0", "-c", "commit.gpgsign=false", "-c", "core.untrackedCache=false", "-c", "core.splitIndex=false"}
+}
+
+func getIncludePathOptions(ctx context.Context, repoDir string) ([]string, error) {
+	configCmd := NewGitCmd(ctx, &GitCmdOptions{RepoDir: repoDir}, "config", "--get-all", "include.path")
+	if err := configCmd.Run(ctx); err != nil {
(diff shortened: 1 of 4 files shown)


alexey-igrychev:
Superseded by a properly named branch/PR per CONTRIBUTING.md conventions.

PR #7669, opened 2026-07-20
fix(build, buildah): retry pull when cached image id is missing

Fral738:
## Summary

With the Buildah backend a build fails with `unable to create container using base image ...: image not known` even though the stage image exists in the registry. It needs a pre-existing host state: werf pulled the image earlier in the same run and cached its local image ID in memory, and the local image behind that ID was then removed (host cleanup, `buildah rmi`, parallel prune) before the stage was used as a base.

## What

- When creating a build container from a cached platform-specific image ID and buildah reports `image not known`, werf re-pulls the image by ref, refreshes the cached ID, and retries container creation once — the build proceeds instead of failing.
- The fallback triggers only when a target platform is set and the image was resolved through the in-memory pulled-image cache; any other container creation error fails as before.
- If the re-pull fails, the build fails with `unable to pull image ... after cached imageID miss`.
- `image not known` is detected by suffix match against the canonical `containers/storage` error, so wrapped errors match too.
- UNVERIFIED: not reproduced end-to-end on a live buildah host — covered by a unit regression test with a stubbed buildah; an e2e run on Linux deleting the local image between stages would settle it.
- Deliberately unchanged: a failed re-pull keeps the stale cache entry, so the next attempt goes through the same fallback instead of starting from an empty cache.

## Why

`BuildahBackend` keeps an in-memory ref+platform → imageID map populated on pull, and `createContainers` resolves base images through it. The map is never invalidated when the local image behind an ID disappears, so `buildah from` receives a dangling ID while the tag is still perfectly pullable from the registry. Left alone, this intermittently fails multi-stage staged-Dockerfile builds on shared runners where host cleanup runs between stages.

### DIFF
--- a/pkg/container_backend/buildah_backend.go
+++ b/pkg/container_backend/buildah_backend.go
@@ -86,6 +86,14 @@ func platformMatches(inspect *thirdparty.BuilderInfo, targetPlatform string) boo
 	return true
 }
 
+func isImageNotKnownError(err error) bool {
+	if err == nil {
+		return false
+	}
+
+	return strings.Contains(strings.ToLower(err.Error()), "image not known")
+}
+
 func (backend *BuildahBackend) Info(ctx context.Context) (info.Info, error) {
 	return backend.buildah.Info(ctx)
 }
@@ -132,9 +140,11 @@ func (backend *BuildahBackend) createContainers(ctx context.Context, images []st
 		}
 
 		resolvedImg := img
+		usedCachedImageID := false
 		if opts.TargetPlatform != "" {
 			if id, ok := backend.getPulledImageID(img, opts.TargetPlatform); ok {
 				resolvedImg = id
+				usedCachedImageID = true
 			} else if inspect, err := backend.buildah.Inspect(ctx, img); err == nil && inspect != nil {
 				if !platformMatches(inspect, opts.TargetPlatform) {
 					return nil, fmt.Errorf("local image %q has platform %s/%s, but target platform is %q; pull the correct image first", img, inspect.OCIv1.OS, inspect.OCIv1.Architecture, opts.TargetPlatform)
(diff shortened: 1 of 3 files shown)


pkg/container_backend/buildah_backend.go
```diff
@@ -86,6 +86,14 @@ func platformMatches(inspect *thirdparty.BuilderInfo, targetPlatform string) boo
 	return true
 }
 
+func isImageNotKnownError(err error) bool {
+	if err == nil {
+		return false
+	}
+
+	return strings.Contains(strings.ToLower(err.Error()), "image not known")
```
nervgh: Let's replace with `strings.HasSuffix(err.Error(), storage.ErrImageUnknown.Error())`

https://github.com/werf/werf/blob/efd1e46bef1c625ebcf28305cecdb57ce83bf250/pkg/buildah/native_linux.go#L922

pkg/container_backend/buildah_backend.go
```diff
@@ -132,17 +140,38 @@ func (backend *BuildahBackend) createContainers(ctx context.Context, images []st
...
 			}
 		}
 
-		_, err := backend.buildah.FromCommand(ctx, containerID, resolvedImg, buildah.FromCommandOpts(backend.getBuildahCommonOpts(ctx, true, nil, opts.TargetPlatform)))
+		fromCommandOpts := buildah.FromCommandOpts(backend.getBuildahCommonOpts(ctx, true, nil, opts.TargetPlatform))
+		_, err := backend.buildah.FromCommand(ctx, containerID, resolvedImg, fromCommandOpts)
+		if err != nil && opts.TargetPlatform != "" && usedCachedImageID && isImageNotKnownError(err) {
+			logboek.Context(ctx).Debug().LogF("Cached imageID %q for %q not found locally, pulling by ref and retrying\n", resolvedImg, img)
+
+			backend.pulledImageIDs.Delete(pulledImageKey{img, opts.TargetPlatform})
```
nervgh: Redundant `Delete` creates unnecessary race window

```go
backend.pulledImageIDs.Delete(pulledImageKey{img, opts.TargetPlatform})  // ← remove this
pulledImageID, pullErr := backend.buildah.Pull(ctx, img, ...)
if pulledImageID != "" {
    backend.storePulledImageID(img, opts.TargetPlatform, pulledImageID)
}
```

`sync.Map.Store` overwrites the existing value for a key — this is explicitly tested in `buildah_backend_test.go:39-46` ("overwrites imageID on repeated pull") and relied upon by the existing `Pull` method (line 841). The preceding `Delete` is therefore unnecessary and introduces a TOCTOU window where concurrent goroutines (building the same base image for the same platform in parallel) get a cache miss despite a valid cache entry being about to be written.

**Proposal / Fix:** Remove the `Delete` call. `Store` will overwrite the stale value anyway. If `Pull` fails, the stale cache entry remains, which is strictly better than the current behavior of leaving the cache empty after a failed retry.

nervgh (changes requested):

Fral738:
## Verification

- Scoped local run: `task test:unit paths="./pkg/container_backend/..."` — the regression spec passes (macOS; full suite fails on unrelated `pkg/true_git` GPG env issues).
- Mutation: retry `FromCommand` removed from the fallback → spec failed with the target `image not known` error.
- Mutation: `storePulledImageID` dropped from the fallback → spec failed on the cache assertion (`sha256:stale` ≠ `sha256:fresh`).
- Mutation: `HasSuffix` → `HasPrefix` in `isImageNotKnownError` → survived: the stub returned a bare `image not known`, indistinguishable by prefix/suffix/substring. The spec was reshaped to return the error wrapped as buildah does (`locating image with name ...: image not known`); the mutation is now caught.

## Review focus

- Fallback conditions in `createContainers`: confirm it cannot fire for a plain ref (no cached ID) and cannot loop.
- Confirm suffix-matching `types.ErrImageUnknown` is the right contract for errors coming out of `buildah.FromCommand` in native mode.

## Follow-up

- [ ] Reproduce the original `image not known` scenario on a Linux host with the Buildah backend (delete local stage images between runs) to confirm the fix end-to-end.

nervgh (approved):

PR #7675, opened 2026-07-22
feat(build): show low-level operations time summary in debug mode

Fral738:
## Summary

`werf build` with `--log-debug` now ends with an `Operations summary` block — per-operation count, total, wall, avg and max durations for the low-level operations of the build, including docker daemon request latency and per-method registry client timings — and a `Stage cache summary` block counting how each stage was satisfied. The same aggregates are saved into the JSON build report, so a degradation between pipelines can be compared mechanically. Together they answer "where did the build time go" — slow registry, slow daemon, missing cache, lock contention — without reading the full debug log.

### Operations summary

- With debug logging enabled, the end of the build log contains an `Operations summary` block: one line per observed operation with `count`, `total`, `wall`, `avg`, `max` in seconds, sorted by total descending.
- Tracked operations: `image pull`, `image push`, `image build`, `local image inspect`, `image save/load`, `import checksum`, `git clone`/`fetch`/`patch`/`archive`/`checksum`, `registry: <Method>` (one line per registry client method, e.g. `registry: TryGetRepoImage`, `registry: Tags`), `docker daemon API`, `stage lock wait`, `context add files`.
- `wall` is the union of possibly overlapping intervals across parallel workers and never exceeds the build time; `total` sums all intervals and may exceed it, and `×N` on the line shows the parallelism ratio when they differ.
- The closing `build time:` line gives the elapsed build wall-clock for comparison.
- A pull on the buildah backend is recorded exactly once even when reached through the fetch-base-image path, and a lost clone race records `git clone` and the fallback `git fetch` as disjoint intervals.

### Stage cache summary

- A `Stage cache summary` block follows: per-source stage counts — `found in local stages storage`, `found in repo stages storage`, `copied from secondary storage`, `built` — plus a `total:` line; a source with zero stages is omitted.

### Docker daemon latency

- `docker daemon API` counts every HTTP request werf makes to the docker daemon, measured at the transport, so a slow-responding daemon shows up as a high `avg` on this line.
- Only `unix://` and `npipe://` daemon hosts are measured. For `tcp://` and `ssh://` hosts the stock client is kept and no line appears: the timing wrapper would silently disable the transport-derived dialer and TLS config that hijacked connections (attach/exec) rely on.
- The wrapper cannot break werf: when the measured client cannot be built, the stock unmeasured client is used, and custom `httpHeaders` from the docker config file are preserved on the measured client.

### Build report

- With debug logging enabled, the JSON build report gains `Operations` (per-operation `Count`, `TotalTimeSeconds`, `WallTimeSeconds`, `AvgTimeSeconds`, `MaxTimeSeconds`) and `StageCache` (per-source counts) fields, documented in the build process page (en and ru).
- Without debug logging both fields are absent and the report is byte-identical to before; the envfile report format never changes.
- The report is written before the build command fully ends, so operations that happen after it (host cleanup) are not included in the report aggregates while they can appear in the log block.

### Scope of measurement

- Without debug logging nothing is printed and nothing is measured: the transport wrapper is a no-op without a collector in the request context, and no collector is created, so regular builds get no overhead.
- Operation categories do not overlap: backend operations talk to the daemon/buildah directly, `registry: <Method>` lines cover only werf's own registry client. UNVERIFIED for the buildah backend at runtime (does not compile on the development host); settled by any Linux e2e run with `--log-debug`.
- `docker daemon API` overlaps by design with the backend operations that are implemented as daemon calls (`image build`, `local image inspect`, …): it is the transport-level view of the same work, not an additional cost.
- Bytes pushed/pulled are deliberately NOT measured: for the docker backend the data plane lives in the daemon and for buildah in its library, so a transport-level byte count at werf's level would only see manifests and mislead.
- `werf run`, attach and exec behavior is unchanged for all daemon host types. VERIFIED by code reading of the docker client dialer fallback (unix/npipe dial by proto without the transport); tcp/ssh hosts never get the wrapper.

## Why

The debug log of a large build is hundreds of thousands of lines, and tracking a performance degradation down means manually correlating timestamps across them; in practice support has to request and read whole log archives. An aggregate printed once at the end is enough to point at the guilty class of operations, the cache-source counters separate "slow because everything rebuilds" from "slow because the network is slow", and the daemon latency line answers a recurring support case — a daemon that responds slowly — that no other line isolates. Registry timings are split per client method because a single line hid the difference between a thousand cheap cache lookups and a handful of slow tag listings. The same aggregates in the build report make regressions comparable across pipelines without log archaeology. Full per-operation tracing was rejected: it needs an export pipeline and answers questions nobody asked yet, while the aggregate is cheap and fits the log.


Fral738:
## Verification

- Smoke run on macOS with the docker backend (unix socket) against a local `registry:2`: `werf build --log-debug --repo localhost:5000/...` prints per-method registry lines (`registry: PushImage 2 op avg 184ms`, `registry: Tags 2 op`, `registry: GetRepoImage 1 op`), a `docker daemon API` line (4 op, avg 5ms) and a `context add files` line; column alignment holds with the longer labels. The same build without the flag prints nothing. A rebuild shows `found in repo stages storage: 1` and a single `registry: GetRepoImage` lookup.
- Report smoke: `werf build --log-debug --save-build-report` produces JSON with `Operations` and `StageCache`; the same command without `--log-debug` produces a report without either key.
- Mutation log (each broke the implementation, confirmed the named test failed, then reverted):
  - `unionDuration` replaced with a plain sum → `fully overlapping intervals` table entry failed.
  - max computation inverted → `computes total, avg and max over intervals and sorts by total time` failed on `MaxTime`; sort order inverted → same test failed on ordering.
  - counter increment replaced with `= 1` → `counts events and sorts by count` failed.
  - `opstats.Observe` removed from `timingRoundTripper.RoundTrip` → `records request duration into the collector from the request context` failed.
  - `StageCache` population dropped from `SetOperationsSummary` → `serializes operations and stage cache aggregates to json` failed.
  - `sync.Once` removed from `Observe`'s done func → `records at most once when the done func is called twice` failed.
  - `Count` zeroed in the report record → report test failed (previously survived — found by independent review, test strengthened to assert the full decoded structure).
  - `Observe` removed from the registry decorator's `Tags` → decorator test failed (previously uncovered — found by independent review).
  - per-method registry label replaced with a static one → decorator test failed on the expected `registry: Tags` operation.
- An adversarial self-review pass (agent-code-review) found and fixed: buildah pull double-count, fatal error path in the docker client wrapper, dropped `httpHeaders`, `git fetch` nested inside `git clone` on the lost-race path, and the misleading `build context` label.
- An independent agent review confirmed the implementation and found two test-evidence gaps, both closed with mutation proof.
- Not run: buildah backend hooks at runtime — the macOS dev binary does not compile the buildah backend; covered by unit tests and review only. Secondary-storage counter at runtime — needs a secondary storage setup. Hijack (attach/exec) over a wrapped unix-socket client — needs an interactive container session; the unix/npipe-only guard rests on code reading of the docker client dialer fallback.

## Review focus

- Interval-union semantics in `pkg/opstats` (`unionDuration`): wall time is the load-bearing number of the block.
- `wrapAPIClientTransport` in `pkg/docker/timing.go`: the rebuilt client must behave identically for unix/npipe hosts; on any construction error the stock client is used silently — check that losing the metric is acceptable as the failure mode.
- Counter placement in `pkg/build/build_phase.go`: confirm the local/repo/secondary/built points cover every way a stage gets satisfied, and none double-fires.
- `createBuildReport` runs per build phase; confirm a multi-phase or multi-platform build cannot write the report before the collector has the bulk of its data.

## Follow-up

- [ ] Measure tcp/ssh daemon hosts once the docker client exposes a safe transport hook.
- [ ] Consider per-stage byte sizes as a proxy for transfer volume (transport-level byte counts were rejected as misleading).
- [ ] HTTP-level registry breakdown (auth/manifest/blob) via the go-containerregistry transport, if per-method timings prove insufficient on real pipelines.

Fral738:
```
┌ Operations summary
│ - image build                      2 op   total  4.42s   wall  2.79s   avg  2.208s   max  2.786s   ×1.6
│ - image push                       2 op   total  1.48s   wall  1.48s   avg  0.738s   max  0.752s
│ - registry: PushImage              4 op   total  1.08s   wall  0.59s   avg  0.269s   max  0.371s   ×1.8
│ - registry: Tags                   3 op   total  0.44s   wall  0.44s   avg  0.147s   max  0.180s
│ - registry: GetRepoImage           2 op   total  0.34s   wall  0.34s   avg  0.171s   max  0.205s
│ - stage lock wait                  2 op   total  0.29s   wall  0.29s   avg  0.144s   max  0.166s
│ - git archive                      1 op   total  0.07s   wall  0.07s   avg  0.069s   max  0.069s
│ - docker daemon API                8 op   total  0.04s   wall  0.03s   avg  0.004s   max  0.013s   ×1.2
│ - context add files                2 op   total  0.00s   wall  0.00s   avg  0.000s   max  0.001s   ×1.4
│ build time: 4.96s (wall must not exceed it; total may)
└ Operations summary

┌ Stage cache summary
│ - built                              2 stage(s)
│ total: 2 stage(s)
└ Stage cache summary
```

alexey-igrychev (changes requested):
7 blocking findings: `--require-built-images` bypasses stats, Docker Stapel and generic/shallow-Git paths are uninstrumented, image-save duration ends before stream consumption, cache source is wrong on publish race, and local stage-digest lock wait is omitted. Details inline.

pkg/git_repo/remote.go
```diff
@@ -468,6 +475,8 @@ func (repo *Remote) fetchOriginFullCore(ctx context.Context, kind mirrorKind) er
 		}
 	}
```
alexey-igrychev, line 477: This observation covers only full-mirror fetches. Remote mappings pinned to a tag or SHA use `shallowFetch` in `remote_shallow.go`, so real network fetches are absent from `git fetch`. Add timing around shallow fetch, and decide separately whether `ls-remote` should be included.

pkg/docker_registry/docker_registry.go
```diff
@@ -102,6 +102,9 @@ func NewDockerRegistry(ctx context.Context, repositoryAddress, implementation st
 	if err != nil {
 		return nil, err
 	}
+
```
alexey-igrychev, line 105: The decorator covers only `NewDockerRegistry`, but the core build uses `docker_registry.API()` for base-image lookup and `FetchDependencies`. `API()` returns a separate global `generic` without the timing wrapper, so these registry requests are absent from the summary. Wrap `GenericApiInterface` at the shared API path.

pkg/build/build_phase.go
```diff
@@ -725,6 +731,8 @@ func (phase *BuildPhase) onImageStage(ctx context.Context, img *image.Image, stg
 			return err
 		}
 
+		opstats.CountEvent(ctx, opstats.EventStageBuilt)
```
alexey-igrychev: `EventStageBuilt` is incremented immediately after `buildStage`, but `atomicBuildStageImage` can then discard the locally built image and select an already-published repo stage. In a race, `StageCache` reports `built` although the selected stage was satisfied from repo storage. Have the atomic path return the winning source and count exactly that source.

pkg/build/conveyor.go
```diff
@@ -703,6 +705,14 @@ func (c *Conveyor) Build(ctx context.Context, opts BuildOptions) ([]*ImagesRepor
 		return nil, err
 	}
 
+	var opsCollector *opstats.Collector
```
alexey-igrychev: `ShouldBeBuilt` does not create or bind an `opstats.Collector`, unlike `Build`. Therefore `werf build --require-built-images --log-debug` prints no summary and omits `Operations` / `StageCache` from the JSON report. Share the collector lifecycle between `Build` and `ShouldBeBuilt`.

pkg/build/conveyor.go
```diff
@@ -773,6 +785,51 @@ func disableUnlessDebugConveyorPhases(logProcess types.LogProcessInterface) type
 	return logProcess
 }
 
+func (c *Conveyor) logOperationsSummary(ctx context.Context, collector *opstats.Collector, buildTime time.Duration) {
```
alexey-igrychev, line 795: `stage lock wait` measures only storage-lock acquisition. The parallel conveyor first waits on local `GetStageDigestMutex(...).Lock`; with the same digest this can be the primary wait, yet this summary omits it. Measure this mutex wait or explicitly rename the metric to storage-lock-only.

pkg/container_backend/docker_server_backend.go
```diff
@@ -472,9 +481,11 @@ func (backend *DockerServerBackend) PruneVolumes(ctx context.Context, options pr
 }
 
 func (backend *DockerServerBackend) SaveImageToStream(ctx context.Context, imageName string) (io.ReadCloser, error) {
+	defer opstats.Observe(ctx, opstats.OperationImageSaveLoad)()
```
alexey-igrychev: `SaveImageToStream` completes its observation before consuming the returned stream. The actual tar is read later via `io.Copy`, so saving a large image gets nearly zero `image save/load` time while the transfer takes time. Return a reader that completes the observation on EOF/Close, and complete it immediately on the initial API error.

pkg/container_backend/docker_server_backend.go
```diff
@@ -77,14 +78,17 @@ func (backend *DockerServerBackend) HasStapelBuildSupport() bool {
 }
 
 func (backend *DockerServerBackend) BuildStapelStage(ctx context.Context, baseImage string, opts BuildStapelStageOptions) (string, error) {
+	defer opstats.Observe(ctx, opstats.OperationImageBuild)()
```
alexey-igrychev: This hook does not cover Docker Stapel builds: `BuildStapelStage` always panics because `HasStapelBuildSupport()` returns false. The actual path is `LegacyStapelStageBuilder.Build` → `LegacyStageImage.Build`, where no observation exists. The debug summary therefore omits `image build` for Docker Stapel; move the hook to the real legacy path.

alexey-igrychev (commented):
Recreated the findings as inline comments.

pkg/git_repo/remote.go
Fral738: Fixed in ac7bceaae: shallowFetch is now observed as git fetch. The ls-remote network call is reported as its own git ls-remote line rather than folded into fetch — it answers a different question (ref resolution vs data transfer) and is cached per process, so merging them would skew fetch avg.

pkg/docker_registry/docker_registry.go
Fral738: Fixed in ac7bceaae: API() now wraps the shared generic client with the same per-method timing decorator (registry: GetRepoImage / MutateAndPushImage / GetRepoImageConfigFile), covered by a unit test.

pkg/build/build_phase.go
Fral738: Fixed in ac7bceaae: both counters moved into atomicBuildStageImage terminal points — the discard path counts the winning primary-storage source (local/repo split preserved), only the newly-built path counts built.

pkg/build/conveyor.go
Fral738: Fixed in ac7bceaae: collector setup extracted into newOperationsCollector, used by both Build and ShouldBeBuilt. Verified: werf build --require-built-images --log-debug now prints both blocks and the report carries Operations/StageCache.

pkg/build/conveyor.go
Fral738: Fixed in ac7bceaae: the digest mutex wait is now reported as stage lock wait (parallel tasks); the storage metric is renamed to stage lock wait (storage) to keep the two distinguishable.

pkg/container_backend/docker_server_backend.go
Fral738: Fixed in ac7bceaae: SaveImageToStream (both backends) returns an observed reader — the measurement completes on first read error (incl. EOF) or Close, and immediately on the initial API error. Covered by a unit test with mutation proof.

pkg/container_backend/docker_server_backend.go
Fral738: Fixed in ac7bceaae: the hook moved to LegacyStageImage.Build (the real Docker Stapel path); the dead hooks on the panicking BuildStapelStage/CalculateDependencyImportChecksum are removed.

PR #7676, opened 2026-07-24
feat(build): add shallow mirrors for immutable remote refs

alexey-igrychev:
## Summary

Reduce network traffic and local cache size for immutable external Git mappings. `git.tag` and `git.commit` use depth-1 shell-git mirrors and skip redundant fetches, while branch mappings and repositories requiring full history retain full go-git mirrors.

## Key changes

- Split each URL cache into `full` and `shallow` mirrors and update Git repo GC for the nested layout.
- Fetch immutable commits and tags at depth 1, cache `ls-remote` tag results per authenticated URL, and recursively peel annotated tags.
- Persist a full-mirror fallback when fetch-by-SHA is unsupported or the target commit contains submodules.
- Preserve shallow mirrors used by sibling mappings during fallback; normal LRU GC removes them later.
- Serialize mirror writes by URL and mirror kind, and document shallow-ref behavior and retag limitations.

## Why

Immutable refs need one commit and its tree rather than the complete remote history. Avoiding full clone/fetch operations reduces build latency, network traffic, and local cache size, especially for large external repositories.

## Review focus / risks

- Fallback marker semantics and lock ordering in `pkg/git_repo/remote_shallow.go`.
- Mixed `git.branch`, `git.tag`, and `git.commit` mappings sharing one URL.
- Cache layout version `5` → `6` causes a one-time remote re-fetch; old cache data is removed by normal GC.
- Retagged tags can no longer reuse ancestry-based cached stages when the old commit is absent from the shallow mirror.


alexey-igrychev:
Superseded by #7677 after renaming the head branch to match the repository branch-name convention.

PR #7682, opened 2026-07-28
refactor(build): curate --log-debug output and add debug env vars reference

alexey-igrychev:
## What

Curates the `werf build --log-debug` output: removes noise, hides high-volume channels behind topic env vars, eliminates unconditional debug work, and documents every debug environment variable in a new reference article.

### Removed (noise)
- Duplicate `-- ...Method full image name: ...` second lines in storage method traces (15 sites).
- Stages iterator chronicle (`Set prev stage`, `stage is empty`, image size — 5 lines per stage).
- `-- BuildPhase.prepareStage` entry trace.
- Giant `%#v` dumps: repo tags list (replaced by a tag count line), digest selection result, stages list caches (replaced by stage counts), duplicate manifest-cache debug line next to Info CACHE HIT/MISS.

### Hidden behind existing gates
- Per-file git archive entries → `WERF_TRUE_GIT_DEBUG_ARCHIVE`.
- Per-file ls-tree checksum entries → `WERF_DEBUG_LS_TREE_PROCESS`.
- Container path removal, buildah mount/unmount/per-file checksum → `WERF_CONTAINER_RUNTIME_DEBUG`.
- Registry push/pull progress ticks → `WERF_DOCKER_REGISTRY_DEBUG`.
- Per-match build context paths → `WERF_DEBUG_DOCKERFILE_STAGE_DEPENDENCIES`.

### New topic env vars (same `== "1"` pattern as the ~20 existing gates)
- `WERF_DEBUG_STAGES_STORAGE` — storage method traces, manifest cache ops, per-tag listings, stage/base image description dumps.
- `WERF_DEBUG_CONVEYOR_PHASES` — per-phase LogProcess wrappers of the conveyor.
- `WERF_DEBUG_IMAGE_SPEC` — YAML snapshots of image spec configs.
- `WERF_DEBUG_IMPORT_SERVER` — rsync import server traces.

Stage suitability decisions (`Stage %q is suitable`, `Skip stage ...`) and the `Stage <name> digest <hash>` block with checksum args stay on plain `--log-debug`.

### Fixed
- `yaml.Marshal(stageDesc)` + `panic(err)` ran unconditionally on every selected stage (even without `--log-debug`); now runs only under the gate and returns an error.
- Both `image_spec.go` marshals and the `MutateAndPushImage` config marshal no longer execute in the non-debug path.
- Imperatively driven log processes are gated via `Disable()` (logboek honors `options.Mute()` only in the `Do`/`DoError` path).
- printf verb/arg mismatch in `GetImageNameStageIDCommitList` trace.

### Docs
New bilingual reference page `reference/debug_env_vars.html` (en/ru) listing all debug/verbose env variables with value formats and `--log-debug` requirements, added to the sidebar.

## Why

Debug output was dominated by O(files)/O(tags) per-item lines and unreadable `%#v`/YAML dumps, slowing builds down (synchronous terminal writes, unconditional marshaling) and drowning the useful signal (stage digest arguments, cache decisions).

Note: gates are process-wide, so the hidden channels of shared packages also disappear from plain `--log-debug` of cleanup/purge/bundle publish — same behavior as the existing gates; all variables are discoverable via the new reference article.


alexey-igrychev:
Recreated as a convention-conforming branch: refactor/build/debug-output

Issue #7686, opened 2026-07-28
v3 branch review: 10 confirmed bugs + 16 commit-history/rationale findings


alexey-igrychev:
## Context

Automated review of the full `origin/main..origin/3` diff (v3 branch, ~1050 files / ~52k lines, 112 non-merge commits) via multi-agent code review + adversarial verification, plus a separate pass over commit history for message/rationale quality.

Base: `3657880a1` (main) — Head: `bb5f9e54f` (3, "restore --no-create-namespace and --lookup-resources")

## Priority: merge regression, not an intentional v3 change

- [x] **[high]** **Recover `pkg/host_cleaning/units/*` and related host-cleanup files from `main` — lost across repeated bad merges, not an intentional v3 removal.** Traced the actual history: `main` added absolute-unit support (`units.UnitValue`, e.g. `10GB`) via `c5dbdd0cc feat(host-cleanup): add support for absolute storage units`. The `3` branch never inherited that commit, and across **5 separate `merge main into 3` events** (`bb7651593`, `d215fdd96`, `5bf24c417`, `cbea4e354`, `6438eebb7`, `78ec56614`), the merge result for this file cluster kept resolving back to the stale pre-feature `3`-side version instead of picking up `main`'s changes — even though `3` itself never touched these files (verified via `git diff --name-only <merge-base> <each-parent>` 3-way comparison per merge; no real conflict should have existed). Very likely a stale `git rerere` resolution being blindly reapplied on every subsequent sync. Affected files, entirely or partially stuck at a pre-main-feature state on `3`:
  ```
  pkg/host_cleaning/units/units.go, units_test.go, suite_test.go   (whole package missing on 3, ~224 lines)
  cmd/werf/common/host_cleanup.go
  cmd/werf/host/cleanup/cleanup.go
  pkg/host_cleaning/host_cleanup.go, host_cleanup_test.go
  pkg/host_cleaning/cleanup_report.go
  pkg/host_cleaning/local_purger.go
  pkg/host_cleaning/container_backend_type.go (partially diverged)
  docs/_includes/reference/cli/werf_host_cleanup.md
  ```
[...]

PR #7687, opened 2026-07-28
docs(resources): add v2 to v3 migration guide

alexey-igrychev:
## Summary

Adds a v2→v3 migration guide (EN+RU) modeled on the existing v1.2→v2.0 guide, and fixes stale host-cleanup docs that still described absolute-unit values (`10GB`, `500MiB`) for flags that no longer (or never actually did) accept them.

## Key changes

- `docs/pages_{en,ru}/resources/migration_from_v2_to_v3.md`: new page listing v3 breaking changes — both the ones with a proper `BREAKING CHANGE`/"Breaking change" note in their commit body, and several that shipped without any rationale or breaking-change flag (stapel `import` `image:`→`from:` rename, external `from:` now requiring a tag/digest, nameless-image removal, `--virtual-merge` removal, `AllowMissedSecretKeyMode` removal).
- Registered the new page in `docs/_data/sidebars/{documentation,_documentation}.yml` (en+ru), right after the v1.2→v2.0 entry.
- `docs/pages_{en,ru}/usage/cleanup/host_cleanup.md`: removed the absolute-unit examples for `--allowed-backend-storage-volume-usage(-margin)` (dropped in v3 — code is percentage-only `uint` now) and for `--allowed-local-cache-volume-usage(-margin)` (docs were already stale on v2 — that flag has been percentage-only `uint` since 2021).

## Why

Found while reviewing the full `origin/main..origin/3` diff for #7686: several v3 breaking changes have no commit-message rationale or breaking-change flag, and there was no consolidated migration doc for v2→v3 users. The host-cleanup docs mismatch means users following current docs would set an env var (e.g. `WERF_ALLOWED_BACKEND_STORAGE_VOLUME_USAGE=10GB`) that now causes a panic at flag-registration time on any werf command wiring that flag.

## Review focus / risks

- Migration guide content is a summary of breaking changes I identified by reading commit bodies/diffs across ~112 commits — worth a maintainer pass to confirm nothing user-facing is missing or misdescribed before this becomes the canonical v2→v3 guide.
- Docs-only change; no code touched in this PR (the underlying `--allowed-backend-storage-volume-usage` panic-on-invalid-input bug itself is tracked separately in #7686, not fixed here).

### DIFF
--- a/pkg/stapel/stapel.go
+++ b/pkg/stapel/stapel.go
@@ -13,7 +13,7 @@ import (
 )
 
 const (
-	VERSION              = "0.7.1"
+	VERSION              = "0.7.2"
 	IMAGE                = "registry.werf.io/werf/stapel"
 	CONTAINER_MOUNT_ROOT = "/.werf"
 )
--- a/stapel/Dockerfile
+++ b/stapel/Dockerfile
@@ -32,11 +32,14 @@ RUN apk add --no-cache --virtual .build-deps \
   automake \
   libtool
 
+# rsync 3.1.x routes every chmod through lchmod() when configure finds it, and musl's
+# lchmod always fails with EOPNOTSUPP on a symlink, which breaks --keep-dirlinks imports.
+# Upstream stopped using lchmod on Linux in 3.2.4; the glibc stapel build never linked it.
 RUN cd /tmp && \
   wget https://github.com/RsyncProject/rsync/archive/v3.1.2.tar.gz && \
   tar xzf v3.1.2.tar.gz && \
   cd rsync-3.1.2 && \
-  ./configure --prefix=/usr && \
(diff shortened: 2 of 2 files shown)


alexey-igrychev:
Superseded: the v2→v3 migration guide landed in `3` via #7692 and was extended by 190c6327c (ancestry-check removal, helm-compatible chart default). The remaining commit here removed absolute-unit docs for the host-cleanup flags, which is no longer correct — c34feddad restored `pkg/host_cleaning/units.UnitValue`, so `--allowed-backend-storage-volume-usage`/`--allowed-local-cache-volume-usage` do accept `10GB`-style values. Branch rebased onto `origin/3`, nothing left to merge.

PR #7697, opened 2026-07-29
fix(host): keep git caches of other werf versions until disk pressure

Fral738:
Stop wiping foreign git cache version dirs (git_repos, git_worktrees, git_archives, git_patches) unconditionally on every GC run. On shared hosts several werf versions run side by side, and after a cache version bump every run of the newer werf deleted the cache of the older one, breaking its in-flight builds with errors like "cannot open git work tree: repository does not exist".

Treat each foreign cache version dir as an LRU entry instead: measure its size, derive last access time from the newest last_access_at marker inside (works for both the old flat and the current mirror layouts, falling back to dir mtime), and delete it only under disk pressure, oldest first. Data of werf v1.1 is still kept while v1.1 is in use, as before.

Also rename paths into a unique trash name before removal so concurrent processes see cache entries disappear atomically instead of observing a partially deleted repository.

### DIFF
--- a/pkg/git_repo/gitdata/gc.go
+++ b/pkg/git_repo/gitdata/gc.go
@@ -14,6 +14,7 @@ import (
 	"github.com/samber/lo"
 
 	"github.com/werf/common-go/pkg/util"
+	"github.com/werf/common-go/pkg/util/timestamps"
 	"github.com/werf/kubedog/pkg/utils"
 	"github.com/werf/lockgate"
 	"github.com/werf/logboek"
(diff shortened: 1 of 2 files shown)


alexey-igrychev:
Closing this in favour of a fix for the root cause. The diagnosis here is correct and the unconditional wipe is indeed the trigger, but this approach leaves the incident open:

1. The only protection for a foreign cache version becomes the 3h LRU window (`shouldPreserveGitDataEntryByLru`, `gitdata/git_data_entry.go:17`). A local cache on a CI runner is chronically above the 70% threshold (`host_cleaning/host_cleanup.go:22`) — that is why GC runs at all — and `last_access_at` of a mirror is refreshed at clone/fetch time, i.e. at the start of a build. The failure we are chasing happens on the `gitArchive` stage hours later, so a 3h window still deletes the cache of an in-flight build.
2. `git_archives` and `git_patches` get no grace period at all: they have no `last_access_at` files, their access time lives in `<id>.meta.json` (`LastAccessTimestamp`, `gitdata/git_data_manager.go:121`). `findMaxLastAccessAt` finds nothing, so the code falls back to the version dir mtime, which only changes when top-level entries are added or removed. An actively used foreign archives/patches version therefore looks ancient and is removed on the first disk pressure.
3. A foreign version dir is a single all-or-nothing LRU entry, so when it is picked, the entire cache of the other werf version goes at once rather than as many bytes as needed.
4. The reverse direction stays broken: werf 2.74.x still wipes `git_repos/6` unconditionally, and this change only lives in the newer binary. Until every pipeline moves to a release containing it, builds on the new werf keep losing their mirrors.
5. `.removing.<nanos>` is created in the same parent dir. If the process dies between `os.Rename` and `os.RemoveAll`, the leftover keeps a fresh mtime — so it is protected from LRU eviction — and the next run counts it as yet another foreign cache version.

The follow-up will attack the cause instead: `GitReposCacheVersion` goes back to `5` with the flat v5-compatible layout and shallow mirrors move to a new cache root, so no already-released werf can see, let alone wipe, the new data; on top of that the unconditional wipe is replaced by a staleness check at version-dir granularity with a window measured in days, which also covers archives and patches because it looks at file mtimes rather than `last_access_at` markers. The layout contract will be documented in the `gitdata` package doc.

PR #7698, opened 2026-07-29
fix(build/stages): guard changed paths listing

Fral738:


alexey-igrychev:
Closing: this locks one call site, but it does not cover the failure we are chasing and does not address its cause.

The incident fails in `error checking git mapping emptiness: unable to get latest commit`, i.e. `Remote.IsEmpty` / `Remote.LatestBranchCommit`, and those paths take no GC lock at all — the shared lock is only taken in `Clone` (`git_repo/remote.go:249`), `remote_shallow.go:57`, `initRepoHandleBackedByWorkTree` (`git_repo/remote.go:590`) and two places in `local.go`.

More fundamentally, a shared lock held for the duration of a single operation cannot protect cache data across a build: GC removes a whole cache version dir *between* operations, and holding an exclusive lock for an entire build is not an option. The actual cause is that GC deletes data owned by another werf version; once that is fixed, extra locking is not needed here. If we still want to close the remaining lock gaps, it should be done for all read paths in one pass, with lock ordering reviewed — not for a single call site.

PR #7722, opened 2026-07-30
fix(build, dockerfile): allow dockerfile outside the build context

### LINKED ISSUE #5305: Unable to use relative path for `dockerfile`
### Before proceeding

### Version

1.2.47

### How to reproduce

Using the following image configuration fails

```
image: my-app
context: app
dockerfile: ../build/app.Dockerfile
```

### Result

`Error: unable to load werf config: `dockerfile: PATH` required and should be relative to context!`

### Expected result

Successful build

### Additional information

_No response_

### LINKED ISSUE #7723: Buildah backend applies .dockerignore a second time to the already filtered build context
### Before proceeding

- [x] I didn't find a similar [issue](https://github.com/werf/werf/issues)

### Version

Present on `main` (2.x) and branch `3`, and the same call shape exists in `1.2`.

### How to reproduce

Buildah backend (`WERF_BUILDAH_MODE=native-chroot`), a `contextAddFiles` entry that the context `.dockerignore` matches:

```yaml
# werf.yaml
project: example
configVersion: 1
---
image: app
context: app
dockerfile: Dockerfile
contextAddFiles: [generated.txt]
```

```yaml
# werf-giterminism.yaml
giterminismConfigVersion: 1
config:
  dockerfile:
    allowContextAddFiles: [app/generated.txt]
```

```
# app/.dockerignore
generated.txt
```

`app/generated.txt` is generated and not committed, `app/Dockerfile` is `FROM <base>` + `COPY . /ctx`. Run `werf build`.

### Result

`generated.txt` is missing from `/ctx` in the built image. No error, no warning. With the Docker backend the same project puts it there.

### Expected result

`contextAddFiles` are present in the build context: werf adds them deliberately *after* applying dockerignore, so nothing downstream should filter them out again.

### Additional information

werf assembles and filters the build context itself, then hands the result to the backend. For Buildah it extracts the archive and passes it as a directory (`pkg/container_backend/buildah_backend.go`, `ContextDir: buildContextTmpDir`), while `pkg/buildah/native_linux.go` builds `define.BuildOptions` without `IgnoreFile`. With an empty `IgnoreFile`, buildah resolves the ignore file itself (`imagebuildah/executor.go` → `pkg/parse.ContainerIgnoreFile`): first `<containerfile>.containerignore`/`.dockerignore` next to the containerfile — werf's containerfile is a temporary file outside the context, so there is none — then `<contextDir>/.containerignore`, then `<contextDir>/.dockerignore`, which werf did put into the context from git. The patterns are therefore applied a second time, to an already filtered context, for every Dockerfile build on Buildah.

Consequences beyond the repro above:

- The dockerfile-specific ignore file diverges: werf resolves `<dockerfile>.dockerignore` relative to the context and gives it precedence, buildah looks for it next to the temporary containerfile and never finds it, so it uses the context `.dockerignore` instead. The same `werf.yaml` yields two different contexts depending on the backend.
- `.containerignore` is unknown to werf but takes precedence in buildah, so it adds exclusions on Buildah only — and is not excluded from the context itself.
- Content-based tags stop implying identical content: the stage digest is computed from werf's own context checksum (including the `contextAddFiles` checksum), not from what buildah subsequently drops. The same stage ID can therefore describe images with different layers depending on `WERF_BUILDAH_MODE`, and in a shared stage cache whichever backend built first wins for everyone.

Suggested fix: pass `define.BuildOptions.IgnoreFile` pointing at a no-op ignore file (a single comment line — an empty file may be treated as absent) so that buildah applies nothing, since the filtering is already done by werf. https://github.com/werf/werf/pull/7722 does the equivalent for buildkit by writing a no-op ignore file next to the temporary dockerfile.

Found while fixing the same defect for the Docker backend in #7722. Established by reading the werf → buildah call chain, **not** reproduced on a live Buildah host (the reporter's machine is macOS, where Buildah does not build).

alexey-igrychev:
## Summary

`dockerfile:` pointing outside the build context was rejected at config load time. This PR allows it as long as the path stays within the project directory, and makes both container backends build such an image from a context that only werf has filtered.

Repro (`werf build`):

```yaml
project: example
configVersion: 1
---
image: app
context: app
dockerfile: ../build/app.Dockerfile
```

- Expected: image is built.
- Actual: `Error: unable to load werf config: 'dockerfile: PATH' required and should be relative to context!`

## Key changes

- Config validation: `dockerfile:` may now escape `context:` via `..`; the resolved `context` + `dockerfile` path must still be relative to the project directory, absolute paths are still rejected.
- Build context assembly: a dockerfile outside the context is no longer injected into the context archive, and the "there is no way to ignore the Dockerfile" warning together with its `!<dockerfile>` dockerignore exception rule is skipped for it — neither applies to a file that is not part of the context.
- Docker backend: for a dockerfile outside the context, the extracted context directory is passed as the build context and the dockerfile content is written to a standalone temporary file, instead of streaming the context archive and addressing the dockerfile inside it.
- Both backends now leave the assembly of the build context entirely to werf, by placing the temporary dockerfile in its own directory next to a no-op ignore file:
  - buildkit reads the ignore file adjacent to the dockerfile and only falls back to the context `.dockerignore` when there is none;
  - buildah does the same, and previously received no `IgnoreFile` at all, so it re-applied the ignore files of the context to a context werf had already filtered — for **every** Dockerfile build, not only the ones added here. That stripped `contextAddFiles` again and let the context `.dockerignore` win over a dockerfile-specific one. Closes #7723.
- werf now reads `.containerignore` as well, after `.dockerignore` at each level: `<dockerfile>.dockerignore` → `<dockerfile>.containerignore` → `.dockerignore` → `.containerignore`. Buildah honoured `.containerignore` while werf knew nothing about it, so silencing buildah's filtering without this would have silently widened the context of such projects.
- Docs: the `context` section of the images page and a new section on excluding files from the build context, EN and RU.

## Why

Closes #5305. A shared `dockerfiles/` (or `build/`) directory with a narrow per-image context is a common layout, and `docker build -f ../x/Dockerfile ctx` works, so the werf-side restriction was surprising. It existed only because werf streams the context as an archive on stdin, where a `..` path cannot be addressed — an implementation detail, not a user-facing constraint.

Fixing that exposed a broader inconsistency: werf assembles and filters the build context, but both backends were then filtering it again on their own, each with its own rules. The context of a build now depends only on werf, so the same `werf.yaml` produces the same context on Docker and on Buildah.

### Behaviour change to be aware of

Projects that have a `.containerignore` and no `.dockerignore` will get a different build context than before: on Docker their patterns were not applied at all, on Buildah they were applied outside werf's knowledge. Their stage digests therefore change and their content-based tags move, which means a full rebuild. Projects with a `.dockerignore` keep their current context and digests — hence `.dockerignore` deliberately stays ahead of `.containerignore` at each level, instead of copying buildah's opposite preference.

### Alternative considered and rejected

Injecting the out-of-context dockerfile into the context archive under a synthetic name (`.werf-dockerfile`) and pointing `--file` at it. It keeps the stdin transport and needs no filtering fix, but the synthetic file becomes a permanent part of every user's build context and lands in images via `COPY . .` — werf would be adding its own artifact to the user's image, which is not the same as today's in-context Dockerfile that the user committed there themselves. Two independent reviewers agreed the trade-off is worse than fixing the filtering.

Passing the dockerfile out of band while keeping the stdin archive is not reachable: buildx v0.33 `commands.BuildOptions` exposes neither `DockerfileInline` nor a usable named-context route (named contexts called `dockerfile`/`context` are renamed to `_dockerfile`/`_context`), and in the stdin-archive branch buildx never sets up a separate dockerfile mount.

## Verification

- Docker backend, real daemon (macOS host, Docker 27.4.0), hand-made projects: out-of-context dockerfile builds; the dockerfile does not appear in the context; rebuild reuses the stage and editing the dockerfile invalidates it; `../../x.Dockerfile` and `/etc/passwd` are rejected by validation; an uncommitted out-of-context dockerfile fails giterminism with "must be committed"; the in-context path is unchanged, including the dockerignore warning.
- The same setup proved the second-filtering defect before the fix and parity with the in-context path after it: with `contextAddFiles` plus a matching context `.dockerignore` the added file was silently missing, and a dockerfile-specific `.dockerignore` lost to the context one.
- Buildah backend, Linux host with buildah 1.31.3 and a CGO build of this branch: `task test:e2e paths="./test/e2e/build" labelFilter="dockerfile-outside-context"` passes for both entries.
- Mutation-verified, each mutation reverted afterwards: reverting the validation fails the config table entry; dropping the `isAbsolutePath` check fails the "absolute" entry; dropping the no-op ignore file fails the e2e on `/ctx/file` for Docker and on `/ctx/keepme` for Buildah; reverting the whole Docker-backend branch fails the e2e on "failed to read dockerfile"; dropping the dockerignore-exception guard fails the e2e on the unexpected warning; removing `.containerignore` from the candidates, and hoisting it above the dockerfile-specific file, both fail the e2e.
- Not covered by any test: removing the archive-injection guard in `build_context_archive.go`. It has no observable effect on the image — a `../…` tar entry escapes the extraction directory instead, because `util.ExtractTar` in common-go does not sanitize `..`. The guard prevents that stray write; hardening `ExtractTar` is filed as werf/common-go#34.

## Review focus / risks

- The context of an out-of-context dockerfile build is passed to buildkit as a directory, and for Buildah every dockerfile build has always been. In both cases the no-op ignore file next to the temporary dockerfile is what keeps the backend from re-filtering the context. If a future buildkit or buildah changes that precedence, `contextAddFiles` and dockerfile-specific ignore files would silently regress — the e2e fixture, which combines a context `.dockerignore`, a context `.containerignore`, a dockerfile-specific `.dockerignore` and `contextAddFiles`, is the guard against that.
- Building from a directory requires extracting the context archive, which the Buildah backend already did; for the Docker backend this is a new cost, paid only for out-of-context dockerfiles. Its shared-key is a fresh temp dir, so buildkit's incremental filesync does not kick in and `COPY` layer mtimes come from extraction time.
- No stage digest or tag movement for projects with a `.dockerignore`: the dockerfile path never enters a digest, the in-context matcher is byte-identical, and every new branch sits behind a `filepath.IsLocal` guard on paths that previous versions rejected outright. Projects relying on `.containerignore` are the exception described above.
- Giterminism gets no new options: `.containerignore` files are read through the same reader as `.dockerignore`, so `allowUncommittedDockerignoreFiles` governs both despite its name.


alexey-igrychev:
Out of scope here, but found while working on this: the Buildah backend has the same double-filtering problem for **all** Dockerfile builds, because `define.BuildOptions.IgnoreFile` is never set and buildah then resolves `.containerignore`/`.dockerignore` from the extracted context itself. Tracked in #7723.

PR #7730, opened 2026-07-31
feat(build, buildah): require netavark instead of CNI and slirp4netns

### LINKED ISSUE #6156: Switch to `netavark` network backend for Buildah
### Before proceeding

### Problem

Buildah changed their default network backend to `netavark`. We still use slirp4netns with Buildah.

### Solution (if you have one)

We should consider switching.

### Additional information

This is breaking change.

alexey-igrychev:
## Summary

werf's native Buildah backend now needs the `netavark` binary in a Podman helper binaries directory on
every host; a rootless build that configures a network additionally needs `pasta`. CNI and
slirp4netns are gone. A host without netavark fails at backend initialization, before any build
instruction runs.

Closes #6156. Host preparation instructions: werf/website#849.

### Breaking

- Any build with `WERF_BUILDAH_MODE=native-chroot` or `native-rootless` on a host without `netavark`
  in `/usr/local/libexec/podman`, `/usr/local/lib/podman`, `/usr/libexec/podman` or
  `/usr/lib/podman` fails with `could not find "netavark" in one of [...]`, even when no instruction
  touches the network. `$PATH` is not searched.
  The way out is the distribution `netavark` package; Ubuntu 22.04 has none, so a release binary has
  to be placed in `/usr/local/lib/podman` by hand. CNI cannot be restored — it is compiled out, and
  `CONTAINERS_CONF_OVERRIDE` can usefully override only `default_rootless_network_cmd`.
- A `RUN` in a `native-rootless` build with the default network needs `pasta` from the `passt`
  package, resolved from a helper directory or `$PATH`. `network: host`, `network: none` and every
  `native-chroot` build never invoke it, because buildah only configures a network for a non-host
  namespace, and a missing `pasta` fails at `RUN` time rather than at backend initialization.
  UNVERIFIED on this branch: the rootless build of a network-using `RUN` with netavark in a helper
  directory, pasta on `$PATH` and slirp4netns absent was exercised *before* the Dockerfile-network fix
  below, which changes which namespace that build gets. Re-running it on a host with slirp4netns
  removed would settle it; CI cannot, because its runners install slirp4netns.
- `network:` in `werf.yaml` and `--backend-network` are now honored for non-staged Dockerfile builds on
  the native Buildah backend, where they were silently ignored. Images built with `staged: true` still
  ignore the image-level setting — `DockerfileStageBuilder.Build` never reads it — and only a
  per-instruction `RUN --network=` flag applies there; that gap is pre-existing and tracked as a
  follow-up. That backend supports only `default`,
  `host` and `none`, so a Docker-style value such as `bridge` or a named network is now rejected with
  `unsupported network mode "bridge" for the native Buildah backend (supported: default, host, none)`
  instead of being dropped. The Docker backend is unchanged and still accepts everything Docker does.
- The official Ubuntu werf image is based on Ubuntu 24.04. `build` and `build1001` keep UIDs 1000 and
  1001 and their subordinate ID ranges, because the base image's own UID-1000 `ubuntu` user is
  removed first. VERIFIED: by running the four base images directly; the werf images themselves were
  not rebuilt by hand.
- `docs/pages_{en,ru}/resources/migration_from_v2_to_v3.md` is the authority for what a v2 user has
  to change and gains two entries for this: the netavark and pasta requirements, and the newly
  honored network settings.

### Rootless networking

- A `RUN` in a rootless Dockerfile build now executes in its own network namespace, configured by
  pasta. It previously ran in the **host** network namespace whenever slirp4netns was absent, because
  buildah probes for that binary when the caller passes no namespace options — and werf passed none
  for Dockerfile builds, unlike the Stapel path.
- `network: none` still cannot be honored under `native-chroot`: buildah forces host networking for
  chroot isolation, so a `RUN` there always has the host's network. Stated, not tested.
- The shared network mapper now returns an error where it used to `panic`. Its input used to be
  internal and buildkit-validated; once the free-form `--backend-network` value reaches it, panicking
  on user input would be a crash, so unsupported values produce the error above.
- netavark is located but never executed: chroot builds are host-networked and rootless builds use
  pasta. The new requirement is a lookup upstream performs while constructing the backend.

### Storage and build

- `${graphroot}/defaultNetworkBackend` is neither read nor rewritten — pinning `network_backend`
  bypasses that file. A storage whose sentinel says `cni` keeps saying it, and it still steers the
  *system* podman or buildah CLI sharing that graphroot.
- The `cni` build tag is gone from `cgoTags`, so CNI is compiled out of every cgo build including the
  release binary.
- Buildah test jobs run on ubuntu-24.04 with `netavark` and `passt` installed and
  `kernel.apparmor_restrict_unprivileged_userns=0` set, because noble blocks the unprivileged user
  namespaces rootless buildah needs. `_test_integration_regular.yml`'s runner default moves with it,
  so its test_daily and docs_validate callers move too and the system buildah CLI on those runners
  goes 1.23 → 1.33.
- Unchanged: `WERF_BUILDAH_MODE=docker` and the Docker backend, and the Stapel path's networking,
  which already passed explicit namespace options.

## Why

GitHub's ubuntu runner images dropped the slirp4netns binary, so every rootless Buildah job failed
with `could not find slirp4netns`. The `default_rootless_network_cmd="slirp4netns"` pin and the `cni`
build tag were 2024-era compatibility pins against the buildah 1.35 defaults flip; upstream is
retiring CNI, so leaving them in place means pinning harder against a stack upstream no longer
configures. Installing slirp4netns on the runners is the alternative that was rejected: it unblocks
CI for one runner-image release and leaves the next defaults flip to fight.


alexey-igrychev:
## Verification

- Host with `netavark` hidden from every helper binaries directory: the cgo binary fails at backend
  init with `could not find "netavark"`, before any instruction runs. Both binaries had to be placed
  by hand — neither is apt-installable on 22.04. Unaffected by the network fix below.
- The four base images run directly: netavark resolves from a default helper directory in each
  (alpine 3.18, fedora 38, centos stream9 → `/usr/libexec/podman`; ubuntu 24.04 → `/usr/lib/podman`),
  and `userdel -r ubuntu` on `ubuntu:24.04` exits 0 with only a mail-spool warning.
- The three new `native-rootless` entries execute rather than skipping, which a green badge hides:
  `e2e_extra` ran `E2E Build suite - 50/81 specs` against `47/78` on the run before the test commit —
  +3 selected, +3 executed.
- Mutation: neutered the reject branch of `generateNamespaceOptionsAndNetworkPolicy` → the new
  rejection spec failed with `Expected an error, got nil`.
- Not run: the e2e revert-mutation — dropping the `NamespaceOptions`/`ConfigureNetwork` assignment in
  `NativeBuildah.BuildFromDockerfile` and re-running `e2e_extra` — because it means pushing a
  knowingly broken commit. Expected there: the runners install slirp4netns, so the pre-fix fallback
  host-networks the build and both `none` entries pass instead of failing. The default-network entry
  does not fail on that mutation, which is why the `none` entries carry the guard.

## Review focus

- `pkg/buildah/common.go` — confirm both `[network]` keys have to be pinned: buildah maps an empty
  `default_rootless_network_cmd` to slirp4netns rather than to pasta.
- `generateNamespaceOptionsAndNetworkPolicy` going from `panic` to `error`. It is the single mapper for
  both the Stapel `RUN` path and the Dockerfile path, so both now agree on `default`, `host` and
  `none`; the signature change is what stops a free-form `--backend-network` value from crashing werf.
- Whether rejecting unsupported network modes on Buildah is the behavior you want, given those values
  previously built successfully with host networking. The alternative — silently treating them as
  `default` — keeps the flag lying.
- `slirp4netns` is still installed on the e2e runners, for the *system* buildah CLI the test helpers
  drive; it reads the host containers.conf, not werf's.

## Follow-up

- [ ] BLOCKER: merge werf/website#849 — the host preparation instructions users are pointed at do not
      mention netavark or pasta until it lands.
- [ ] Warn when `network: none` or `host` is requested under `native-chroot`: buildah forces host
      networking for chroot isolation, so the value is silently not honored — the same class of silent
      mismatch this PR fixes for the Dockerfile path, currently only documented in the migration guide.
- [ ] Thread the image-level network setting into the staged Dockerfile path:
      `DockerfileStageBuilder.Build` (`dockerfile_stage_builder.go:62-67`) builds
      `BuildDockerfileStageOptions` without `opts.Network`, unlike its non-staged sibling and
      `StapelStageBuilder`, so `staged: true` images still drop `network:` / `--backend-network`.
      Pre-existing, out of this PR's scope; the migration guides are qualified accordingly.
- [ ] Propose the `BREAKING CHANGE:` footer's release note wording when the release-please PR opens;
      the note is the only mitigation for hosts that upgrade without netavark.

PR #7736, opened 2026-08-01
fix(build): stop re-fetching submodules the checkout already has

### LINKED ISSUE #2257: Git worktree switch procedure with submodules conflicts with gitlab submodule update
When using Gitlab CI auto updating/initializing of git submodules:

```
  GIT_SUBMODULE_STRATEGY: "recursive"
```

— gitlab performs git submodule update procedure by itself.

Werf on the other hand performs submodule update procedure for main repo in the separate worktree, somewhere in `~/.werf/local_cache/.../worktree`.

Then on some gitlab invocation there may occur following error from gitlab submodules update procedure: "cannot chdir to ... No such file or directory"

image

alexey-igrychev:
## Summary

werf re-fetched every submodule from its remote inside its own worktree under `$WERF_HOME`, even when the user or CI had already fetched them. Where credentials are injected per command (GitLab CI with `GIT_SUBMODULE_STRATEGY: recursive`), werf's git subprocess has none and the build fails:

```
fatal: could not read Username for 'https://git.example.com'/: No such device or address
Failed to clone '<submodule>'. Retry scheduled
```

This happens on a cold worktree (clone) and again on every commit that moves a submodule gitlink (fetch), so a long-lived worktree cache does not make it go away.

Fixes #2257 — the credential-prompt symptom, confirmed still current on `3`. The original 2020 "cannot chdir" report is not addressed and no longer reproduces: git keeps the submodule git dir per worktree, so werf does not disturb the shared module store.

### Reuse

- When every submodule of the committed tree, nested levels included, has its pinned commit in a complete local store, werf checks them all out from those objects — no network, no credentials. VERIFIED: the #2257 GitLab shape reproduced by hand against credential-less local remotes.
- An already-populated submodule in werf's worktree is served from the local store when its gitlink moves, not only on its first clone.
- A shallow store (`GIT_SUBMODULE_DEPTH`) stays eligible for reuse; a partial or promisor store does not, because a present commit does not imply present blobs.
- An ambient `insteadOf` prefix rewrite (`GIT_SUBMODULE_FORCE_HTTPS`) does not defeat reuse: werf keys its rewrite on the full submodule URL and git resolves `insteadOf` by longest match.
- Nothing changes about which commit a submodule ends up at — objects are content-addressed, so a served checkout cannot differ from what the remote would have produced — and a repository without submodules is unaffected.

### When reuse does not apply

- Reuse is all-or-nothing: a submodule that is missing locally, in a partial clone, duplicated or unsafely named, whose store path contains `=`, or whose remote URL is shared with another submodule or prefixes a store path, suppresses every redirect and keeps the plain remote update.
- The blockers are named in one `WARNING: unable to reuse local git objects for submodules: …`, printed once per process, which tells the user to run `git submodule update --init --recursive` beforehand.
- `GIT_SUBMODULE_STRATEGY: none`, and any strategy that leaves a submodule uninitialized, still clones from the remote and may still request credentials.
- No build that used to work fails because of reuse: a probe that cannot determine eligibility falls back to the plain remote update, and a store that passes every check yet cannot serve the checkout falls back after discarding the failed attempt, since an initialized submodule is never re-cloned. Every fallback is automatic; there is deliberately no user-facing toggle.
- A committed `.gitmodules` `path` that is absolute or contains `..` is ignored instead of aborting the update, exactly as git's own update ignores it. VERIFIED: with server-side `transfer.fsckObjects=true` git rejects `url = -x` but accepts both path shapes, so such an entry can be pushed.

### Safety

- `protocol.file.allow=always` is added only alongside redirects, and all-or-nothing is what keeps it safe: no URL read from committed `.gitmodules` is used while the file transport is relaxed, so a `file://` submodule URL stays blocked (CVE-2022-39253 class). Ambient trusted `url.*.insteadOf` configuration can still rewrite the store paths werf computes — the invariant covers committed content, not every URL being werf's own.
- The eligibility probe cannot reach a promisor remote: a partial store is rejected before any object lookup instead of relying on `GIT_NO_LAZY_FETCH`, which lookups honor only from git 2.45 while werf supports git 2.18.
- The discard before the fallback removes only what werf's own worktree owns: it resolves the git dir from `<worktree>/.git` on disk rather than with `git rev-parse`, so an inherited `GIT_DIR` cannot redirect it at another repository; it refuses a work tree that is not linked; and it removes a submodule directory only at a work-tree-local path that HEAD records as a gitlink.

## Why

git does not reuse the superproject's `<git-common-dir>/modules/<name>` for a linked worktree — it keeps a separate git dir at `<worktree-gitdir>/modules/<name>`. werf therefore cloned from the remote the first time and, because `git submodule sync` rewrites the recorded remote from committed `.gitmodules` before every update, fetched from the remote whenever a gitlink moved. That redoes work the checkout already did, and where credentials are injected per command it cannot be redone at all. Two redirects cover the two paths: `submodule.<name>.url`, documented as user-overridable before `git submodule update`, for the clone, and `url.<store>.insteadOf` for the fetch of an already-populated submodule.

`<git-common-dir>/modules/<name>` is documented in gitsubmodules(7) as the usual, not the guaranteed, location, and `-c` propagation into nested updates rides on the undocumented `GIT_CONFIG_PARAMETERS`. Both fail safe: a wrong guess fails the pinned-object probe, and anything that slips past every probe falls back to the remotes.

Rejected alternatives: `submodule.alternateLocation`, `--reference` and making the service worktree a main worktree — none of them covers this case. Partial reuse, achievable by splitting the update into a pathspec-limited invocation per group, rescues nothing where credentials are missing, since a submodule absent locally still needs the network; it would only save bandwidth where credentials already work.

### DIFF
--- a/pkg/true_git/submodule.go
+++ b/pkg/true_git/submodule.go
@@ -147,6 +147,11 @@ func submoduleLocalURLOverrides(ctx context.Context, workTreeDir string) (overri
 			return nil, nil, err
 		}
 		if fetchURL == "" {
+			blockers = append(blockers, fmt.Sprintf("%s (submodule has no remote URL to redirect)", node.displayPath))
+			continue
+		}
+		// An earlier run already recorded the store as the remote, so the fetch is local as is.
+		if fetchURL == node.moduleDir {
 			continue
 		}
 		if _, ok := seenURL[fetchURL]; ok {
@@ -155,6 +160,16 @@ func submoduleLocalURLOverrides(ctx context.Context, workTreeDir string) (overri
 		}
 		seenURL[fetchURL] = struct{}{}
 		fetchURLs[node.displayPath] = fetchURL
+	}
+
+	// insteadOf matches by prefix, so a remote URL that prefixes one of the store paths would also
+	// rewrite the store URL the clone path relies on.
+	for _, fetchURL := range fetchURLs {
+		for _, node := range nodes {
+			if strings.HasPrefix(node.moduleDir, fetchURL) {
+				blockers = append(blockers, fmt.Sprintf("%s (remote URL prefixes the local object store path)", node.displayPath))
+			}
+		}
 	}
 
 	if len(blockers) > 0 {
@@ -222,9 +237,14 @@ func walkSubmodule(ctx context.Context, parentModuleDir, parentWorktreeModuleDir
 		return nil
 	}
 
-	worktreeModuleDir := filepath.Join(parentWorktreeModuleDir, "modules", name)
-	if _, err := os.Stat(worktreeModuleDir); err != nil {
-		worktreeModuleDir = ""
+	// Absent parent means the whole subtree is unpopulated in the service worktree; keep the empty
+	// marker instead of joining onto it, which would yield a relative path probed against the CWD.
+	var worktreeModuleDir string
+	if parentWorktreeModuleDir != "" {
+		candidate := filepath.Join(parentWorktreeModuleDir, "modules", name)
+		if _, err := os.Stat(candidate); err == nil {
+			worktreeModuleDir = candidate
+		}
 	}
 
 	*nodes = append(*nodes, submoduleNode{
--- a/pkg/true_git/submodule.go
+++ b/pkg/true_git/submodule.go
@@ -37,14 +37,14 @@ func updateSubmodules(ctx context.Context, repoDir, workTreeDir string) error {
 			return err
 		}
 
-		localURLOpts, remoteOnly, err := submoduleLocalURLOverrides(ctx, workTreeDir)
+		localURLOpts, blockers, err := submoduleLocalURLOverrides(ctx, workTreeDir)
 		if err != nil {
 			return err
 		}
-		if len(remoteOnly) > 0 {
+		if len(blockers) > 0 {
 			logboek.Context(ctx).Warn().LogF(
-				"WARNING: submodules %s are not present in the local object store and will be fetched from their remotes (credentials may be requested).\nRun `git submodule update --init --recursive` in your working tree beforehand so werf reuses the local objects and touches no network.\n",
-				strings.Join(remoteOnly, ", "),
+				"WARNING: unable to reuse local git objects for submodules: %s.\nAll submodules will be checked out from their remotes, which may request credentials. Run `git submodule update --init --recursive` in your working tree beforehand to let werf reuse the local objects instead.\n",
+				strings.Join(blockers, ", "),
 			)
 		}
 
@@ -53,6 +53,8 @@ func updateSubmodules(ctx context.Context, repoDir, workTreeDir string) error {
 		if len(localURLOpts) > 0 {
 			// The overrides point submodule URLs at the on-disk module store, which git clones over
 			// the file transport — blocked by the default protocol.file.allow=user for submodules.
+			// Safe only because overrides are all-or-nothing: every URL used by this invocation is
+			// then a path werf computed itself, never one taken from committed .gitmodules.
 			updateArgs = append(updateArgs, "-c", "protocol.file.allow=always")
 			updateArgs = append(updateArgs, localURLOpts...)
 		}
@@ -73,18 +75,21 @@ type submoduleNode struct {
 	moduleDir   string
 }
 
-// submoduleLocalURLOverrides walks the superproject's submodules recursively and, for those whose
-// pinned commit is already present in the local module store (<parent-module-dir>/modules/<name>,
-// rooted at <git-common-dir>), returns `-c submodule.<name>.url=<local-path>` overrides so
-// `git submodule update --recursive` clones them from those local objects instead of their remotes
-// — no network, no credentials. These -c options propagate to the nested update invocations via
-// GIT_CONFIG_PARAMETERS, so nested submodules are covered too. Submodules missing locally keep
-// their remote URL and are returned in remoteOnly so the caller can warn they will be fetched.
+// submoduleLocalURLOverrides walks the superproject's submodules recursively and, when EVERY
+// submodule's pinned commit is already present in the local module store
+// (<parent-module-dir>/modules/<name>, rooted at <git-common-dir>), returns
+// `-c submodule.<name>.url=<local-path>` overrides so `git submodule update --recursive` checks them
+// all out from those local objects instead of their remotes — no network, no credentials. These -c
+// options propagate into the nested update invocations via GIT_CONFIG_PARAMETERS.
 //
-// A submodule.<name>.url override is global to the git process, so when the same submodule name
-// occurs more than once at different module dirs it cannot be expressed unambiguously; such names
-// are dropped from the overrides and fall back to their remotes instead of risking a wrong source.
-func submoduleLocalURLOverrides(ctx context.Context, workTreeDir string) (overrides, remoteOnly []string, err error) {
+// The overrides are deliberately all-or-nothing: whatever cannot be served locally is reported in
+// blockers and NO override is emitted, leaving the plain remote update werf has always done. Partial
+// reuse is unsafe because `submodule.<name>.url` is global to the git process, so an override for one
+// submodule also redirects a same-named submodule elsewhere in the tree — and a subtree that is not
+// available locally cannot be enumerated at all, so such a collision could not even be detected.
+// All-or-nothing additionally keeps `protocol.file.allow=always`, which the local paths require, from
+// ever applying to a URL werf did not compute itself.
+func submoduleLocalURLOverrides(ctx context.Context, workTreeDir string) (overrides, blockers []string, err error) {
 	names, paths, err := parseGitmodulesPaths(ctx, &GitCmdOptions{RepoDir: workTreeDir}, "config", "-f", ".gitmodules")
 	if err != nil {
 		return nil, nil, fmt.Errorf("unable to list submodules: %w", err)
@@ -98,46 +103,68 @@ func submoduleLocalURLOverrides(ctx context.Context, workTreeDir string) (overri
 		return nil, nil, err
 	}
 
-	var localNodes []submoduleNode
+	var nodes []submoduleNode
 	for i, name := range names {
 		pinnedCommit, err := lsTreeGitlink(ctx, workTreeDir, "HEAD", paths[i])
 		if err != nil {
 			return nil, nil, err
 		}
-		if err := walkSubmodule(ctx, commonDir, name, paths[i], pinnedCommit, &localNodes, &remoteOnly); err != nil {
+		if err := walkSubmodule(ctx, commonDir, name, paths[i], pinnedCommit, &nodes, &blockers); err != nil {
 			return nil, nil, err
 		}
 	}
 
-	nameCount := make(map[string]int, len(localNodes))
-	for _, node := range localNodes {
-		nameCount[node.name]++
-	}
-	for _, node := range localNodes {
-		if nameCount[node.name] == 1 {
-			overrides = append(overrides, "-c", "submodule."+node.name+".url="+node.moduleDir)
-		} else {
-			remoteOnly = append(remoteOnly, node.displayPath)
+	seen := make(map[string]struct{}, len(nodes))
+	for _, node := range nodes {
+		if _, ok := seen[node.name]; ok {
+			blockers = append(blockers, fmt.Sprintf("%s (submodule name %q is used more than once)", node.displayPath, node.name))
 		}
+		seen[node.name] = struct{}{}
 	}
 
-	return overrides, remoteOnly, nil
+	if len(blockers) > 0 {
+		return nil, blockers, nil
+	}
+
+	for _, node := range nodes {
+		overrides = append(overrides, "-c", "submodule."+node.name+".url="+node.moduleDir)
+	}
+
+	return overrides, nil, nil
 }
 
 // walkSubmodule resolves one submodule against the local module store and recurses into its nested
 // submodules, reading their definitions from the module store at the pinned commit (no checkout
 // needed). parentModuleDir is the git dir whose modules/<name> hosts this submodule's store.
-func walkSubmodule(ctx context.Context, parentModuleDir, name, displayPath, pinnedCommit string, localNodes *[]submoduleNode, remoteOnly *[]string) error {
+func walkSubmodule(ctx context.Context, parentModuleDir, name, displayPath, pinnedCommit string, nodes *[]submoduleNode, blockers *[]string) error {
+	// A .gitmodules entry without a gitlink in the tree is not checked out by git either.
 	if pinnedCommit == "" {
+		return nil
+	}
+
+	if submoduleNameUnsafe(name) {
+		*blockers = append(*blockers, fmt.Sprintf("%s (unsupported submodule name %q)", displayPath, name))
 		return nil
 	}
 
 	moduleDir := filepath.Join(parentModuleDir, "modules", name)
 	if !moduleStoreHasCommit(ctx, moduleDir, pinnedCommit) {
-		*remoteOnly = append(*remoteOnly, displayPath)
+		*blockers = append(*blockers, fmt.Sprintf("%s (not initialized locally)", displayPath))
 		return nil
 	}
-	*localNodes = append(*localNodes, submoduleNode{name: name, displayPath: displayPath, moduleDir: moduleDir})
+
+	// A partial clone can hold the commit while its blobs still live on the promisor remote, so
+	// cloning from it would fail or need the network — exactly what reuse is meant to avoid.
+	isPartial, err := moduleStoreIsPartial(ctx, moduleDir)
+	if err != nil {
+		return err
+	}
+	if isPartial {
+		*blockers = append(*blockers, fmt.Sprintf("%s (local clone is partial)", displayPath))
+		return nil
+	}
+
+	*nodes = append(*nodes, submoduleNode{name: name, displayPath: displayPath, moduleDir: moduleDir})
 
 	nestedNames, nestedPaths, err := parseGitmodulesPaths(ctx, &GitCmdOptions{RepoDir: moduleDir}, "config", "--blob", pinnedCommit+":.gitmodules")
 	if err != nil {
@@ -148,12 +175,35 @@ func walkSubmodule(ctx context.Context, parentModuleDir, name, displayPath, pinn
 		if err != nil {
 			return err
 		}
-		if err := walkSubmodule(ctx, moduleDir, nestedName, displayPath+"/"+nestedPaths[i], nestedPin, localNodes, remoteOnly); err != nil {
+		if err := walkSubmodule(ctx, moduleDir, nestedName, displayPath+"/"+nestedPaths[i], nestedPin, nodes, blockers); err != nil {
 			return err
 		}
 	}
 
 	return nil
+}
+
+// submoduleNameUnsafe rejects names that cannot be expressed safely as a `-c submodule.<name>.url`
+// option or that would resolve outside the module store: `=` terminates the config key early and
+// would set an unrelated option, and separators or `.`/`..` components would move the resolved path.
+// .gitmodules is repo-controlled and `git config -f` does not reject these itself.
+func submoduleNameUnsafe(name string) bool {
+	if strings.ContainsAny(name, "=\n/\\") {
+		return true
+	}
+	return name == "." || name == ".."
+}
+
+func moduleStoreIsPartial(ctx context.Context, moduleDir string) (bool, error) {
+	configCmd := NewGitCmd(ctx, &GitCmdOptions{RepoDir: moduleDir}, "config", "--get-regexp", "^(extensions\\.partialclone|remote\\..*\\.(promisor|partialclonefilter))$")
+	if err := configCmd.Run(ctx); err != nil {
+		var exitErr *exec.ExitError
+		if errors.As(err, &exitErr) && exitErr.ExitCode() == 1 {
+			return false, nil
+		}
+		return false, fmt.Errorf("git config command failed: %w", err)
+	}
+	return configCmd.OutBuf.Len() > 0, nil
 }
 
 // parseGitmodulesPaths runs a `git config ... -z --get-regexp \.path$` over a .gitmodules source
@@ -192,7 +242,7 @@ func parseGitmodulesPaths(ctx context.Context, opts *GitCmdOptions, configArgs .
 }
 
 func lsTreeGitlink(ctx context.Context, repoDir, treeish, submodulePath string) (string, error) {
-	lsTreeCmd := NewGitCmd(ctx, &GitCmdOptions{RepoDir: repoDir}, "ls-tree", treeish, "--", submodulePath)
+	lsTreeCmd := NewGitCmd(ctx, &GitCmdOptions{RepoDir: repoDir}, "ls-tree", treeish, "--", ":(literal)"+submodulePath)
 	if err := lsTreeCmd.Run(ctx); err != nil {
 		return "", fmt.Errorf("git ls-tree command failed: %w", err)
 	}
@@ -208,7 +258,8 @@ func moduleStoreHasCommit(ctx context.Context, moduleDir, commit string) bool {
 	if _, err := os.Stat(moduleDir); err != nil {
 		return false
 	}
-	catFileCmd := NewGitCmd(ctx, &GitCmdOptions{RepoDir: moduleDir}, "cat-file", "-e", commit+"^{commit}")
+	// GIT_NO_LAZY_FETCH keeps the probe itself from reaching out to a promisor remote (git 2.41+).
+	catFileCmd := NewGitCmd(ctx, &GitCmdOptions{RepoDir: moduleDir, Env: []string{"GIT_NO_LAZY_FETCH=1"}}, "cat-file", "-e", commit+"^{commit}")
 	return catFileCmd.Run(ctx) == nil
 }
 
--- a/pkg/true_git/submodule.go
+++ b/pkg/true_git/submodule.go
@@ -4,11 +4,14 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"io/fs"
 	"os"
 	"os/exec"
 	"path/filepath"
 	"strings"
+	"sync"
 
+	"github.com/werf/common-go/pkg/util"
 	"github.com/werf/logboek"
 )
 
@@ -39,13 +42,19 @@ func updateSubmodules(ctx context.Context, repoDir, workTreeDir string) error {
 
 		localURLOpts, blockers, err := submoduleLocalURLOverrides(ctx, workTreeDir)
 		if err != nil {
-			return err
+			// Reuse is an optimization, so a failure to even determine whether it applies must cost a
+			// slower build rather than a failed one, exactly like a store that cannot serve the checkout.
+			localURLOpts, blockers = nil, []string{fmt.Sprintf("unable to inspect the local object store: %s", err)}
 		}
 		if len(blockers) > 0 {
-			logboek.Context(ctx).Warn().LogF(
-				"WARNING: unable to reuse local git objects for submodules: %s.\nAll submodules will be checked out from their remotes, which may request credentials. Run `git submodule update --init --recursive` in your working tree beforehand to let werf reuse the local objects instead.\n",
-				strings.Join(blockers, ", "),
-			)
+			// The blockers describe the user's checkout, not this worktree switch, so warning on every
+			// switch of every worktree in a build would repeat the same text many times.
+			submoduleReuseWarnOnce.Do(func() {
+				logboek.Context(ctx).Warn().LogF(
+					"WARNING: unable to reuse local git objects for submodules: %s.\nAll submodules will be checked out from their remotes, which may request credentials. Run `git submodule update --init --recursive` in your working tree beforehand to let werf reuse the local objects instead.\n",
+					strings.Join(blockers, ", "),
+				)
+			})
 		}
 
 		runUpdate := func(reuseOpts []string) error {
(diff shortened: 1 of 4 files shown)


pkg/true_git/submodule.go
```diff
@@ -32,11 +37,416 @@ func updateSubmodules(ctx context.Context, repoDir, workTreeDir string) error {
...
+	}
+
+	worktreeGitDir, err := gitAbsoluteGitDir(ctx, workTreeDir)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	var nodes []submoduleNode
+	for i, name := range names {
+		pinnedCommit, err := lsTreeGitlink(ctx, workTreeDir, "HEAD", paths[i])
```
alexey-igrychev, line 148: BLOCKER: unvalidated `.gitmodules` path reaches `git ls-tree :(literal)…`, which exits 128 for an absolute or `..` path, failing the build where plain `git submodule update` succeeds; add the `filepath.IsLocal` guard that `discardWorktreeSubmoduleGitDirs:252` already has (same at :350 for nested)

pkg/true_git/submodule.go
```diff
@@ -32,11 +37,416 @@ func updateSubmodules(ctx context.Context, repoDir, workTreeDir string) error {
 			return err
 		}
 
-		submUpdateCmd := NewGitCmd(ctx, &GitCmdOptions{RepoDir: workTreeDir}, append(includePathOpts, "submodule", "update", "--checkout", "--force", "--init", "--recursive")...)
-		if err := submUpdateCmd.Run(ctx); err != nil {
-			return fmt.Errorf("submodule update command failed: %w", err)
+		localURLOpts, blockers, err := submoduleLocalURLOverrides(ctx, workTreeDir)
+		if err != nil {
+			return err
```
alexey-igrychev: MAJOR: every probe error is fatal, so the reuse optimization can only ever break builds that used to work; warn and fall through to `runUpdate(nil)` instead, as the PR's own "slower, not broken" rule requires

pkg/true_git/submodule.go
```diff
@@ -32,11 +37,416 @@ func updateSubmodules(ctx context.Context, repoDir, workTreeDir string) error {
...
+	if err != nil {
+		return err
+	}
+	commonDir, err := gitCommonDir(ctx, workTreeDir)
+	if err != nil {
+		return err
+	}
+	// The two are compared through EvalSymlinks because they are produced differently: git resolves
+	// the one, werf joins the other onto a caller-supplied path that may still hold a symlink.
+	if resolvePathForCompare(worktreeGitDir) == resolvePathForCompare(commonDir) {
```
alexey-igrychev: MAJOR: this refusal guards an `os.RemoveAll` but is defeatable by an inherited `GIT_DIR`/`GIT_COMMON_DIR`, which makes the two probes disagree for an unrelated linked worktree; require `worktreeGitDir` to be under `<commonDir>/worktrees/` instead of merely different

pkg/true_git/submodule.go
```diff
@@ -32,11 +37,416 @@ func updateSubmodules(ctx context.Context, repoDir, workTreeDir string) error {
...
+// walkSubmodule resolves one submodule against the local module store and recurses into its nested
+// submodules, reading their definitions from the module store at the pinned commit (no checkout
+// needed). parentModuleDir is the git dir whose modules/<name> hosts this submodule's store.
+func walkSubmodule(ctx context.Context, parentModuleDir, parentWorktreeModuleDir, name, displayPath, pinnedCommit string, nodes *[]submoduleNode, blockers *[]string) error {
+	// A .gitmodules entry without a gitlink in the tree is not checked out by git either.
+	if pinnedCommit == "" {
+		return nil
+	}
 
+	if submoduleNameUnsafe(name) {
```
alexey-igrychev, line 335: MAJOR: the guard against a repo-controlled submodule name injecting a git config key has no test — neutering it leaves the whole suite green; add a table test plus one end-to-end `=`-in-name case

pkg/true_git/submodule.go
```diff
@@ -32,11 +37,416 @@ func updateSubmodules(ctx context.Context, repoDir, workTreeDir string) error {
...
+
+	moduleDir := filepath.Join(parentModuleDir, "modules", name)
+	if !moduleStoreHasCommit(ctx, moduleDir, pinnedCommit) {
+		*blockers = append(*blockers, fmt.Sprintf("%s (not initialized locally)", displayPath))
+		return nil
+	}
+
+	// A partial clone can hold the commit while its blobs still live on the promisor remote, so
+	// cloning from it would fail or need the network — exactly what reuse is meant to avoid.
+	isPartial, err := moduleStoreIsPartial(ctx, moduleDir)
```
alexey-igrychev, line 356: MINOR: four all-or-nothing blockers survive mutation untested (partial-clone here, shared-remote-URL :179, store-prefix :195, no-remote-URL :169); the partial-clone and store-prefix ones are the two the PR itself calls load-bearing and most subtle

pkg/true_git/submodule.go
```diff
@@ -32,11 +37,416 @@ func updateSubmodules(ctx context.Context, repoDir, workTreeDir string) error {
...
+		// disabling reuse outright.
+		if fetchURL == node.moduleDir {
+			continue
+		}
+		if _, ok := seenURL[fetchURL]; ok {
+			blockers = append(blockers, fmt.Sprintf("%s (remote URL is shared with another submodule)", node.displayPath))
+			continue
+		}
+		seenURL[fetchURL] = struct{}{}
+		fetchURLs[node.displayPath] = fetchURL
```
alexey-igrychev: MINOR: `fetchURLs` is keyed by `displayPath`, documented as "used only for messages"; two `.gitmodules` entries sharing a `path` collide and both nodes then emit an `insteadOf` rule built from the other's remote URL — store the fetch URL on `submoduleNode`

pkg/true_git/submodule.go
```diff
@@ -32,11 +37,416 @@ func updateSubmodules(ctx context.Context, repoDir, workTreeDir string) error {
...
+	}
+
+	if len(blockers) > 0 {
+		return nil, blockers, nil
+	}
+
+	for _, node := range nodes {
+		overrides = append(overrides, "-c", "submodule."+node.name+".url="+node.moduleDir)
+		if fetchURL, ok := fetchURLs[node.displayPath]; ok {
+			overrides = append(overrides, "-c", "url."+node.moduleDir+".insteadOf="+fetchURL)
```
alexey-igrychev: MINOR: the store path lands in a config *key*, so a `=` in it makes git reject the whole `-c` (`error: invalid key`), costing a failed reuse plus the destructive discard on every build; validate it like `submoduleNameUnsafe` validates names

pkg/true_git/submodule.go
```diff
@@ -32,11 +37,416 @@ func updateSubmodules(ctx context.Context, repoDir, workTreeDir string) error {
...
+			)
+		}
+
+		runUpdate := func(reuseOpts []string) error {
+			updateArgs := make([]string, 0, len(includePathOpts)+len(reuseOpts)+8)
+			updateArgs = append(updateArgs, includePathOpts...)
+			if len(reuseOpts) > 0 {
+				// The redirects point submodule URLs at the on-disk module store, which git reaches
+				// over the file transport — blocked by the default protocol.file.allow=user for
+				// submodules. Safe only because reuse is all-or-nothing: every URL used by this
```
alexey-igrychev: MINOR: "every URL used by this invocation is a path werf computed itself" is false — an ambient longer-matching `insteadOf` rewrites the computed store path while the file transport is relaxed; the true invariant is "no committed `.gitmodules` URL", state that instead

pkg/true_git/submodule.go
```diff
@@ -32,11 +37,416 @@ func updateSubmodules(ctx context.Context, repoDir, workTreeDir string) error {
...
+// options propagate into the nested update invocations via GIT_CONFIG_PARAMETERS.
+//
+// The overrides are deliberately all-or-nothing: whatever cannot be served locally is reported in
+// blockers and NO override is emitted, leaving the plain remote update werf has always done. Partial
+// reuse is unsafe because `submodule.<name>.url` is global to the git process, so an override for one
+// submodule also redirects a same-named submodule elsewhere in the tree — and a subtree that is not
+// available locally cannot be enumerated at all, so such a collision could not even be detected.
+// All-or-nothing additionally keeps `protocol.file.allow=always`, which the local paths require, from
+// ever applying to a URL werf did not compute itself.
+func submoduleLocalURLOverrides(ctx context.Context, workTreeDir string) (overrides, blockers []string, err error) {
```
alexey-igrychev: MINOR: named returns violate AGENTS.md/CODESTYLE.md ("NEVER use named or naked returns") and no other function in pkg/true_git does this; same at :388

pkg/true_git/submodule.go
```diff
@@ -32,11 +37,416 @@ func updateSubmodules(ctx context.Context, repoDir, workTreeDir string) error {
...
+	}
+
+	*nodes = append(*nodes, submoduleNode{
+		name:              name,
+		displayPath:       displayPath,
+		moduleDir:         moduleDir,
+		worktreeModuleDir: worktreeModuleDir,
 	})
+
+	nestedNames, nestedPaths, err := parseGitmodulesPaths(ctx, &GitCmdOptions{RepoDir: moduleDir}, "config", "--blob", pinnedCommit+":.gitmodules")
```
alexey-igrychev, line 403: MINOR: this prints `error: unable to resolve config blob …` to stderr for every leaf submodule on the happy path, and its exit 1 also silently means "store holds the commit but not its tree", in which case nested submodules are neither enumerated nor blocked

pkg/true_git/submodule.go
```diff
@@ -32,11 +37,416 @@ func updateSubmodules(ctx context.Context, repoDir, workTreeDir string) error {
...
+	return configCmd.OutBuf.Len() > 0, nil
+}
+
+// parseGitmodulesPaths runs a `git config ... -z --get-regexp \.path$` over a .gitmodules source
+// (a working-tree file via -f, or a tree blob via --blob) and returns the submodule names and
+// paths. A missing source or no matches (git exit code 1) yields empty slices, not an error.
+func parseGitmodulesPaths(ctx context.Context, opts *GitCmdOptions, configArgs ...string) (names, paths []string, err error) {
+	args := make([]string, 0, len(configArgs)+3)
+	args = append(args, configArgs...)
+	args = append(args, "-z", "--get-regexp", "\\.path$")
```
alexey-igrychev: MINOR: `\.path$` matches any `<section>.path`, so an `[include] path = /abs/x` line in `.gitmodules` is reported as a submodule and is enough on its own to trigger the BLOCKER above; anchor it as `^submodule\..*\.path$`

pkg/true_git/submodule.go
```diff
@@ -32,11 +37,416 @@ func updateSubmodules(ctx context.Context, repoDir, workTreeDir string) error {
...
 
-		submUpdateCmd := NewGitCmd(ctx, &GitCmdOptions{RepoDir: workTreeDir}, append(includePathOpts, "submodule", "update", "--checkout", "--force", "--init", "--recursive")...)
-		if err := submUpdateCmd.Run(ctx); err != nil {
-			return fmt.Errorf("submodule update command failed: %w", err)
+		localURLOpts, blockers, err := submoduleLocalURLOverrides(ctx, workTreeDir)
+		if err != nil {
+			return err
+		}
+		if len(blockers) > 0 {
+			logboek.Context(ctx).Warn().LogF(
```
alexey-igrychev: MINOR: this 3-line WARNING now fires on every build for a repo cloned without `--recursive` or `GIT_SUBMODULE_STRATEGY: none`, a case that used to be silent and worked; print once per process or lower the level

pkg/true_git/submodule_ai_test.go
```diff
@@ -62,6 +62,462 @@ func TestAI_UpdateSubmodulesForwardsIncludePath(t *testing.T) {
 	require.Contains(t, err.Error(), "werf-test-marker")
 }
 
+func TestAI_UpdateSubmodulesReusesLocalObjectsWithoutRemote(t *testing.T) {
```
alexey-igrychev: MINOR: these 9 new test functions use testify, but AGENTS.md mandates Ginkgo/Gomega with `DescribeTable` for new tests — every hand-written test file in this package is Ginkgo, only the three `*_ai_test.go` files are testify; converting them is also the natural place to close the five untested blocker conditions from findings 4 and 5 (the file's pre-existing 80 lines are out of scope)

alexey-igrychev (commented):
## Independent review — 13 findings

Reviewed against `origin/3`. **The branch was rebased onto `origin/3` as part of this review**: head moved `d1ddf57a2` → `9d41b3f6a`, all 7 commits verified patch-equivalent (`git range-diff` 7/7 `=`), diff unchanged at 3 files / +879/−3. Inline comments are anchored to the rebased head.

**Verified at `9d41b3f6a`:** `task build` exit 0 · `task test:unit paths="./pkg/true_git/..."` passed (63/63 specs) · `task lint:golangci-lint golangciPaths="./pkg/true_git/..."` 0 issues · `task lint` unscoped 0 issues.

### Verdict

**Not mergeable as-is.** The mechanism is sound and, as far as I could establish, the CVE-2022-39253 invariant holds — I could not construct a submodule that `git submodule update --recursive` checks out but the walk misses. But the *probe* that decides whether to use the mechanism can hard-fail a build that works today, on nothing but committed `.gitmodules` content. That directly contradicts this PR's own stated rule ("a slower build rather than a broken one").

### Findings: 13 posted, 13 anchored inline, 0 downgraded

| Tier | Count | Summary |
| :--- | :--- | :--- |
| BLOCKER | 1 | Unvalidated `.gitmodules` `path` reaches `ls-tree :(literal)…` → exit 128 → build fails where plain git exits 0 |
| MAJOR | 3 | Probe errors are fatal (no fallback); the `RemoveAll` guard is env-defeatable; the config-key-injection guard is untested |
| MINOR | 9 | Four more untested blockers, `displayPath` keying, `=` in store path, an overclaiming comment, named returns, `--blob` stderr noise, a too-broad key regex, warning frequency, testify vs mandated Ginkgo |

### Evidence highlights

**The blocker, reproduced at the Go level** with plain git as the baseline in the same test — `git submodule update --checkout --force --init --recursive` exits 0, then `updateSubmodules` fails:

```
--- FAIL: TestZZ_AbsoluteGitmodulesPathBreaksUpdate
    updateSubmodules err = git ls-tree command failed: ... ls-tree HEAD -- :(literal)/abs/evil": exit status 128
      Stderr: fatal: Invalid path '/abs': No such file or directory
--- FAIL: TestZZ_DotDotGitmodulesPathBreaksUpdate
    updateSubmodules err = ... :(literal)../outside": exit status 128
```

Note `discardWorktreeSubmoduleGitDirs` already ships the guard that would prevent this (`filepath.IsLocal`, submodule.go:252) — the probe just doesn't have it. The fix is one line, in one of the two sibling functions.

**Mutation testing** over the full `TestAI_` suite (12 mutants, throwaway copy). All three redirect mechanisms are genuinely load-bearing — good. But five safety guards can be deleted with the suite still green:

```
killed    protocol.file.allow=always   killed  retry-from-remotes      killed  IsLocal guard in discard
killed    submodule.<name>.url         killed  discard-before-retry    killed  gitlink guard in discard
killed    url.<store>.insteadOf        killed  main-worktree refusal
killed    duplicate-name blocker       killed  all-or-nothing suppression
killed    fetchURL==moduleDir skip
SURVIVED  submoduleNameUnsafe guard    ← guards against `=` in a repo-controlled name injecting a config key
SURVIVED  partial-clone blocker        ← the PR calls this load-bearing
SURVIVED  store-prefix blocker         ← the PR calls this "the most subtle part"
SURVIVED  shared-remote-URL blocker
SURVIVED  no-remote-URL blocker
```

The PR says "each of the three mechanisms was mutation-checked" — true, and confirmed. The gap is the guards that make those mechanisms *safe*.

**On `protocol.file.allow=always`:** the committed-`.gitmodules` invariant holds. But the comment claiming "every URL used by this invocation is a path werf computed itself" is false — an ambient longer-matching `insteadOf` rewrites the computed store path while the transport is relaxed (`git -c protocol.file.allow=always -c submodule.one.url=/safe/store/repo -c url.file:///attacker/.insteadOf=/safe/store ls-remote /safe/store/repo` → `upload-pack /attacker//repo`). Ambient config is trusted by git's model, so this is not an escape — but a comment documenting a security invariant should state the one that's true.

### Design proportionality

The two-redirect mechanism is justified — I checked the alternatives before accepting it. `submodule.alternateLocation=superproject` only derives alternates when the superproject itself has them and still contacts the clone source (`fatal: transport 'file' not allowed`, `alternates=NONE`); `--reference` has no per-submodule recursive mapping; main-worktree conversion is a far larger change.

What is disproportionate is the ~420 lines of *policy* around it. Four of the seven commits are "cross-review found N ways the previous commit misbehaves" in the same ~100 lines, and this pass found several more. If the goal is the smallest change that fixes #2257, the probe should shrink — validate paths, be non-fatal on error — rather than grow.

<sub>One caveat on scope: no real GitLab runner or kind cluster; CI shapes were reproduced with credential-less local remotes. macOS, non-CGO. `task format` deliberately not run.</sub>

pkg/true_git/submodule.go
```diff
@@ -32,11 +37,416 @@ func updateSubmodules(ctx context.Context, repoDir, workTreeDir string) error {
...
+		return "", nil
+	}
+	return fields[2], nil
+}
+
+func moduleStoreHasCommit(ctx context.Context, moduleDir, commit string) bool {
+	if _, err := os.Stat(moduleDir); err != nil {
+		return false
+	}
+	// GIT_NO_LAZY_FETCH keeps the probe itself from reaching out to a promisor remote (git 2.41+).
```
alexey-igrychev: **MAJOR (new).** The comment dates `GIT_NO_LAZY_FETCH` to git 2.41+, but `cat-file` only honors it from **2.45** (commit `e6d5479e7a`, based on `2.44.0-rc1`); the 2.39-era appearance of that name was an upload-pack-only backport. Meanwhile `pkg/true_git/init.go:15` requires only **2.18**.

So on any supported git below 2.45 this variable is ignored and `cat-file -e` against a *partial* store lazily fetches the missing commit from its promisor remote — a network round trip and possible credential prompt inside the probe whose whole purpose is to avoid that. The partial-clone blocker cannot intercept it, because `moduleStoreHasCommit` is called at `:312` and `moduleStoreIsPartial` only at `:319`.

**Fix:** test `moduleStoreIsPartial` before `moduleStoreHasCommit`, so an ineligible store is blocked without ever invoking `cat-file`, and correct the version in the comment. Simpler than gating on `gitVersion`.

pkg/true_git/submodule.go
alexey-igrychev, line 335: **Correction to my earlier comment on this line.** I said this guard prevents a repo-controlled name from injecting an *arbitrary* git config option. That is wrong — and this function's own doc comment makes the same overstatement. The `submodule.` prefix is fixed, so `=` in a name cannot reach a key outside `submodule.*`:

```
$ git -c 'submodule.a.url=/evil.url=/store' config --get submodule.a.url
/evil.url=/store          # value corrupted; key unchanged
$ git -c 'submodule.core.pager=/evil' config --get core.pager
                          # unset
```

What the guard genuinely prevents: a repo-controlled *value* for a submodule URL, and traversal out of the module store via `/`, `\`, `.` or `..`, since the name is joined into `moduleDir`. Both worth keeping. My severity stands on the missing test — neutering the guard still leaves the whole suite green — but the mechanism description in my earlier comment was inaccurate, and the doc comment should be corrected too.

alexey-igrychev (commented):
## Follow-up: one new MAJOR, and a correction to my own finding 4

A second review pass turned up one finding I missed and one claim I got wrong. Both are inline below.

**New — MAJOR: the store probe's "no network" guarantee depends on a git version werf doesn't require.** `moduleStoreHasCommit` relies on `GIT_NO_LAZY_FETCH=1`, and the comment dates it to "git 2.41+". `cat-file` only honors that variable from **git 2.45** (commit `e6d5479e7a`, based on `2.44.0-rc1`; the 2.39-era appearance of the same name was an upload-pack-only backport). `pkg/true_git/init.go:15` sets `MinGitVersionConstraintValue = "2.18"`. On any supported git below 2.45 the variable is ignored, so `cat-file -e` against a *partial* store lazily fetches from the promisor remote — a network round trip and possible credential prompt inside the probe whose entire purpose is avoiding exactly that. The partial-clone blocker can't intercept it because of ordering: `moduleStoreHasCommit` is called at `:312`, `moduleStoreIsPartial` only at `:319`. Swapping those two fixes it without a version check.

**Correction — my finding 4 overstated the risk.** I described `submoduleNameUnsafe` as the guard against a repo-controlled name injecting an *arbitrary* git config option. That is wrong, and the guard's own doc comment shares the error. The `submodule.` prefix in `-c submodule.<name>.url=…` is fixed, so a name containing `=` cannot reach any key outside `submodule.*`:

```
$ git -c 'submodule.a.url=/evil.url=/store' config --get submodule.a.url
/evil.url=/store          # value corrupted; key unchanged
$ git -c 'submodule.core.pager=/evil' config --get core.pager
                          # unset — the prefix cannot be escaped
```

The guard is still worth keeping and still untested — but for two narrower reasons: a repo-controlled *value* for a submodule URL, and traversal out of the module store via `/`, `\`, `.` or `..`, since the name is joined into `moduleDir`. My original severity (MAJOR, on the missing test) stands; the mechanism description does not.

Running total: **14 findings** — 1 blocker, 4 majors, 9 minors. Everything else in the earlier review is unchanged.

pkg/true_git/submodule.go
alexey-igrychev: **Correction to my earlier comment on this line.** I recommended requiring `worktreeGitDir` to live beneath `filepath.Join(commonDir, "worktrees")`, with clearing the git environment as optional. That is backwards — containment alone does **not** close this hole, because both probes are redirected together:

```
$ export GIT_DIR=.../victim/.git/worktrees/victim-wt
$ git -C .../service-wt rev-parse --absolute-git-dir
.../victim/.git/worktrees/victim-wt
$ git -C .../service-wt rev-parse --git-common-dir
.../victim/.git
# worktreeGitDir IS under commonDir/worktrees -> containment passes
# RemoveAll still targets the victim
```

The environment-immune signal is the gitdir pointer in the work tree actually being operated on: `cat <workTreeDir>/.git` → `gitdir: .../service/.git/worktrees/service-wt`, correct regardless of `GIT_DIR`.

**Revised fix:** resolve the git dir from `<workTreeDir>/.git`, or run these probes with `GIT_DIR`/`GIT_COMMON_DIR`/`GIT_WORK_TREE` cleared, and verify before deleting. Assert containment afterwards as a secondary check only. Severity and root cause unchanged — only my proposed remedy was inadequate.

alexey-igrychev (commented):
## Correction to my finding 3: the fix I recommended does not actually close the hole

In my first review I said the `os.RemoveAll` guard at `submodule.go:234` should require `worktreeGitDir` to live beneath `filepath.Join(commonDir, "worktrees")`, and treated clearing the git discovery environment as an optional extra. **That ranking was wrong, and the containment check alone is not sufficient.**

Both probes are redirected *together*. With `GIT_DIR` naming an unrelated linked worktree, `--absolute-git-dir` and `--git-common-dir` both resolve inside the *victim* repository, so the containment predicate is satisfied and the deletion proceeds anyway:

```
$ export GIT_DIR=$T/victim/.git/worktrees/victim-wt
$ git -C $T/service-wt rev-parse --absolute-git-dir   # worktreeGitDir
.../victim/.git/worktrees/victim-wt
$ git -C $T/service-wt rev-parse --git-common-dir     # commonDir
.../victim/.git
# is worktreeGitDir under commonDir/worktrees ?  -> YES. Containment passes.
# RemoveAll still targets the victim.
```

The authoritative signal is the gitdir pointer in the work tree werf was actually asked about, which is immune to the ambient environment:

```
$ cat $T/service-wt/.git
gitdir: .../service/.git/worktrees/service-wt     # correct, regardless of GIT_DIR
```

**Revised recommendation:** resolve the git dir from `<workTreeDir>/.git` — or run these two probes with `GIT_DIR`, `GIT_COMMON_DIR` and `GIT_WORK_TREE` removed from the environment — and confirm the result before deleting. Containment beneath `<commonDir>/worktrees` is still worth asserting, but as a secondary check only; it cannot stand alone.

The finding's severity and its underlying cause are unchanged. Only my proposed fix was inadequate, and since you might have acted on it, it seemed worth correcting explicitly.

---

That closes out this review: **14 findings** — 1 blocker, 4 majors, 9 minors — across this and the two earlier reviews. Three of my own items were corrected along the way (finding 4's mechanism, finding 3's fix, and the addition of MAJOR 14); the underlying findings all stand.

pkg/true_git/submodule.go
alexey-igrychev, line 148: **Severity note — the assumption behind this blocker is now measured, and git's own protections do not cover it.**

I originally rated this on an assumed premise (that an absolute or `..` `path` is reachable in practice). Tested with a positive control to prove the server check was armed:

```
# server: transfer.fsckObjects=true
url  = -x         -> REJECTED  gitmodulesUrl: disallowed submodule url: -x   (exit 1)
path = /abs/evil  -> ACCEPTED  (exit 0)
path = ../escape  -> ACCEPTED  (exit 0)
```

`git fsck --strict` behaves the same way: exit 1 on the control, exit 0 on both path variants. git's `.gitmodules` hardening is URL-only; paths are unguarded.

So a contributor PR can land such a `.gitmodules`, after which line 139 exits 128 and aborts the update on a repo where plain `git submodule update --init --recursive` exits 0. Recommendation unchanged (`filepath.IsLocal`, as at line 252) — but "git would have rejected that" is not a downgrade argument.

alexey-igrychev (commented):
## Closing the one open assumption behind BLOCKER 1

When I filed BLOCKER 1 I marked its severity as resting on an *assumed* premise: that a `.gitmodules` whose `path` is absolute or contains `..` is reachable in practice. That premise is now measured rather than assumed, and it holds — including the hostile-PR vector.

The natural mitigation to reach for is that git refuses to accept a malicious `.gitmodules` at all. That protection is real, but it covers **urls only**, not paths. With a positive control to prove the check is actually armed:

```
# control — server has transfer.fsckObjects=true
url = -x        -> remote: error: gitmodulesUrl: disallowed submodule url: -x
                   remote: fatal: fsck error in pack objects          push exit=1  REJECTED

path = /abs/evil -> push exit=0   ACCEPTED
path = ../escape -> push exit=0   ACCEPTED
```

`git fsck --strict` likewise exits 0 on both path variants while exiting 1 on the control. So a contributor's PR can land either `.gitmodules` on a GitHub-style server, and from there `lsTreeGitlink` at line 139 exits 128 and aborts the whole update — on a repository where plain `git submodule update --init --recursive` exits 0.

Nothing changes in the recommendation: the one-line `filepath.IsLocal` guard the sibling function already applies at line 252. What changes is that the severity no longer rests on a likelihood estimate, and "git would have rejected that push" is not available as a reason to downgrade it.

alexey-igrychev:
## The description asserts three safety properties the findings contradict

My 17 inline comments are anchored to code lines, so nothing in them reaches a reader of the description. Raising it separately because werf squashes on merge: this text becomes the commit body and outlives the review in `git log`, so a safety claim that is wrong here is wrong permanently.

**1. "the probe runs under `GIT_NO_LAZY_FETCH` so it cannot itself trigger a fetch"** — stated unconditionally, holds only on git >= 2.45. `cat-file` began honoring that variable in 2.45 (`e6d5479e7a`, based on `2.44.0-rc1`); the 2.39-era same-named backport was upload-pack-only, and `pkg/true_git/init.go:15` still admits git 2.18. On a supported git below 2.45 the probe can contact a promisor remote and prompt for credentials — inside the check whose purpose is to avoid exactly that. Detail in the comment on `submodule.go:437`.

**2. "an unanticipated reuse failure costs a slower build rather than a broken one ... every fallback is automatic"** — true for failures *after* `runUpdate(localURLOpts)` starts (the retry at lines 76-89), not for failures of the probe itself. `submoduleLocalURLOverrides` errors return directly at line 42 with no fallback, and `lsTreeGitlink` at line 139 exits 128 on a `.gitmodules` `path` that is absolute or contains `..` — aborting an update that plain `git submodule update --init --recursive` completes at exit 0. That is the blocker, and it contradicts this sentence rather than merely qualifying it.

**3. "every URL of such an invocation is then a path werf computed itself, never one taken from committed `.gitmodules`"** — the committed-`.gitmodules` half of this held against everything I tried, and is the half that matters for CVE-2022-39253. But "werf computed it" does not follow: trusted ambient `url.*.insteadOf` configuration can rewrite a computed store path before git uses it. The defensible invariant is the narrower one — no URL read from committed `.gitmodules` is used while file transport is relaxed — and stating that instead would cost the claim nothing.

Separately, **"Each of the three mechanisms was mutation-checked"** is accurate — I reproduced all three, and each redirect is genuinely load-bearing. It reads as broader coverage than exists, though: five of the seven all-or-nothing *blockers* can be deleted with the whole `TestAI_` suite still green, including the partial-clone and store-prefix ones this description calls the most subtle part of the change.

Claims 1 and 2 I would call wrong as written; claim 3 is overstated rather than wrong, and you may reasonably read it as shorthand. Nothing here changes the mechanism's justification — I looked for a cheaper alternative (`submodule.alternateLocation`, `--reference`, making the service worktree a main worktree) and none of them covers this case.

alexey-igrychev:
## Verification

- Reproduced the #2257 GitLab shape by hand: a CI-style fetch plus `submodule update --init --recursive`, then a werf-style linked worktree with the remote removed — `exit 1` before the change, checkout from local objects after. Credential-less local remotes stood in for the runner; no real GitLab runner was involved.
- Checked against git 2.50.1 that the mechanism holds for submodule names containing spaces, dots, unicode and a leading dash, and for three-level nesting with every remote deleted.
- Measured that git's `.gitmodules` hardening is URL-only: with `transfer.fsckObjects=true` on the server, `url = -x` is rejected while `path = /abs/evil` and `path = ../escape` are accepted; `git fsck --strict` agrees. That is what makes the pathspec abort reachable from a contributor PR.
- Mutation: reverted the discard to `git rev-parse --absolute-git-dir` → the inherited-`GIT_DIR` spec failed, having deleted another worktree's `modules/`.
- Mutation: dropped the `filepath.IsLocal` guard in `lsTreeGitlink` → the absolute-and-`..` path spec failed.
- Mutation: dropped, one at a time, each of the partial-clone, shared-remote-URL, store-prefix, empty-remote-URL, store-path-`=`, duplicate-name and unsafe-name blockers, the anchored `.gitmodules` regex, the `url.<store>.insteadOf` emission, the `^{tree}` probe and both fallbacks → in each case the matching spec failed.
- Not run: a spec for the once-per-process warning. The `sync.Once` is consumed by whichever spec warns first, so observing a second call needs a package state reset plus logboek capture.

## Review focus

- `insteadOf` matches by URL prefix. Every submodule in the tree gets its own entry, longest match wins, and a remote URL that prefixes a store path blocks reuse — still the most subtle part of the change.
- The discard's own `filepath.IsLocal` check is redundant now that `lsTreeGitlink` guards the same thing; it is kept as defence in depth, and no spec can isolate it while both exist.
- `pkg/true_git/submodule_ai_test.go` is deliberately mixed: the specs this PR adds are Ginkgo/Gomega, the three inherited from #7660 stay testify.

## Follow-up

- [ ] Convert the three testify tests inherited from #7660 in `pkg/true_git/submodule_ai_test.go` to Ginkgo.

PR #7737, opened 2026-08-01
feat(cleanup, host-cleanup): add machine-readable JSON cleanup reports

alexey-igrychev:
## Summary

Two opt-in JSON reports for the cleaning commands, so cleanup results can be consumed without scraping the log: a registry-side report on `cleanup`, `purge` and `host purge`, and a host-side report on `host cleanup` and `host purge`. Both carry `apiVersion: v1`, a `dryRun` marker and a list of deleted items, the registry one also a list of kept items.

### Registry report

- `--save-cleanup-report` / `WERF_SAVE_CLEANUP_REPORT` on `cleanup`, `purge` and `host purge` writes `.werf-cleanup-report.json`; `--cleanup-report-path` / `WERF_CLEANUP_REPORT_PATH` overrides it, appends `.json` when the path has no extension, and rejects any other extension with `invalid --cleanup-report-path "<path>": extension must be either .json or unspecified`.
- Each report names its own schema in `kind` — `CleanupReport` or `HostCleanupReport`. `werf host purge` writes the registry report with `--project-name` and the host report without it, and both spell `command` as `host purge`, so `kind` is what tells a consumer which of the two schemas it was handed and lets the two be versioned independently.
- `kept` and `deleted` hold flat items discriminated by `type`: `stage`, `finalStage`, `customTag`, `rejectedStage`, `rejectedStageMarker`, `imageMetadata`, `managedImage`. `kept` covers protected stages in both the main and the final repository.
- The report is written even when the run fails partway, and the command then exits non-zero with both the run error and any write error. A cleanup that deletes objects and then fails still reports what it deleted.
- Writability of the destination is checked before anything is deleted, and the file is replaced by renaming a fully written sibling, so an interrupted run leaves the previous report intact rather than a truncated one.
- `repo`, plus `finalRepo` when a final repository is configured and `metaRepo` when `--meta-repo` points the metadata storage elsewhere. Which address an item was deleted from follows from its `type`: `stage`, `rejectedStage` and `rejectedStageMarker` live in `repo`, `finalStage` in `finalRepo`, `managedImage` and `imageMetadata` in `metaRepo`, and a `customTag` spans two — the tag is deleted from `repo` and only its registration from `metaRepo`.
- Under `--dry-run` the report lists what would be deleted and `dryRun` is `true`; with the flag off no file is written and an existing one is left alone.
- Deletion order inside `deleted` is nondeterministic: entries are appended from `parallel.DoTasks` callbacks under a mutex that guarantees safety, not order.

### Host report

- `--save-host-cleanup-report` / `WERF_SAVE_HOST_CLEANUP_REPORT` on `host cleanup` and `host purge` writes `.werf-host-cleanup-report.json`; `--host-cleanup-report-path` / `WERF_HOST_CLEANUP_REPORT_PATH` overrides it with the same extension rule.
- `deleted` items are typed `image`, `container` or `volume`. An `image` item always carries the backend image `id`, plus `reference` when the removal was addressed to a tag, so a multi-tagged image yields one item per tag sharing one `id`.
- `spaceReclaimed` sums the four `RunGC` phases and matches the log's "Total freed space". It is omitted entirely when the command frees space without measuring it, as `host purge` does: absent means not measured, `0` means measured as zero.
- Deletions performed by a phase that later fails are still recorded, so a partially failed `host cleanup` reports the containers and images it really removed.
- Auto host cleanup triggered by `build`/`converge` never writes a report: the detached child's argv pins `--save-host-cleanup-report=false`, which beats the `WERF_SAVE_HOST_CLEANUP_REPORT` it inherits through `os.Environ()`.
- `host purge` warns when a report flag has no effect on the run it was given — `--save-cleanup-report` without `--project-name`, `--save-host-cleanup-report` with it, and either one on the buildah backend, which cannot clean local storage.

### Deliberately unchanged

- Log and colour output is byte-identical: `pkg/cleaning/style.go` and every `LogFWithCustomStyle` call are untouched, so the documented colour-grep recipes keep working. The new docs section sits above the existing one, in both `pages_en` and `pages_ru`, and the CLI reference pages are regenerated.
- `deleteCustomTags` still logs a tag as deleted when its deletion failed; the report omits it, because the report is meant to list what actually happened.

### Developer-visible

- A new `test/e2e/cleanup` suite covers the registry report end to end: it builds an image into the test registry, purges it with `--save-cleanup-report` and asserts the written document — the `v1` schema tags, the `repo` address, and stage tags that also appear in the purge log. The host report gets no e2e entry, because `werf host purge` writes it only without `--project-name`, and in that mode it wipes every werf image, werf home and stapel on the host, which would break the sibling suites sharing the runner.
- The two report-path tests in `cmd/werf/common` are ginkgo specs rather than plain Go tests. The package's ginkgo bootstrap declares the suite, and in ginkgo's parallel mode — what `task test:unit` uses — a plain Go test in such a package never runs: the two tests reported `Ran 0 of 0 Specs — SUCCESS!` while a mutation that removed the extension validation went unnoticed. As specs they run, and that mutation now fails them.

## Why

The only documented way to get cleanup results out of werf is grepping ANSI colour codes from the log to build a `--keep-list`, which breaks under `--log-color-mode=off`, in non-tty output, and on any change to log formatting. `build` and `converge` have had report flags for a long time; the cleaning commands had none.

The reports live in a new `pkg/cleanup_report` rather than one type per subsystem because `pkg/cleaning` (a registry, through `StorageManager`) and `pkg/host_cleaning` (a local container backend) share nothing, and a shared package is what keeps them from importing each other. Both report types are nil when the flag is off and every method early-returns on a nil receiver, so no recording site needs a guard and each one stays a single line next to the log call it mirrors.

A report that only appears after a fully successful run would be worthless for the case that motivates it, which is why saving is unconditional and every partial phase result is recorded before its error propagates.


cmd/werf/cleanup/cleanup.go
```diff
@@ -252,8 +260,13 @@ func runCleanup(ctx context.Context, cmd *cobra.Command) error {
...
+		Report:                                  report,
 	}
 
 	logboek.LogOptionalLn()
-	return cleaning.Cleanup(ctx, projectName, storageManager, cleanupOptions)
+	if err := cleaning.Cleanup(ctx, projectName, storageManager, cleanupOptions); err != nil {
+		return err
+	}
+
+	return report.Save(reportPath)
```
alexey-igrychev: **The report is thrown away when the cleaning run fails**
All five call sites:

```go
if err := cleaning.Cleanup(ctx, projectName, storageManager, cleanupOptions); err != nil {
    return err
}
return report.Save(reportPath)
```

- `cmd/werf/cleanup/cleanup.go:271`
- `cmd/werf/purge/purge.go:169`
- `cmd/werf/host/cleanup/cleanup.go:130`
- `cmd/werf/host/purge/purge.go:105` and `:138`

A cleanup that irreversibly deletes 500 stages and *then* hits a registry 500, a rate limit, or a propagated SIGINT writes no report at all. Worse: a pre-existing `.werf-cleanup-report.json` from an earlier run stays on disk and is indistinguishable from this run's output, so a CI job that reads it acts on stale data.

This is exactly the case the feature exists for. The PR's own motivation — "the only documented way to extract cleanup results is grepping ANSI colour codes out of the log" — applies with more force after a partial failure than after a clean one, because the log is the only surviving record and it is the thing being replaced.

Fix: save unconditionally and join the errors, e.g.

```go
runErr := cleaning.Cleanup(ctx, projectName, storageManager, cleanupOptions)
return errors.Join(runErr, report.Save(reportPath))
```

(SIGKILL is out of scope — nothing survives that.)

pkg/host_cleaning/local_backend_cleaner.go
```diff
@@ -331,6 +336,7 @@ func (cleaner *LocalBackendCleaner) RunGC(ctx context.Context, options RunGCOpti
 
 		logboek.Context(ctx).LogF("Freed space: %s\n", logging.RedF("%s", humanize.Bytes(reportWerfContainers.SpaceReclaimed)))
 		logDeletedItems(ctx, reportWerfContainers.ItemsDeleted)
+		recordBackendPruneReport(options.Report, cleanup_report.ItemTypeContainer, reportWerfContainers)
```
alexey-igrychev: **Partial phase reports are discarded inside `RunGC`, so even an unconditional `Save` would under-report**
- `pkg/host_cleaning/local_backend_cleaner.go:339` — the container phase does `if err != nil { return fmt.Errorf("unable to remove werf containers: %w", err) }` *before* `recordBackendPruneReport`.
- `cleanupWerfContainers` appends each removed container at `:484` (`report.ItemsDeleted = append(report.ItemsDeleted, container.ID)`) and then discards all of it at `:492`, returning `backendPruneReport{}` — an **empty** report — if the follow-up `volumeutilsGetVolumeUsageByPath` fails.
- `cleanupWerfImages` (`:552`) does better, returning the partial `report`, but `RunGC:358` discards it on error anyway.

So a host cleanup that removes 40 images and then fails a disk-usage read reports zero of them. Same root cause as finding 1, different layer: record each successful deletion into the `HostReport` as it happens (or merge the partial phase report before propagating its error), instead of only after the whole phase succeeded.

pkg/host_cleaning/host_purge.go
```diff
@@ -22,6 +25,7 @@ func HostPurge(ctx context.Context, backend container_backend.ContainerBackend,
 		RmForce:                       true,
 		RmContainersThatUseWerfImages: options.RmContainersThatUseWerfImages,
 		DryRun:                        options.DryRun,
+		Report:                        options.Report,
```
alexey-igrychev, line 28: **`spaceReclaimed` is structurally always `0` for `werf host purge`**
`HostPurge` (`pkg/host_cleaning/host_purge.go:28`) passes the report only to `containersRemove` and `imageReferencesRemove`; neither measures bytes. `AddSpaceReclaimed` is called exclusively from `recordBackendPruneReport`, which only `LocalBackendCleaner.RunGC` reaches. And the field is declared non-optional at `pkg/cleanup_report/cleanup_report.go:101` — `SpaceReclaimed uint64` with a plain `json:"spaceReclaimed"` tag, no `omitempty` — so it cannot even be absent. `werf host purge --save-host-cleanup-report` therefore emits a list of deleted objects next to `"spaceReclaimed": 0` after freeing gigabytes.

The docs sentence — "`spaceReclaimed` accounts for the container backend only" — does not cover this: `host purge` *is* the container backend. Encoding "unavailable" as `0` in a machine-readable field is a wrong value, not a missing one. Either omit the field when nothing measured it (`*uint64` + `omitempty`) or say so explicitly in the docs.

pkg/cleanup_report/cleanup_report.go
```diff
@@ -0,0 +1,159 @@
...
+}
+
+type Report struct {
+	mux sync.Mutex
+
+	APIVersion string `json:"apiVersion"`
+	Command    string `json:"command"`
+	DryRun     bool   `json:"dryRun"`
+	Repo       string `json:"repo"`
+	FinalRepo  string `json:"finalRepo,omitempty"`
```
alexey-igrychev, line 56: **The report has no `metaRepo` field, but most of what it reports was deleted from there**
Both `werf cleanup` and `werf purge` register `--meta-repo` (`SetupMetaRepo`, `cmd/werf/common/common.go:572` at `f387085`), which points project metadata at a *separate* registry; `StorageManager.GetMetaStorage()` returns it when set and falls back to the stages storage otherwise (`pkg/storage/manager/storage_manager.go:194-199`).

The report records only `repo` and `finalRepo` (`pkg/cleanup_report/cleanup_report.go:45`, filled from `GetStagesStorage().Address()` and the final storage at `cmd/werf/common/common.go:254-259`). Nothing carries the meta-repo address — yet four of the seven registry item types (`managedImage`, `imageMetadata`, `customTag`, `rejectedStageMarker`) are deleted from meta storage, not from `repo`: `ForEachRmImageMetadata` and `ForEachRmManagedImage` both operate on `GetMetaStorage()` (`storage_manager.go:1017`, `:1027`), and custom-tag cleanup unregisters its metadata there too (`:1051`). Run `werf cleanup --repo registry/app --meta-repo registry/meta --save-cleanup-report` and the report claims `"repo": "registry/app"` while listing objects that were removed from `registry/meta`.

The items themselves are accurate; their *location* is unrecoverable from the report. That is exactly what a machine-readable deletion record is for, and `apiVersion: v1` freezes the field set. Fix is three lines: a `MetaRepo string` field next to `FinalRepo` tagged `json:"metaRepo,omitempty"`, populated from `GetMetaStorage().Address()` only when `MetaStorage != nil` — when it is nil the address equals `repo` and the field stays absent, so no noise for the common case.

pkg/cleanup_report/cleanup_report.go
```diff
@@ -0,0 +1,159 @@
...
+}
+
+func writeJSON(path string, report any) error {
+	data, err := json.MarshalIndent(report, "", "\t")
+	if err != nil {
+		return fmt.Errorf("marshal cleanup report: %w", err)
+	}
+	data = append(data, '\n')
+
+	if err := os.WriteFile(path, data, 0o644); err != nil {
```
alexey-igrychev: **No writability preflight — the destructive run happens, then the write fails**
`--cleanup-report-path` extension validation happens early (good: before any deletion). Writability does not: `os.WriteFile` at `pkg/cleanup_report/cleanup_report.go:154` is the first time the path is touched. `werf cleanup --save-cleanup-report --cleanup-report-path /nonexistent/dir/r.json` deletes everything it planned to and *then* exits non-zero with a write error and no report.

Note the tension with finding 1: opening the target file up front would truncate a still-valid previous report. Write to a sibling temp file (created before the run to prove writability) and `os.Rename` it into place at the end — that also makes the report atomic, which `os.WriteFile` is not.

pkg/host_cleaning/images.go
```diff
@@ -129,6 +130,8 @@ func imageReferencesRemove(ctx context.Context, backend container_backend.Contai
 				return fmt.Errorf("container_backend rmi: %w", err)
 			}
 		}
+
+		options.Report.AddDeleted(cleanup_report.Item{Type: cleanup_report.ItemTypeImage, ID: ref})
```
alexey-igrychev: **`type: "image"` `id` mixes image IDs and repo references**
`host cleanup` records real IDs (`local_backend_cleaner.go:544` `imgSummary.ID`, plus backend prune report IDs); `host purge` records whatever `imageReferencesRemove` was handed — a repo reference like `registry/app:tag`, falling back to `img.ID` only for dangling/tagless images (`pkg/host_cleaning/images.go:134`). A multi-tagged image also yields one `image` entry per tag, so `deleted | length` over-counts images. Nothing in code or docs promises `id` is a content ID, so this is schema ambiguity rather than a bug — but `apiVersion: v1` freezes it. Worth splitting (`id` + `reference`) or documenting now.

cmd/werf/host/purge/purge.go
```diff
@@ -58,6 +58,12 @@ func NewCmd(ctx context.Context) *cobra.Command {
 	common.SetupContainerRegistryMirror(&commonCmdData, cmd)
 
 	common.SetupDryRun(&commonCmdData, cmd)
+
+	common.SetupSaveCleanupReport(&commonCmdData, cmd)
+	common.SetupCleanupReportPath(&commonCmdData, cmd)
+	common.SetupSaveHostCleanupReport(&commonCmdData, cmd)
```
alexey-igrychev, line 65: **Report flags are silently ignored on `werf host purge` depending on `--project-name`**
`werf host purge` registers all four flags (`cmd/werf/host/purge/purge.go:62-65`), but the `if projectName == ""` branch at `:93` picks exactly one report: `--save-host-cleanup-report` is silently ignored when `--project-name` is given, `--save-cleanup-report` when it is not. No file, no warning, no error. The Buildah early return at `:110` also exits 0 without writing the registry report the user explicitly asked for. Same applies to the env-var forms. A warning ("`--save-cleanup-report` has no effect without `--project-name`") is enough.

pkg/cleaning/cleanup.go
```diff
@@ -980,6 +1007,7 @@ func (m *cleanupManager) cleanupUnusedStages(ctx context.Context) error {
 				logboek.Context(ctx).Default().LogProcess("%s (%d)", reason, stageDescSetToKeep.Cardinality()).Do(func() {
 					for stageDescToKeep := range stageDescSetToKeep.Iter() {
 						logboek.Context(ctx).Default().LogFWithCustomStyle(keptStyle, "%s\n", stageDescToKeep.Info.Tag)
+						m.report.AddKept(cleanup_report.Item{Type: cleanup_report.ItemTypeStage, Tag: stageDescToKeep.Info.Tag, Reason: reason.String()})
```
alexey-igrychev: **`kept` omits protected final-repo stages**
`cleanupFinalStages` marks final stages as protected but never calls `AddKept`, while deleted final stages do get `type:"finalStage"` (`pkg/cleaning/cleanup.go:614`, `:632`). A consumer reading `kept` as "what survived" cannot see final-repo survivors. This mirrors the log exactly (there is no `keptStyle` line for final stages either), so it is defensible — but then say so in the docs, because `deleted` is symmetric and `kept` is not. Anchored at `cleanup.go:1010`, the existing `AddKept` site.

pkg/cleanup_report/cleanup_report.go
```diff
@@ -0,0 +1,159 @@
...
+		return
+	}
+
+	r.mux.Lock()
+	defer r.mux.Unlock()
+
+	r.Deleted = append(r.Deleted, items...)
+}
+
+func (r *Report) Save(path string) error {
```
alexey-igrychev: **`pkg/cleanup_report` public API takes no `context.Context`**
CODESTYLE.md:24 — "All public functions and methods must accept `context.Context` as the first parameter." None of `NewReport`, `NewHostReport`, `AddKept`, `AddDeleted`, `AddSpaceReclaimed`, `Save` do (`pkg/cleanup_report/cleanup_report.go:50, 62, 73, 84, 105, 114, 125, 136`). `Save` is the one that matters — it does file I/O and would plausibly want to log. (The companion CODESTYLE rule about an `Options` struct does *not* apply here: `NewReport`'s four arguments are all required.)

pkg/cleaning/cleanup.go
```diff
@@ -595,13 +599,19 @@ func (m *cleanupManager) deleteStages(ctx context.Context, stageDescSet image.St
...
+func deleteStageDescSet(ctx context.Context, storageManager manager.StorageManagerInterface, dryRun bool, deleteStageOptions manager.ForEachDeleteStageOptions, stageDescSet image.StageDescSet, isFinal bool, report *cleanup_report.Report) error {
+	stageItemType := cleanup_report.ItemTypeStage
+	if isFinal {
+		stageItemType = cleanup_report.ItemTypeFinalStage
+	}
+
 	if dryRun {
 		for stageDesc := range stageDescSet.Iter() {
 			logboek.Context(ctx).Default().LogFWithCustomStyle(deletedStyle, "%s\n", stageDesc.StageID.String())
+			report.AddDeleted(cleanup_report.Item{Type: stageItemType, Tag: stageDesc.StageID.String()})
```
alexey-igrychev: **Dry-run and real runs derive the `deleted` stage tag differently**
`pkg/cleaning/cleanup.go:614` uses `stageDesc.StageID.String()`; `:632` uses `stageDesc.Info.Tag`. Verified equivalent in practice — the storage tag is constructed from the stage ID, and the `CreationTs == 0` case degrades to the digest on both paths — and it mirrors the two log lines. Consistency nit only; the documented keep-list workflow reads `kept`, which always uses `Info.Tag` (matching `markWhitelistStagesAsProtected`).

pkg/host_cleaning/host_cleanup.go
```diff
@@ -61,10 +64,14 @@ func RunAutoHostCleanup(ctx context.Context, backend container_backend.Container
...
 	var args []string
 
+	// exec.Detach passes os.Environ() to the fork, so WERF_SAVE_HOST_CLEANUP_REPORT would otherwise
+	// enable the report here and write it long after the parent command exited. Pinned off like
+	// --dry-run and --force, which are spelled out for the same reason.
 	args = append(args,
 		"host", "cleanup",
 		fmt.Sprintf("--dry-run=%v", options.DryRun),
 		fmt.Sprintf("--force=%v", options.Force),
+		"--save-host-cleanup-report=false",
```
alexey-igrychev, line 63: **No regression test for the detached auto-host-cleanup argv pin**
`pkg/host_cleaning/host_cleanup.go:74` appends `--save-host-cleanup-report=false` because `exec.Detach` inherits `os.Environ()`, so `WERF_SAVE_HOST_CLEANUP_REPORT` would otherwise make a detached child write a report long after the parent exited. The PR verified this by hand and says so honestly, but the argv is built inline with no seam, so nothing stops a regression. Extracting the argv build into a function and asserting on the slice is a small, cheap test.

docs/pages_en/usage/cleanup/cr_cleanup.md
```diff
@@ -244,6 +244,64 @@ werf uses the _GitLab container registry API_ or _Docker Registry API_ (dependin
...
 
+### Cleanup report
+
+The `--save-cleanup-report` option makes `werf cleanup` write a machine-readable JSON report of what was kept and what was deleted. The report path is `.werf-cleanup-report.json` by default and is configured with `--cleanup-report-path` (the extension must be `.json`).
+
+Generating a keep-list then takes a single `jq` call:
+
+```bash
+werf cleanup --repo registry.mydomain.com/app --dry-run --save-cleanup-report
+jq -r '.kept[].tag' .werf-cleanup-report.json > keep-list.txt
```
alexey-igrychev: **Documented `jq` recipe pulls custom tags into the keep-list**
`jq -r '.kept[].tag'` (docs/pages_en/usage/cleanup/cr_cleanup.md, and the ru mirror) also emits `type:"customTag"` entries. Harmless — `markWhitelistStagesAsProtected` matches against `Info.Tag` and custom tags never match — but `jq -r '.kept[] | select(.type=="stage") | .tag'` is the honest recipe.

alexey-igrychev (commented):
## Review: machine-readable JSON cleanup reports

Reviewed at `f387085`, all anchors re-validated on `cf6820b8` after the rebase. **13 findings: 1 blocker, 5 major, 7 minor** — 12 posted as line comments below.

The design is sound: nil-report-as-off keeps each recording site to a single guard-free line, records sit next to the existing coloured log lines (all 11 sites covered), and the log output is byte-identical, so the documented grep recipes keep working. The problems are about what happens when a run does **not** succeed, plus three `apiVersion: v1` schema details.

### Already done on this branch

- **Rebased onto `3`** — it was 36 commits behind. The rebase was mechanical: those 36 commits touch none of the PR's 28 files, and `git diff` of the rebased range is byte-identical to the pre-rebase one, so every finding below carries over unchanged.
- **Finding 13, fixed in `cf6820b8`** — the two new `cmd/werf/common` report-path tests never ran. `cmd/werf/common/suite_test.go` called `RunSpecs` while the package declares no ginkgo specs, and ginkgo's parallel mode (the default for `task test:unit`) then skips the package's plain Go tests entirely: `Ran 0 of 0 Specs — SUCCESS!`. Deleting the report-path extension validation left the suite green. The cause is pre-existing rather than introduced here, and it silenced the package's other tests too (`TestKeyValueArrayToMap_*`, `Test_hashOriginUrl`, telemetry). Dropping the empty bootstrap makes all of them run, and the mutation now fails the two tests that cover it.

### The headline problem

Findings 1 and 2 are one bug at two levels: a run that deletes objects and *then* fails writes no report at all, and leaves any earlier report at the path looking current — exactly the situation the feature exists for. Saving unconditionally and returning `errors.Join(runErr, saveErr)` at the five call sites fixes the top level; `RunGC` needs the same treatment one level down.

Findings 3, 4 and 6 are `apiVersion: v1` schema decisions: cheap now, expensive once released. Finding 5 is an operational hazard on the same destructive path as finding 1.

### Evidence

The tests were checked by mutation rather than by reading assertions — five load-bearing tests each had a plausible fault injected and each failed for the right reason: recording a custom tag unconditionally, dropping the `r == nil` guard, removing all six mutex pairs (562/1000 items recorded — it discriminates even without `-race`, which `task test:unit` does not enable), removing the report-path validation, and skipping `AddSpaceReclaimed`. Unmutated, `pkg/cleanup_report`, `pkg/cleaning` and `pkg/host_cleaning` are green. Check-gaming screen clean: the one modified existing test is a pure `cleanupReport` → `backendPruneReport` rename — normalising the names makes the 18 removed and 18 added lines identical — with no skipped specs, no test-only branches, and no changes to CI or linter config.

**Not verified:** `task build` and `task lint` were run only for this branch's own fix, not across the whole tree; the Buildah early return in finding 7 is a control-flow reading, since Buildah does not compile on macOS; and no cleanup was executed against a real registry or container backend, so findings 3, 4 and 5 are derived from control flow rather than observed output.

alexey-igrychev:
## Verification

- `werf host cleanup --dry-run` in a scratch dir: no file without the flag, a valid `v1` payload with it. With `WERF_SAVE_HOST_CLEANUP_REPORT=1` exported, the detached fork wrote a report before the argv pin and stopped after it.
- Mutation: record a custom tag unconditionally instead of only on success → `TestAI_deleteCustomTags_ReportSkipsFailedDeletion` fails.
- Mutation: drop the `r == nil` guard in `Report.AddKept` → `TestAI_NilReportIsNoOp` fails with a panic.
- Mutation: remove all six `mux.Lock`/`Unlock` pairs → `TestAI_ConcurrentAdd` fails with 562 of 1000 items recorded, so it discriminates even without `-race`, which `task test:unit` does not enable.
- Mutation: return the path instead of the extension error → the `GetCleanupReportPath` and `GetHostCleanupReportPath` specs fail. The same mutation passed unnoticed while those two were plain Go tests in a ginkgo package, which is how that was found.
- Mutation: skip `AddSpaceReclaimed` in `recordBackendPruneReport` → the host-report spec fails, 41 of 42 specs passing.
- Mutation: record the container phase only after it returns nil, and return an empty report when the follow-up usage read fails → the `should keep already removed containers in the report when the phase then fails` spec fails, which is the regression it exists for.
- Mutation: drop `--save-host-cleanup-report=false` from the detached argv → both `autoHostCleanupArgs` specs fail.
- Mutation: give both schemas the same `kind` → `TestAI_KindDistinguishesTheTwoSchemas` and `TestAI_HostReportJSON` fail.
- The three specs that pin the exact serialised JSON failed when `kind` was added and were updated deliberately, which is the point of pinning the whole document rather than individual fields.
- `deleteStageDescSet`'s dry-run branch had no test at all, so one was added asserting that a planned deletion names the same tag a real one would; without it the two derivations could drift unnoticed.
- Not run: no e2e coverage for either report — the repo has only `test/legacy_e2e/suites/cleanup`, no modern cleanup suite.
- Not run: the Buildah paths, which do not compile on macOS, so the `host purge` Buildah branch is unexercised.

## Review focus

- `apiVersion: v1` freezes the shape on release. It now names itself with `kind`, omits `spaceReclaimed` rather than reporting a false `0`, separates image `id` from `reference`, and adds `metaRepo`. Two independent reviews of the schema argued to keep all four; the one change both rejected was a per-item `repo` field, because the address is derivable from `type` and the one case motivating it — `customTag` — spans two registries and so could not be expressed by a single per-item address anyway.
- The documented type-to-registry map was wrong for two of four types and is corrected in `f95c7e3b`: `rejectedStageMarker` is a `<stage>-rejected` tag removed from the stages storage, and a `customTag` deletes the tag from `repo` while unregistering only its registration from `metaRepo`. Since an item does not name its own registry, that map is the contract, so it is worth a second pair of eyes.
- `errors.Join(runErr, saveErr)` changes what these commands return: a cleanup that both fails and cannot write its report now surfaces both errors rather than only the first.
- `pkg/cleanup_report` public functions take `context.Context` first per CODESTYLE.md, though only `Save` has a use for one today.

## Follow-up

- [x] Add e2e coverage for the registry report. `test/e2e/cleanup` builds an image into the test registry, purges it with `--save-cleanup-report` and asserts the written document. Verified on a Linux host against a real registry: dropping the `AddDeleted` for a deleted stage while keeping its log line fails the spec on empty stage tags, recording a tag that differs from the deleted one fails the log-parity check, dropping the `managedImage` record fails the item-type assertion, and giving the registry report the host `kind` fails the schema assertion; unmutated it passes.
- [ ] Add e2e coverage for the host report. `werf host purge` writes it only without `--project-name`, and in that mode it wipes every werf image, werf home and stapel on the host, so the spec needs a runner it does not share with the other e2e suites.

alexey-igrychev:
All thirteen findings from the review above are addressed.

| # | Finding | Fix |
| :-- | :-- | :-- |
| 1 | Report lost when the run fails | Saved unconditionally at all five sites, `errors.Join(runErr, saveErr)` |
| 2 | `RunGC` discards partial phase reports | Each phase records before its error propagates; container and volume phases return the accumulated report |
| 3 | `spaceReclaimed: 0` on `host purge` | `*uint64` with `omitempty` — absent means not measured |
| 4 | No `metaRepo` | Recorded when the meta storage address differs from `repo` |
| 5 | Writability found only after deleting | `CheckWritable` before the run, `os.Rename` of a written sibling after |
| 6 | `image` `id` meant two things | `id` is always the backend id, new `reference` carries the tag |
| 7 | Report flags silently ignored on `host purge` | Warns for each inapplicable flag, including the buildah branch |
| 8 | `kept` omitted protected final stages | Recorded as `finalStage` |
| 9 | No `context.Context` on the public API | Added, per CODESTYLE.md |
| 10 | Two derivations of the stage tag | Both paths use `Info.Tag` |
| 11 | Argv pin untested | Extracted to `autoHostCleanupArgs`, two specs cover it |
| 12 | `jq` keep-list recipe caught custom tags | `select(.type == "stage")`, both languages |
| 13 | Two new tests never ran | Rewritten as ginkgo specs |

Each fix that changes behaviour has a test that fails without it — the mutations are listed in the verification comment. `task build`, `task lint` (0 issues) and the full `task test:unit` are green.

Findings 3, 4 and 6 change the `v1` shape, so they are worth confirming before release; everything else is behaviour-preserving for a successful run.


The branch has since been rebased onto `3` for `01eeb94f`, which adds ginkgo specs to `cmd/werf/common`; the commit hashes above no longer resolve. Finding 13's fix changed shape with it: the empty bootstrap those specs now need is kept, and the two report-path tests are ginkgo specs instead.

alexey-igrychev:
rejected

PR #7739, opened 2026-08-01
feat(build): add per-project meta-repo safeguard and migration

alexey-igrychev:
## Summary

werf records a per-project marker in `--repo` naming the `--meta-repo` that holds the project's metadata, and refuses to run when a later invocation disagrees with it. Previously, dropping `--meta-repo` from one pipeline step silently sent werf to read metadata from `--repo`, where the managed-images set is incomplete — and the next `werf cleanup` deleted images that were still in use. Two new commands, `werf meta-repo migrate` and `werf meta-repo detach`, move existing metadata into a meta-repo and remove the marker.

### Safeguard

- A project whose marker names a meta-repo fails every storage-touching command run without `--meta-repo`, with an error naming the recorded address and `werf meta-repo detach` as the way out. This ends the previous silent fallback, so a pipeline that relied on it now fails instead — the behavior being fixed. Only setups already passing `--meta-repo` are affected, and that flag exists solely in unreleased v3, so no released setup changes.
- Passing `--meta-repo` to a project whose metadata still sits in `--repo` no longer fails: the new metadata goes to the meta-repo immediately. `migrate` is still what brings the existing records over, and nothing enforces that it ran — a `cleanup` before it decides from an incomplete metadata view, which the docs state.
- Passing a `--meta-repo` that disagrees with the marker fails the same way; a `--meta-repo` that canonically resolves to `--repo` is validated too, and gets its own error rather than being told to pass the flag it already passed.
- Address comparison is tag-insensitive: `--meta-repo repo:tag` and `--repo repo` compare equal.
- The marker is written lazily, on the first real metadata write, so read-only commands never push and pull-only credentials keep working.
- `cleanup.disable: true` together with `--meta-repo` now writes neither a cleanup record nor a marker; previously the meta-repo gained a record while `--repo` kept no marker, so a later run without the flag passed validation and silently read `--repo`.
- The marker tag is `werf-meta-repo-marker-<project>`, which matches no tag parser in werf v2 or v3, so a v2 running against the same `--repo` ignores it.

### `werf meta-repo migrate`

- Takes the source in `--from` and destination in `--to`, moves the `managed-image-*`, `meta-*`, `custom-tag-meta-*` and `cleanup` records, then plants the marker; it copies, verifies the copy, and only then deletes. `--from` also exposes the source registry credential flags (`--from-docker-hub-token`, `--from-quay-token`, …) that the default delete needs on Docker Hub, Harbor and Quay. Every flag is scoped to this command through `WERF_META_REPO_MIGRATE_FROM[_*]` and `WERF_META_REPO_MIGRATE_TO` env vars, so the generic `WERF_FROM`/`WERF_TO` used by another copy command cannot leak in. `--repo` and `--meta-repo` are not accepted by this command.
- `--remove-source` defaults to `true` (`$WERF_REMOVE_SOURCE`), so the command moves rather than copies; `--remove-source=false` copies and leaves `--from` intact.
- Matches metadata by record shape, not by the `werf` label: a repository holds one project, and records written before werf started labeling them (2021) carry no label. Previously migrate refused outright to delete any unlabeled record, so the default `--remove-source` failed on exactly the long-lived repositories the meta-repo is meant to drain.
- Never treats a stage image as a metadata record, so a custom tag alias named like a metadata record (`--add-custom-tag managed-image-release`) is left untouched instead of being copied and deleted — deleting it would have destroyed the stage, because the registry drops the manifest by digest. Multi-platform aliases are covered: an index carries no top-level labels, so it is recognized as a stage by being an index.
- The `--meta-repo` record wins for the two families whose content is mutable under a stable tag (`cleanup`, `custom-tag-meta-*`), since that is the copy werf reads; the `--repo` one is skipped with a warning naming the tag and both addresses. Under `--remove-source=false` the records left in `--repo` are frozen after the first migration, so the reverse rule pushed a stale stage ID back over the current one and `cleanup`, which groups aliases by that stage ID, could delete a custom tag pointing at a live stage.
- Refuses to run when `--to` is the same repository as `--from`.

### `werf meta-repo detach`

- Removes the marker only, and says so: metadata is NOT moved back, and the command is not a reverse migration.
- Refuses to delete anything if the marker tag holds a stage image or custom tag alias rather than a marker.
- `werf purge` removes the marker as its last step, once every other deletion has succeeded, so a re-created project builds without `--meta-repo` again and never needs `detach` for that. A purge that fails partway leaves the marker in place, keeping metadata still sitting in the meta-repo protected. It is a no-op for the non-registry stages storage `werf host purge` operates on.

### Docs

- `docs/_includes/reference/cli/werf_meta_repo*` and `docs/pages_en/reference/cli/werf_meta_repo*` are the generated reference for the new command group, added by `task doc:gen` along with their sidebar and overview entries.
- `docs/pages_{en,ru}/usage/build/process.md` is the authority for the safeguard, and states the motivation as tag-count: image-metadata records are written per image, per commit and per stage, and every stage lookup lists the whole tag set of `--repo`.

## Why

`--meta-repo` (#7637) split metadata into a second repository but left the two repositories with no link between them, so the correct `--meta-repo` value existed only in each pipeline's flags. Nothing detected its absence: `GetOptionalMetaStorage` falls back to the stages storage, so werf read an empty managed-images set from `--repo` and `cleanup` treated every image whose metadata it could not see as unmanaged. The cost is deleted in-use images, discovered only when a deploy pulls a missing tag.

The marker is per-project rather than per-repository because werf never enforces one project per repository: a repository-wide marker would let one project's `werf purge` drop the safeguard for every other project in that `--repo`. `detach` is not called `reset`, because `reset` reads as the inverse of `migrate` — metadata moved back — which is exactly the expectation that ends in deleted images.


alexey-igrychev:
## Verification

- `task doc:gen` rewrites twelve unrelated `werf_*.md` files when `WERF_DISABLE_AUTO_HOST_CLEANUP` is set in the environment, since the flag default is rendered from it. Regenerated with `env -u` so the diff holds `werf_meta_repo_migrate.md` only.
- Checked the claim that werf v2 ignores the marker tag against `origin/main`: its record prefixes are `managed-image-`, `meta-`, `custom-tag-meta-`, `cleanup`, `import-metadata-`, `client-id-`, `sync-server` and the `-rejected` suffix, none of which matches `werf-meta-repo-marker-`.
- Rebased onto `origin/3` with zero conflicts: all 7 original commits preserved and the changed-file set byte-identical before and after.
- `werf meta-repo migrate` and `detach` both died with a SIGSEGV before doing any work, so both new commands shipped unrunnable in the pre-rebase branch. `NewStorageManagerWithOptions` builds the final, secondary and cache stages storages unconditionally, while these two commands register only the `--repo` flag group. No unit test reaches that constructor, so a fully green unit suite said nothing about whether the commands start; only the legacy e2e suite caught it. Fixed by nil-guarding the three getters, matching the guard `GetOptionalMetaStorage` already had.
- e2e needed a delete-enabled registry: stock `registry:2` answers DELETE with `UNSUPPORTED: The operation is unsupported`, so anything exercising `--remove-source` or the marker deletion needs `REGISTRY_STORAGE_DELETE_ENABLED=true`. The `purge` marker case was run against such a registry via `WERF_TEST_K8S_DOCKER_REGISTRY`, outside CI's matrix.
- Mutations, each reverted after being observed to fail for the right reason:
- Mutation: restore the `--meta-repo == --repo` early return → equal-address-with-marker spec failed.
- Mutation: clear the meta storage's cleanup-disabled state → `writes neither record nor marker when cleanup is disabled` failed on the record assertion.
- Mutation: unlabeled-record check passed through → both refusal specs failed; narrowed to key-absence only → only the present-but-empty spec failed, so the two specs discriminate the two shapes.
- Mutation: delete before verifying the destination copy → `keeps the source when the destination is absent despite a reported copy` failed. This ordering is invisible to the e2e: when the copy succeeds the observable end state is identical either way, so it is covered at unit level only.
- Mutation: restore the ownership-only skip → the `cleanup` timestamp spec and the retargeted custom-tag spec failed.
- Mutation: restore the `werf`-label ownership test in the record predicate → the `foreign werf label` case failed.
- Mutation: restore the refusal to delete unlabeled records → `moves an unlabeled legacy record like any other` failed.
- Mutation: restore the refusal to enable `--meta-repo` while `--repo` still holds metadata → `configured + no marker + metadata already in --repo → decorated` failed.
- Mutation: return the full reference instead of the repository from `CanonicalRepoAddress` → the tag-dropping and port cases failed.
- Mutation: restore the source-wins rule for the two mutable families → `keeps the destination cleanup record over the source one` and `keeps the retargeted destination custom-tag metadata over the source one` both failed.
- Mutation: drop the `IsIndex` term from stage recognition → all three multi-platform collision specs failed; drop the content-digest term → the two single-manifest collision specs failed.
- Mutation: force `RemoveSource: true` in the e2e → the `--remove-source=false` case failed with `managed-image records kept in stages-repo`.
- Mutation: drop the marker-deletion step from `purge` → `purge MUST remove the meta-repo safeguard marker` failed in the cleanup suite.
- Mutation: delete the `PutImageMetadata` override from the decorator — which still compiles, since the embedded interface supplies it → only that table entry failed, so the entries discriminate per method rather than failing together.
- Mutation: remove one name from the decorator's method classification, standing in for a method arriving in `PrimaryStagesStorage` unclassified → the method-set assertion failed.
- Not run: no mutation covers the `CleanupDisabled` wiring line itself. `RepoStagesStorage.cleanupDisabled` is private with no accessor and `CreateStagesStorage` takes no injectable registry, so a test in `cmd/werf/common` can neither read the field nor observe the push; widening visibility for the test alone was not worth it. The structural guard is the required parameter — omitting it no longer compiles — and the contract it feeds is covered in `pkg/storage`.

## Review focus

- `pkg/storage/meta_repo_marker.go` — `isMetadataRecord` and `isStageImageOrAlias` are the one choke point separating werf's dummy metadata records from real stage manifests. Getting it wrong in either direction is data loss: too strict orphans metadata that `cleanup` still reads, too loose deletes stage images by digest.
- `metaRepoMarkerStorage` embeds `PrimaryStagesStorage`, so a metadata-writing method added later is forwarded unwrapped and bypasses the marker with no compile error. The guard is a test, not the compiler: `classifies every PrimaryStagesStorage method` pins the interface's method set against an explicit split, and the table above exercises each guarded method. Check that split is where a new method would actually be classified correctly.
- `docs/_includes/reference/cli/werf_meta_repo*` and `docs/pages_en/reference/cli/werf_meta_repo*` are `task doc:gen` output for the new command group, new in full against `3`; read the commands' help text in `cmd/werf/meta_repo` instead.

alexey-igrychev:
## Verification

- Mutation: `--from` address env `WERF_META_REPO_MIGRATE_FROM` → `WERF_FROM` → `scopes source and destination address env to the command` failed.
- Mutation: `--from-quay-token` env → `WERF_FROM_QUAY_TOKEN` → `exposes command-scoped source registry credential flags but not the destination` failed.
- Mutation: `--from` back to `OnlyAddress` (no credential flags) → credential-flags spec failed.
- `./bin/werf meta-repo migrate --help`: `--from`/`--to` default from `WERF_META_REPO_MIGRATE_FROM`/`WERF_META_REPO_MIGRATE_TO`, `--from-*` credentials from `WERF_META_REPO_MIGRATE_FROM_*`, no `--to-*` credentials, no `--repo`/`--meta-repo`.
- Reviewed EN/RU v2-to-v3 guide diffs: transition applies only with v3 `--meta-repo`; source records removed by default; restriction is on `cleanup`, not `purge`.

## Review focus

- Independent review flagged the missing source-registry delete credentials (default `--remove-source` calls the registry delete API); fixed and now command-scoped.

## Not run

- Live registry e2e against Docker Hub for the credentialed delete path (needs a real Docker Hub account); exercised by the fake-registry cleanup suite only.

PR #7749, opened 2026-08-03
docs(dev): spec PR descriptions instead of narrating the patch

alexey-igrychev:
## Summary

A PR description is now the spec of the change: `Summary`, `What`, `Why`, and nothing else. Verification, review guidance and follow-up work move to the author's first comment.

## What

- the description carries three sections, named verbatim, and omits the ones with nothing to say; `Verification`, `Review focus` and `Follow-up` go in the author's first comment, skipped when empty
- `What` takes one falsifiable behavior claim per line, one level deep, naming every user-visible surface with its default — flag, annotation, env var, `werf.yaml` field, log or error text, exit code
- a claim carries its own blast radius and confidence: `BREAKING:` names who it breaks and the way out, `VERIFIED:` names a check CI cannot repeat, `UNVERIFIED:` says nothing settled it
- every risk is a claim, so `git log` shows where a change bites; `Review focus` keeps only where to look, and a safety property nobody established stays there as a question
- the trivial tier is decided by a test — nothing observable, and no marker to hang — instead of by a list of examples, which keeps one-line release-workflow edits and skill edits out of it
- PRs stay draft, and only the user marks one ready
- the review-time half no longer reaches `git log`: it is reachable through the `(#NNNN)` in the squash subject, and the two artifacts have to be kept in sync on every push
- UNVERIFIED: nothing enforces the honesty of a marker or the completeness of a surface list — these are rules of wording, and only review catches an author who skips them

The file in this diff is the authority for the rest; the claims above are what changes for someone writing a PR, not an index of the rules that moved.

## Why

werf squashes with the description prefilled as the commit body, so a description outlives its review by years: 29% of the characters in the bodies of the last 40 commits were `## Verification` and `## Review focus`. The old criterion — write what the reviewer cannot infer from the diff — made the diff the source of truth and the description a supplement, which is what produced descriptions narrating the patch instead of the behavior. Two earlier attempts tied length to the diff size (#7715) and to the number of unreviewable items (#7741); both measured something other than what a reader needs.

Blast radius and confidence ride on each claim rather than in a `Risk` section of their own, because a separate section drifts from the claims it describes. The review-time half lives in a comment rather than below a separator line in the body, because a separator has to be cut by whoever merges, while GitHub cannot pull a comment into a squash body in the first place — the kernel's `---` convention with a different enforcement mechanism. Four rules come from `Documentation/process/submitting-patches.rst` directly: rejected-alternative rationale, self-containment, `Fixes:` with a commit reference, and splitting the patch instead of shortening the description.


alexey-igrychev:
## Verification

- Dry-run on eleven PRs across three rounds — this is the only check that exists, since nothing here is executable. Round one took five open PRs of different shapes (#7669, #7730, #7725, #7712, #7377); round two picked six more to break specific new rules (#7675, #7736, #7548, #7673, #7678, #7718); round three re-derived five of them under the trimmed rules and diffed the output against the earlier drafts.
- Two independent agent reviewers, then a necessity audit run with the instruction to assume a reader hostile to size. Every finding is folded in; the audit cut a fifth of the file without losing a rule.
- Not run: nobody but me has written a PR to this form. This description and #7739 are its first two uses.

## Review focus

- This PR's own `## What` is the only in-repo sample of the format. If it reads as patch narration, the rules failed.
- A fourth `## How` section was proposed and A/B-tested on five PRs, then rejected: its content was either behavior already claimed in `What`, or narration its own gates failed to stop, and the small-fix control leaked. Recorded so the question is not reopened blind.

## Follow-up

- [ ] Document the convention in `CONTRIBUTING.md`, once it has been used on enough PRs to be worth teaching — it currently covers only the PR title, so an external contributor has no way to learn it from the repo.

PR #7751, opened 2026-08-03

alexey-igrychev:
## Summary

Brings `main` into `3`. Four commits, two of which change behavior: parallel builds on the Buildah backend no longer pull the same base image several times at once, and PR descriptions follow a new form.

## What

- with the native Buildah backend, images sharing a base pull it once instead of once per goroutine: pulls are serialized per reference and the resulting image id is cached for the run, so a parallel build no longer races on the same base image
- a PR description is now the spec of the change — `Summary`, `What`, `Why`, with review-time material in the author's first comment — as `.agents/skills/pull-request/SKILL.md` in this merge defines it
- `feat`/`fix` commit subjects name the user-visible outcome rather than the mechanism, since `CHANGELOG.md` is generated from them verbatim
- `3`'s copy of the skill keeps this branch's exclusion of `werf/3p-helm` from the lists of werf-owned dependencies, which the incoming version had reintroduced
- this supersedes `Why` before `Key changes` (#7726, restored again in c895dd060): the new form has no `Key changes` at all, and `Why` comes last

## Why

`3` is 176 commits ahead of `main` and takes its fixes through a periodic merge, so the Buildah pull fix and the agent conventions only reach the v3 line this way. The `Why`-first decision is reversed because the section it was ordered against no longer exists: claims now carry behavior, and reasoning reads after them rather than before.


alexey-igrychev:
## Verification

- Not run: no build against a container backend, so the Buildah pull serialization is carried on its own PR's testing (#7664) rather than re-checked here.
- The only hand-resolved conflict is `.agents/skills/pull-request/SKILL.md`, taken from `main` and then stripped of `werf/3p-helm` in the two lists that name werf-owned dependencies; `git show origin/3:go.mod | grep 3p-helm` returns nothing, which is why f7c090bb7 removed it from this branch in the first place.

## Review focus

- the skill file is the whole content conflict — everything else auto-merged, including `pkg/container_backend/buildah_backend.go`.
- `Why` moving after `What` reverses a decision this branch made twice (#7726 and c895dd060). If the reasoning-before-claims order still has support, this is the place to say so.

PR #7752, opened 2026-08-04
docs(dev): validate a verification check before citing its result

alexey-igrychev:
## Summary

Agents verifying a change in this repo prove conditions from shell output — a match count, or a command "succeeding". Both can report success while being structurally incapable of reporting failure: a mis-typed `grep -c` pattern yields the same `0` as a genuinely absent condition, and `… | grep x || echo clean` always prints `clean` because a pipeline returns only its last command's status. `AGENTS.md` now forbids citing such a check until a failing case has been shown to look different from a passing one.

## What

- The delta in what an agent produces: a count- or status-based verification is now accompanied by a control proving the check *can* fail, or restructured so failure is visible (`if` / `$?` instead of `|| echo`). Neither a bare count nor a bare "the command succeeded" is acceptable as evidence.
- VERIFIED: the rule's own text is checked the way it prescribes — `^- NEVER` returns 15 hits as a known-positive control, `^- NEVER cite a count` returns 1, and the superseded wording returns 0.
- No `task` target, `.golangci.yml` rule or CI check is added: the rule governs commands an agent composes in a shell at runtime, which leave no artifact for a check to read.
- The `NEVER trust LSP/gopls diagnostics … as proof of build failure` rule directly above is unchanged; the new line joins it as the same class of rule.

## Why

Three incidents, two mechanisms. First: patch-equivalent commits counted with `git range-diff … | grep -c '  = '` returned 0 of 7 and read as a broken rebase. The rebase was correct — git's column spacing differed from what the pattern assumed.

Then, within the hour of the first wording landing, its own author broke the rule twice more by a different route: `… | grep -E '<pattern>' || echo none` and `… | grep … | head || echo accepted`. A pipeline reports only its last command's status, so the `||` branch can never run and the reassuring message always prints. One of the two produced a false `rejected` in a push test that was deciding a finding's severity.

The first wording named `grep -c`/`wc -l` only and would not have caught either recurrence. Both mechanisms share one shape and one direction of failure: the check does not merely stop working, it reports success. That is worse than having no check, because the result then gets cited as proof.

Rejected alternative: a linter or `task` target flagging `grep -c` or `|| echo`. These commands are composed ad hoc in a shell and never committed, so there is no artifact for a check to inspect.


alexey-igrychev:
## Verification

- Rule text checked the way the rule prescribes: `^- NEVER` → 15 hits (known-positive, proves the grep works against this file), `^- NEVER cite a count` → 1 hit, superseded wording → 0 hits. The pipe-masking clause was checked with `if grep -q`, deliberately not with `|| echo`.
- `git diff --check` clean; net diff against `3` is 1 insertion, 0 deletions.

## Review focus

- Placement. In **Highest-priority rule**, beside the LSP-is-not-proof rule, because both say an indirect signal must not substitute for the authoritative one. **Verifying changes (MANDATORY)** is a defensible alternative home — it is where the authoritative `task` commands live.
- The narrowness question raised on the first revision is settled and needs no further input: the wording *was* too narrow, two recurrences demonstrated it within the hour, and this revision covers both mechanisms. Recorded here because the branch was force-pushed and the earlier reasoning no longer matches the diff.

PR #7754, opened 2026-08-04
chore(dev): stop the caller's WERF_* leaking into generated CLI docs

alexey-igrychev:
## Summary

Regenerating the CLI reference no longer picks up flag defaults from whoever runs it. `scripts/docs/regen.sh` already cleared `HOME`, `DOCKER_DEFAULT_PLATFORM` and `WERF_PLATFORM`, but any other `WERF_*` variable in the caller's environment still became a documented default: with `WERF_DISABLE_AUTO_HOST_CLEANUP=1` exported, one `task doc:gen` rewrote `--disable-auto-host-cleanup=true` into twelve unrelated command pages.

```
WERF_DISABLE_AUTO_HOST_CLEANUP=1 task doc:gen && git status --porcelain docs
```

## What

- `task doc:gen` produces identical output regardless of which `WERF_*` variables the caller has exported; previously each one silently became a flag default in every page listing that flag.
- The corruption was indistinguishable from ordinary generator output in review — twelve one-line changes in pages the actual work never touched.
- VERIFIED: reproduced the twelve-file rewrite, applied the fix, reran with the same variable exported for a clean tree, and reran in a clean environment to confirm the generator is still idempotent.
- `AGENTS.md` gains four facts this session had to learn by hitting them: a green `task test:unit` does not prove a command starts, since nothing in the unit suite constructs the storage manager and a command missing a flag group crashes before doing any work; `git diff --check` cannot gate the whole repo, since the reference generator emits column-aligned help and its pages carry trailing whitespace on every branch; e2e against a local registry needs `REGISTRY_STORAGE_DELETE_ENABLED=true`, because stock `registry:2` refuses every DELETE with a message that reads like a werf bug; and file reads and searches belong in the harness tools rather than `cat`, `sed` or `grep` in `bash`.
- No generated page changes in this PR: the fix removes an environment dependency, it does not alter correct output.

## Why

The generator reads flag defaults exactly as the commands do, so the developer's shell is an undeclared input to a committed artifact. Clearing offenders one variable at a time — the previous approach — only holds until someone exports a variable nobody thought of, and the failure is silent: the diff looks like normal regeneration, so it survives review and lands as documentation stating a default werf does not have.

The `AGENTS.md` additions are the session's other dead ends, and each cost real work: two commands shipped unrunnable behind a fully green unit suite, a whitespace gate that could never pass, and a registry rejecting deletes in a way that looked like a product bug.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -49,6 +49,7 @@ Follow [Effective Go](https://go.dev/doc/effective_go) and [Go Code Review Comme
 
 - ALWAYS use LSP (`goToDefinition`, `findReferences`, `documentSymbol`, `hover`, `goToImplementation`, call hierarchy) to find definitions, usages, implementations, and callers. `grep` matches strings blindly: it hits comments and unrelated identifiers, and misses interface dispatch and aliased imports.
 - Use `grep` ONLY for literal text — config keys, error message strings, annotation names.
+- ALWAYS use your harness's file-read and search tools instead of `cat`, `sed -n`, `head` or `grep` inside `bash`. They cap what enters the context — byte and match limits, offset windows, long-line truncation — where a shell command dumps the whole file or every hit. Reading an entire file to look at one function is the most expensive habit there is.
 - If your harness has a semantic code-search tool, prefer it over `grep` for intent-based questions ("how does X work"). If it does not, read the code: NEVER substitute keyword grepping for understanding.
 
 ## Commands (MANDATORY)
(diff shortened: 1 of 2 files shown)


alexey-igrychev:
## Verification

- Reproduced on a clean tree: `WERF_DISABLE_AUTO_HOST_CLEANUP=1 task doc:gen` → 12 changed pages, each gaining `--disable-auto-host-cleanup=true`.
- After the fix, the same command → 0 changed pages; a clean-environment run → 0 changed pages, so idempotence is preserved.
- Not run: no automated check guards this. A test would have to regenerate the docs twice under different environments and diff them, which costs a full `task build` plus two generator runs; the shell fix is a one-liner and the failure mode is now impossible rather than detected.

## Review focus

- `scripts/docs/regen.sh` — the `env -u` list is built from `env | grep '^WERF_'`, so `WERF_PLATFORM=''` must stay set after the unsets to keep its intentional empty value. That ordering is the only subtle part.
- Whether any `WERF_*` variable is deliberately meant to influence generated docs. I found none, but this fix assumes so.

## Follow-up

- [ ] Consider whether `task doc:gen` should fail loudly when `WERF_*` is exported instead of silently clearing it — clearing is right for correctness, but a developer who thought a variable mattered gets no signal.

PR #7755, opened 2026-08-04
docs(dev): document three agent rules for PR review and doc:gen

alexey-igrychev:
## Summary

Three rules an agent needs when reviewing a pull request whose head is not checked out, and when regenerating the CLI reference. Each one was written after an agent broke it during the review of #7737 and produced a wrong artifact — a comment anchored on the wrong line, a false finding, a documentation diff that changed twelve pages nobody touched.

## What

- A review of a PR head that is not checked out now reads every `file:line` from that commit's blob, so a posted line comment lands on the line the finding is about. Previously the numbers came from the worktree, which sits on the base: `SetupMetaRepo` was quoted at `common.go:480` while the PR head has it at `:572`.
- The review skill now states that LSP indexes the worktree and therefore answers about the base when the head is not checked out. A review in that situation either adds a worktree at the head or records in *Not verified* that call tracing was limited, instead of reporting untraced callers as if they had been traced.
- `AGENTS.md`'s code-navigation section, which mandates LSP and forbids grep for definitions, now names the one case where LSP cannot work at all. It also warns that a plain worktree grep for the reviewed commit's identifiers returns nothing and resembles a finding — the failure that produced a false "these constants are unused" claim.
- `task doc:gen` is now documented as rendering flag defaults from the current environment, with the instruction to unset `WERF_*` first and to check the diff for untouched flags. An exported `WERF_DISABLE_AUTO_HOST_CLEANUP` rewrote the documented default of `--disable-auto-host-cleanup` in twelve command pages, a diff that looked like ordinary drift.
- Deliberately unchanged: the review skill's *Output* section keeps its single rule — print the report, do not write it into the repository. A harness that wants a review file adapts to that, it does not override it.
- Deliberately unchanged: no rule tells an agent to skip LSP or prefer grep. The navigation mandate stands; only the case where the tool is structurally blind is now named.

## Why

Every one of these was reachable because the existing documents are written for the ordinary case — a checked-out branch and a clean shell — and say nothing about the two situations an agent hits constantly: reviewing a commit it has not checked out, and running a generator whose output depends on the environment it inherits.

That gap is expensive in a specific way. Each failure produces output that looks correct: a line number is a line number, an empty grep looks like an absent symbol, and a `doc:gen` diff looks like drift the agent is supposed to commit. None of them announce themselves, so they survive review unless a rule names them in advance.

The alternative was to leave these in a session retro, which is where they started. A retro is not read before the next review; `AGENTS.md` and the skills are, and they are the documents the harness already loads.

### DIFF
--- a/.agents/skills/review/SKILL.md
+++ b/.agents/skills/review/SKILL.md
@@ -15,9 +15,10 @@ If you wrote the diff you are reviewing now, say so explicitly in the report's V
 
 1. Ask the user for numbered acceptance criteria (DoD). If there are none, derive them from the PR description or the linked issue, mark them `(inferred)`, and proceed — do not stall, and do not invent criteria silently.
 2. Resolve the base first: `git fetch`, then diff against the branch the PR actually targets. werf maintains release branches (`1.2`, `2.63`, `3`, …), so `main` is the wrong base for a backport. State the resolved base in the report. For uncommitted work, review `git diff` / `git diff --cached` instead.
-3. Read every changed file. Then trace callers of the changed exported symbols whose signature or behavior changed, and of anything crossing a persistence boundary — via LSP call hierarchy and references, not grep.
-4. For 10+ changed files, split the reading by area (e.g. new files, storage/cleanup, build pipeline) across subagents if your harness has them, and synthesize the findings yourself.
-5. If the worktree holds the branch and `task` works, run `task build` and `task test:unit` — a review that never compiled the change is an opinion. NEVER run `task format`: it would rewrite the diff under review.
+3. If the PR head is not checked out, take every `file:line` from that commit's blob (`git show <head>:<path>`), never from the worktree. The worktree sits on the base, so any file the PR itself changed has different line numbers there, and a comment anchored on a worktree line number lands on the wrong line — or is rejected outright.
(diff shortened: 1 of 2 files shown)


alexey-igrychev:
## Verification

- The `doc:gen` rule is the one that was reproduced rather than recalled: running it with `WERF_DISABLE_AUTO_HOST_CLEANUP=1` exported changed twelve pages under `docs/_includes/reference/cli/`, all of them flipping `--disable-auto-host-cleanup=false` to `true`; re-running with the variable unset produced an empty diff. That is the before/after the new sentence describes.
- The worktree-grep failure was hit live during #7737: grepping the checked-out tree for that PR's own `ItemType*` constants returned zero hits for all ten, which reads as "these constants are unused" and is purely an artifact of the tree sitting on the base. Re-running against the blob returned the real counts.
- The line-number failure is checkable from two blobs: `git show 78a2fafb2:cmd/werf/common/common.go` has `SetupMetaRepo` at `:480`, and at PR #7737's head it sits at `:572`.
- Not run: nothing mechanical can check that these rules change agent behaviour. The only evidence available is that each rule names a failure that already happened and states the action that would have prevented it; whether an agent then follows them is not verifiable here.
- No tests: the diff is three prose rules in two Markdown documents, so there is no implementation to mutate.

## Review focus

- `.claude/skills` is a symlink to `.agents/skills`, so the review-skill edit reaches both harnesses through one file; there is no second copy to update.

PR #7762, opened 2026-08-05
test(build): convert the include.path tests to Ginkgo

alexey-igrychev:
## Summary

The three tests covering `include.path` forwarding in `pkg/true_git` were written with testify, while AGENTS.md mandates Ginkgo and Gomega and every other spec in the package already uses them. This converts them and drops the testify-only git helpers they needed. Follow-up to #7736, which converted the specs it added and left these behind.

## What

- No production code changes: the invariant is that the same three behaviors stay pinned — `getIncludePathOptions` resolving absolute, repository-relative and `~` include paths; `updateSubmodules` forwarding those includes to its git invocations; and `switchWorkTree` on a repository without submodules being unaffected by an ambient include. VERIFIED: mutation-checked, both falsifiable behaviors caught.
- `pkg/true_git` no longer uses testify: `helpers_ai_test.go` is deleted and the converted specs use the helpers already in `helpers_test.go`.
- The suite reports 102 specs where it reported 99 plus three `go test` functions; nothing is skipped and no assertion is dropped.

## Why

Two test styles in one package mean two sets of git helpers doing the same thing and two failure formats to read. The testify tests also ran invisibly under `task test:unit`, which reports Ginkgo specs only, so their result was only ever observable through a failure.


alexey-igrychev:
## Verification

- Mutation: `getIncludePathOptions` returning `nil, nil` immediately → both `resolves every include.path of the repository config` and `forwards the include to the submodule update` failed.
- Mutation: dropping `includePathOpts` from the argument list of `git submodule update` → `forwards the include to the submodule update` failed.

## Review focus

- `leaves a work tree switch without submodules unaffected` is a smoke guard, not a falsifiable claim: the included rewrite only matches `https://` URLs, which a work tree switch never uses, so no plausible mutation of `switchWorkTree` makes it fail. It is carried over as-is from the testify version rather than strengthened, since strengthening it is not this PR's subject.

PR #7763, opened 2026-08-05
fix(build, buildah): stop imports failing on symlinked paths like /bin

Fral738:
## Summary

On the Buildah backend, an import whose path traverses a symlink with an absolute target inside the source image (e.g. `/bin -> /usr/bin` on usrmerge distros) fails at the import checksum step with `error accessing <storage>/overlay/<id>/merged/bin/<file>: lstat ...: no such file or directory`. werf resolved such symlinks against the host root instead of the mounted image rootfs; now they are resolved strictly within it. It needs no special host state — only a source image where the configured import path goes through an absolute symlink.

### Imports

- An `import` whose `add` path traverses an absolute symlink in the source image no longer fails checksum calculation with "no such file or directory"; parent components are resolved inside the image rootfs.
  - VERIFIED: on a Linux host with rootless Buildah — a Stapel `import` with `add: /opt/link/tool` where `/opt/link -> /opt/real` reproduced the exact reported error on a main-built binary and built successfully on this branch, with the file present in the image and a rebuild fully reusing cache.
- An `add` path whose final component is itself a symlink keeps its previous meaning: the symlink is hashed and imported as a symlink, never its target tree, so existing checksums do not change and no stage rebuilds are triggered — for any previously working configuration, symlink-final or not. The checksum format is pinned by unit tests computing expected values independently.
- `owner`/`group` name lookup for imports reads `etc/passwd`/`etc/group` of the image rootfs even when `/etc` in the image is a symlink; previously it silently read the host files.

### Writes and cleanup inside the image rootfs

- Import copying, git/data archive extraction, and remove-inside-path cleanup resolve destination paths fully inside the container rootfs; previously a destination behind an absolute symlink was written to the host filesystem path the symlink named.
  - A destination whose final component is a symlinked directory now receives files in its target inside the rootfs — for relative symlinks that is what already happened, for absolute ones it previously hit the host.
- Exact-path removal of a path that is itself a symlink removes the link, not its target; a dangling symlink is now removed too, where it was previously skipped.
- Git-mapping cleanup keeps the mapped destination directory after removing its last file: the keep-parent-dirs check compared container paths against rootMount-prefixed on-disk parents and never matched — broken on main as well, not introduced by this branch.
- Empty-parent pruning after git cleanup can never climb past the container root mount into the host directory tree.
- A symlink pointing outside the image rootfs (`../../../etc`) is clamped to the rootfs root and can never reach the host.

### Unchanged

- The Docker backend and the legacy Stapel builder are unchanged: they calculate import checksums inside a container, where symlinks always resolve against the container root.
- `github.com/cyphar/filepath-securejoin` moves from indirect to direct dependency; version pinned before and after at v0.2.5.

## Why

In-container paths were joined onto the mounted rootfs with a plain `filepath.Join`, so the kernel resolved any absolute symlink component against the host root: reads miss (`/usr/bin/<file>` of the host does not exist) and writes land on the host — a correctness bug on every usrmerge image and a host-filesystem escape in rootful mode. Running the checksum inside a container, as the legacy Stapel builder does, would fix resolution too, but costs a container run per import; `securejoin.SecureJoin` is the containers-stack standard for host-side rootfs path handling and is already in the dependency graph. Resolving every component including the final one was rejected for sources and exact removals: it silently changes checksums and imported content for `add` paths that are themselves symlinks, so only parent components are resolved there.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -189,7 +189,7 @@ require (
 	github.com/coreos/go-systemd/v22 v22.5.0 // indirect
 	github.com/cpuguy83/go-md2man/v2 v2.0.4 // indirect
 	github.com/cyberphone/json-canonicalization v0.0.0-20231217050601-ba74d44ecf5f // indirect
-	github.com/cyphar/filepath-securejoin v0.2.5 // indirect
+	github.com/cyphar/filepath-securejoin v0.2.5
 	github.com/davecgh/go-spew v1.1.2-0.20180830191138-d8f796af33cc
 	github.com/daviddengcn/go-colortext v1.0.0 // indirect
 	github.com/disiqueira/gotree/v3 v3.0.2 // indirect
--- a/pkg/container_backend/buildah_backend.go
+++ b/pkg/container_backend/buildah_backend.go
@@ -392,6 +392,14 @@ func (backend *BuildahBackend) CalculateDependencyImportChecksum(ctx context.Con
 		return "", err
 	}
 
+	return calculateDependencyImportChecksum(ctx, fromPath, dependencyImport)
+}
+
+// calculateDependencyImportChecksum hashes the files under fromPath (the already resolved
+// on-disk location of dependencyImport.FromPath), keying every entry by the configured
+// FromPath so the checksum does not depend on where the container root is mounted or on
+// symlink resolution of the parent directories.
+func calculateDependencyImportChecksum(ctx context.Context, fromPath string, dependencyImport DependencyImportSpec) (string, error) {
 	pathMatcher := path_matcher.NewPathMatcher(path_matcher.PathMatcherOptions{
 		BasePath:     fromPath,
 		IncludeGlobs: dependencyImport.IncludePaths,
@@ -400,7 +408,7 @@ func (backend *BuildahBackend) CalculateDependencyImportChecksum(ctx context.Con
 
 	var files []string
 
-	err = filepath.Walk(fromPath, func(path string, f os.FileInfo, err error) error {
+	err := filepath.Walk(fromPath, func(path string, f os.FileInfo, err error) error {
 		if err != nil {
 			return fmt.Errorf("error accessing %s: %w", path, err)
 		}
--- a/pkg/container_backend/buildah_backend.go
+++ b/pkg/container_backend/buildah_backend.go
@@ -18,6 +18,7 @@ import (
 	"sync"
 
 	"github.com/containers/storage/types"
+	securejoin "github.com/cyphar/filepath-securejoin"
 	"github.com/google/uuid"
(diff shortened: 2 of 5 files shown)


Fral738:
## Verification

- End-to-end on Linux (Ubuntu 22.04, rootless Buildah, `WERF_BUILDAH_MODE=auto`): a two-image Stapel project where `src` creates `/opt/real/tool` plus absolute symlink `/opt/link -> /opt/real`, and `app` imports `add: /opt/link/tool`. A binary built from `origin/main` fails with the exact reported error (`unable to calculate dependency import checksum ...: error accessing .../merged/opt/link/tool: lstat ...: no such file or directory`); the branch binary builds successfully, the imported file is present in the mounted image with correct content, and an immediate rebuild reuses all stages from cache (checksum stable across runs).
- Mutation: reverted `resolveContainerRootPath` to a plain `filepath.Join` → 4 specs failed (absolute/relative/escaping symlink resolution and the `applyDataArchives` absolute-symlink destination spec); restored, suite green.
- Mutation: degraded `resolveContainerRootPathNoFollow` to full resolution → the specs pinning the unresolved final component failed; restored, suite green.
- Mutation: reverted the checksum hash-entry keying to the on-disk path → all 5 `calculateDependencyImportChecksum` regression specs failed; restored, suite green. The expected values are computed independently in the tests with `crypto/md5`, not by calling the production code.
- Mutation: passed `KeepParentDirs` unresolved (prior behavior) → both `applyRemoveData` specs failed (symlinked keep dir pruned, pruning climbed past the root mount); restored, suite green.
- Two independent adversarial review passes (fresh agents, no author context) were run; the first surfaced the final-component symlink semantics (fixed in the second commit), the second surfaced the dead `KeepParentDirs` check (fixed in the fourth commit).

## Review focus

- Sources vs destinations resolve differently by design (`resolveContainerRootPathNoFollow` vs `resolveContainerRootPath`): checksum `fromPath` and `applyDependenciesImports` `absFrom` must stay in lockstep with each other, and destinations must stay fully resolved — confirm no call site mixes them up.
- The extraction of `calculateDependencyImportChecksum` is intended to be behavior-preserving; the regression specs pin the format (FromPath-keyed entries, mount-location independence, symlink-final hashing, include/exclude filtering).
- The `KeepParentDirs` fix changes cleanup behavior relative to main (main never kept anything on Buildah): confirm keeping the git-mapping `to` directory after its last file is removed is the intended contract, matching `git_mapping.go` passing `gm.To` as the keep dir.

reyreavman (changes requested):
lets resolve conflicts and add a regression test for the import checksum

reyreavman (approved):

PR #7764, opened 2026-08-05
fix(build): reuse content anchors without git commits

alexey-igrychev:
## Summary

Cached content anchors from Git-backed images are now reused even when the source commit recorded by an existing stage is unavailable. Ordinary Git stages retain commit validation because they need history to generate patches.

## What

- A cached content anchor with a matching digest is selected without requiring its recorded Git commit to exist.
- Non-anchor Git stages continue to reject cached stages whose recorded commits are unavailable.

## Why

The terminal Git stage serving as a content anchor inherited ordinary stage selection, which validates the cached stage's recorded commit. A content anchor is content-addressed, so that validation wrongly prevents reuse when the commit is unavailable despite an intact remote stage. Selecting it through the base-stage policy removes the unnecessary Git dependency without weakening patch-stage safety.

### DIFF
--- a/pkg/build/stage/git.go
+++ b/pkg/build/stage/git.go
@@ -33,6 +33,10 @@ func (s *GitStage) PrepareImage(ctx context.Context, c Conveyor, cb container_ba
 	return s.BaseStage.PrepareImage(ctx, c, cb, prevBuiltImage, stageImage, nil)
 }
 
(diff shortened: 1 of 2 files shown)


alexey-igrychev:
## Verification

- Mutation: restore the commit-dependent Git selector → the regression test receives `nil` instead of the cached content-anchor descriptor and fails.

PR #7765, opened 2026-08-05
fix(build): make repo-built from:scratch images readable by dive

Fral738:
## Summary

Images built with `from: scratch` and a stages repo (`--repo`) could not be read by `dive` and other `docker save` consumers on Linux Docker engines: they failed with "cannot fetch image: EOF" on the `from` stage and every stage inheriting it. Reproduces from a clean checkout when no local stage cache exists:

```yaml
# werf.yaml
project: scratch
configVersion: 1
---
image: source
from: alpine:latest
shell:
  install:
    - apk add wget
---
image: destination
from: scratch
import:
  - image: source
    add: /usr/bin/wget
    to: /usr/bin/wget
    before: install
```

```shell
werf build --repo localhost:5000/scratch
docker pull localhost:5000/scratch:<final-stage-tag>
dive localhost:5000/scratch:<final-stage-tag>   # cannot fetch image: EOF
```

## What

- The empty base layer of a `from: scratch` stage created directly in the stages repo is now a valid empty tar archive (diffID `sha256:5f70bf18…`, same as the local build path) instead of a zero-byte blob (`sha256:e3b0c442…`).
- `dive` and `docker save` readers no longer fail with "cannot fetch image: EOF" on repo-built scratch stages or their descendants. VERIFIED: hand-run on Ubuntu 22.04 (Docker 29.2.1, overlay2, dive 0.13.1) — stock v2.74.0 fails on all scratch-chain stages, the patched binary passes on all of them.
- The fix covers both backends in repo mode: the broken layer came from the backend-agnostic manifest push, and with it reverted the e2e scratch suite fails on the vanilla-docker and the native-chroot repo entries alike.
- Stage digests do not change: existing build caches stay valid, no rebuilds are triggered.
- Already-published scratch stages keep the broken layer — werf reuses them by digest; recreating the `from` stage (e.g. bumping `fromCacheVersion`) is required to heal an existing project.
- The e2e readability helper (`ExpectImageIsReadable`) now rejects zero-byte layers; previously go-containerregistry read them back as gracefully empty, which is why CI never caught this.
- The local Docker build path (fixed in #7609) and local Buildah behavior do not change.

## Why

`RepoStagesStorage.PostManifest` creates the scratch `from` stage as a manifest-only image whose rootfs was `EmptyUncompressedLayer`: nil content with a hardcoded diffID of the empty string. Registries and `docker pull` accept a zero-byte layer blob, but it is not a parseable tar archive, so any consumer that streams the image back (dive, `docker save` readers on graphdriver engines) hits EOF — and every descendant stage inherits the layer, poisoning the whole chain. Docker Desktop on macOS (containerd image store) tolerates it, which masked the bug outside Linux. #7609 fixed the same symptom for the local `ImageImport` path only; this aligns the repo path with it.

Fixes: 490d6e0df8d0 ("feat(build): scratch stapel docker backend")

### DIFF
--- a/pkg/docker_registry/container_registry_extensions/manifest_only_image_test.go
+++ b/pkg/docker_registry/container_registry_extensions/manifest_only_image_test.go
@@ -1,0 +1,47 @@
+package container_registry_extensions
+
+import (
+	"archive/tar"
+	"bytes"
+	"io"
+
+	v1 "github.com/google/go-containerregistry/pkg/v1"
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+)
+
+// The manifest-only image is pushed to registries as the scratch from stage base, and every
+// descendant stage inherits its layer. A zero-byte layer body is not a parseable tar and breaks
+// consumers like dive with EOF, so the layer must always be a valid, non-empty tar archive.
+var _ = Describe("manifest-only image", func() {
+	It("has a single valid empty tar layer with a matching diffID", func() {
+		img := NewManifestOnlyImage(map[string]string{"werf": "test"})
+
+		layers, err := img.Layers()
+		Expect(err).NotTo(HaveOccurred())
+		Expect(layers).To(HaveLen(1))
+
+		rc, err := layers[0].Uncompressed()
+		Expect(err).NotTo(HaveOccurred())
+		data, err := io.ReadAll(rc)
+		Expect(err).NotTo(HaveOccurred())
+		Expect(rc.Close()).To(Succeed())
+
+		Expect(len(data)).To(BeNumerically(">", 0), "a valid tar archive is never zero bytes")
+
+		tr := tar.NewReader(bytes.NewReader(data))
+		_, err = tr.Next()
+		Expect(err).To(Equal(io.EOF), "expected a well-formed empty tar with no entries")
+
+		expectedDiffID, _, err := v1.SHA256(bytes.NewReader(data))
+		Expect(err).NotTo(HaveOccurred())
+		diffID, err := layers[0].DiffID()
+		Expect(err).NotTo(HaveOccurred())
+		Expect(diffID).To(Equal(expectedDiffID))
+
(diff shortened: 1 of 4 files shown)


Fral738:
## Verification

- Hand-run repro on Ubuntu 22.04 (Docker 29.2.1, overlay2 graphdriver, dive 0.13.1): clean `werf build --repo` against a local registry with stock v2.74.0 → `dive` fails with "cannot fetch image: EOF" on the scratch `from` stage and its descendant; the same build with the patched linux/amd64 binary → all four stage images readable, empty-layer diffID `sha256:5f70bf18…`, stage tags identical to the broken run (cache parity confirmed).
- `docker save` exits 0 even for the broken images — only streaming consumers like dive fail, which is why the symptom looks tool-specific.
- e2e scratch suite (`task test:e2e paths="./test/e2e/build" labelFilter="scratch"`) run on the same Ubuntu host against this branch: all four entries pass, including both `WithLocalRepo` ones.
- Mutation (unit): reverted the layer content to nil (the prior behavior) → the new manifest-only image test failed on the non-empty-tar assertion; fed an empty stream into the diffID computation → the diffID-match assertion failed. Suite green after restore.
- Mutation (e2e): reverted `uncompressed_layer.go` to its pre-fix version with the tests kept → the same e2e scratch suite failed on the strengthened check, "expected layer 0 to be a non-empty tar archive", on both the vanilla-docker and the native-chroot repo entries; restored to the branch commit afterwards. Project names are unique per run, so a stage-cache false pass is excluded.

## Review focus

- Every `NewManifestOnlyImage` consumer gets the new layer: import metadata and cleanup-record images change their digests on next push. They are looked up by tag, so this should be invisible — worth a second pair of eyes.

## Follow-up

- [ ] Decide how existing projects heal already-published broken scratch stages (they are reused by digest); document `fromCacheVersion` bump or stage recreation if user reports keep coming.

reyreavman (approved):

PR #7767, opened 2026-08-05

alexey-igrychev:
## Summary

Merges `origin/main` into `3`, bringing 6 commits that had diverged. One is user-visible: concurrent builds pulling the same base image no longer race under the Buildah backend. The rest are agent-tooling docs (`AGENTS.md`, `.agents/skills`) with no runtime effect.

## What

- Under the Buildah backend, concurrent stage builds that pull the same base image are serialized, so parallel builds no longer race on the shared pull (#7664).
- No other user-visible werf behavior changes; remaining commits edit `AGENTS.md`, `.agents/skills/pull-request`, and CLI-docs generation env handling (#7754, #7749, #7748, #7747).
- Conflict resolutions are docs-only: `AGENTS.md` keeps branch-3's macOS Docker-backend e2e guidance and adds main's `git diff --check`/`task test:unit` caveats and registry-deletion note; `pull-request/SKILL.md` takes main's rewording but keeps branch-3's dependency-bump lists without `werf/3p-helm` (not a dependency on `3`).

## Why

`3` was 6 commits behind `main`; merging keeps the release branches from diverging further. All conflicts were in agent-facing docs edited independently on both branches — resolved by unioning the two sets of guidance, no code conflict.


alexey-igrychev:
## Verification

- `task build` passes on macOS (non-CGO); the only manually-resolved files are docs, and the Buildah change (#7664) came intact from `main`.

## Review focus

- `AGENTS.md` conflict resolution: verify no guidance from either branch was dropped when unioning the macOS/test-caveat paragraphs.

## Follow-up

- [ ] Merge non-draft only after the buildah base-image-pull fix (#7664) is confirmed already validated on `main`.

PR #7768, opened 2026-08-05
chore(docs): add v3 docs deploy routing

alexey-igrychev:
## Summary

Wire up deployment and routing for the v3 documentation so pushes to the `3` branch publish a separate `werfio-v3-production` release and `/docs/v3` becomes an addressable docs root. The default docs root stays `v2` until v3 lands on `main`.

## What

- Push to branch `3` (also `repository_dispatch`/`workflow_dispatch` for it) now deploys production docs as the `werfio-v3-production` release.
- The production deploy maps refs `3`/`v3`/`v3.*` to docs root `v3`; previously such a ref exited with `Unsupported docs production ref`.
- `docs:deploy:test-versions` gains a `v3` target (`werfio-v3-test`, `global.github_ref=v3`).
- `/docs/v3` and `/docs/v3.*` are addressable docs roots; `supportedRoots` is now `[v3, v2, v1.2]`. Bare `/docs/` still redirects to `/docs/v2/`.
- v3 pages declare canonical URL prefix `/docs/v3`.
- Does NOT change the default docs root or the v2/v1.2 deploys.

## Why

werf 3 is released on the `3` branch but had no docs deploy path: the production workflow's ref mapping and the deploy-latest trigger knew only `main`/`2`/`1.2`, and the routing values listed only `v2`/`v1.2`, so v3 docs were neither built nor reachable. The production Helm templates are already generic over `global.github_ref`, so wiring is limited to the ref mapping, trigger, supported-roots list and canonical prefix. Keeping `currentRoot` at `v2` avoids flipping the public default before v3 is promoted on `main`.

### DIFF
--- a/.github/workflows/_docs_deploy_production.yml
+++ b/.github/workflows/_docs_deploy_production.yml
@@ -47,6 +47,10 @@ jobs:
           fi
 
           case "$github_ref_value" in
+            3|v3|v3.*)
+              docs_ref_value="v3"
+              release_name="werfio-v3-production"
+              ;;
             2|v2|v2.*)
               docs_ref_value="v2"
               release_name="werfio-v2-production"
--- a/.github/workflows/docs_deploy_latest.yml
+++ b/.github/workflows/docs_deploy_latest.yml
@@ -4,6 +4,7 @@ on:
   push:
     branches:
       - main
+      - "3"
       - "2"
       - "1.2"
     paths:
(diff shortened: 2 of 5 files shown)


alexey-igrychev:
## Review focus

- Paired with werf/website#850 (registers v3 in the router and version switcher); both must land for `/docs/v3` to resolve.

## Follow-up

- [ ] Flip the default docs root to v3 (`currentRoot` here, `CURRENT_DOCS_MAJOR`/`currentRoot` in werf/website) once v3 lands on `main`.

PR #7769, opened 2026-08-05
test(ci): enable race detector in unit tests

alexey-igrychev:
## Summary

Regular unit-test CI now runs with Go's race detector. The coverage path remains unchanged.

## What

- Non-coverage invocations of the reusable unit-test workflow pass `--race` to Ginkgo.
- Coverage invocations keep their existing command and coverage artifact behavior.

## Why

Regular unit tests previously did not detect unsafe concurrent memory access. Keep coverage separate so this change only adds race detection to the existing non-coverage CI path.


alexey-igrychev:
## Verification

- `git diff --check -- .github/workflows/_test_unit.yml`.
- `task --yes -p test:unit paths="./pkg ./cmd" -- --race --keep-going --skip-package ''` on macOS executed 41 suites; it reported no race detector findings before failing in existing fixtures: GPG could not allocate memory in `pkg/true_git`, and `ps: cmd: keyword not found` in `pkg/werf/exec`.
- Mutation: not run — this workflow change has no assertion; removing `--race` would silently weaken the CI check.
- Not run: a green full race pass against `main`, blocked by those local fixture/environment failures.

PR #7773, opened 2026-08-06
chore(release): sign newest version tag

alexey-igrychev:
## Summary

`task sign` now signs most recently created `v*` version tag instead of tag with highest semantic version. This keeps automatic signing aligned with release creation order.

## What

- Without `refs`, `task sign` selects one `v*` tag by newest Git `creatordate`.
- Automatic selection excludes non-version tags, including git-signatures bookkeeping tags.
- Explicit `refs` and signing `origin/main` remain unchanged.
- UNVERIFIED: remote signature update was not run because it pulls, adds, and pushes signature refs.

## Why

Semantic-version ordering can select an older release when a new tag is cut on a maintained release line after a higher version already exists. Selecting by tag creation date signs tag created by current release workflow.

### DIFF
--- a/Taskfile.dist.yaml
+++ b/Taskfile.dist.yaml
@@ -540,12 +540,12 @@ tasks:
       - go install github.com/dmarkham/enumer@v1.5.11
 
   sign:
-    desc: 'Sign last version tag + origin/main and push signatures. Important vars: "refs".'
+    desc: 'Sign last created version tag + origin/main and push signatures. Important vars: "refs".'
     cmds:
       - git fetch --tags -f
       - git signatures pull {{.CLI_ARGS}}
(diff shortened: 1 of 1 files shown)


alexey-igrychev:
## Verification

- `git for-each-ref --sort=-creatordate --count=1 --format='%(refname:short)' 'refs/tags/v*'` selected `v2.75.4` while newer non-version signature tags exist.
- `TASK_X_REMOTE_TASKFILES=1 task --yes --taskfile Taskfile.dist.yaml --dry sign` parsed target successfully.
- `git diff --check -- Taskfile.dist.yaml` passed.
- Not run: `task sign`; it updates remote signature refs.

PR #7774, opened 2026-08-06
fix(build, buildah): prevent parallel recovery failures

alexey-igrychev:
## Summary

Buildah staged builds can recover the same stale cached image ID concurrently after local image cleanup. Recovery pulls for one image ref now serialize, so shared containers storage does not return `image not known` again.

## What

- When multiple target-platform build stages find the same cached image ID missing locally, only one recovery pull for that image ref enters Buildah at a time.
- A successful recovery stores its image ID in the cache before another recovery pull for that ref starts.
- UNVERIFIED: live Linux Buildah recovery with local-image removal and concurrent consumers has not run; a Linux e2e that removes the local image between stages would settle it.
- No flags, environment variables, output, or retry count change.

## Why

The original stale-ID fallback pulled through the low-level Buildah client, bypassing the existing per-ref lock. Concurrent recovery could therefore recreate the documented containers-storage race and fail a build with `image not known`. Reuse the existing ref-only mutex because storage names are ref-based across platforms.

Fixes: 9d640357576a ("fix(build, buildah): retry pull when cached image id is missing (#7669)")

### DIFF
--- a/pkg/container_backend/buildah_backend.go
+++ b/pkg/container_backend/buildah_backend.go
@@ -182,14 +182,24 @@ func (backend *BuildahBackend) createContainers(ctx context.Context, images []st
 		if err != nil && opts.TargetPlatform != "" && usedCachedImageID && isImageNotKnownError(err) {
 			logboek.Context(ctx).Debug().LogF("Cached imageID %q for %q not found locally, pulling by ref and retrying\n", resolvedImg, img)
 
-			pulledImageID, pullErr := backend.buildah.Pull(ctx, img, buildah.PullOpts(backend.getBuildahCommonOpts(ctx, true, nil, opts.TargetPlatform)))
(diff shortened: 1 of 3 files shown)


alexey-igrychev:
## Verification

- Scoped local: `task test:unit paths="./pkg/container_backend"`.
- Mutation: remove recovery-pull mutex → concurrent regression fails because second pull reaches fake Buildah while first remains blocked.
- `task build` and `task lint` pass locally.
- Local `task test:unit` ran but `pkg/true_git` failed while signing a temporary Git commit: `gpg: signing failed: Cannot allocate memory`.
- Not run: Linux Buildah e2e that removes a cached local image between concurrent stage consumers.

## Review focus

- Confirm recovery pull remains serialized by ref across platforms; containers-storage uses ref-derived shared names.

Issue #7775, opened 2026-08-06
Track races reported by e2e race detector


alexey-igrychev:
## Version

Race coverage is proposed by [#7770](https://github.com/werf/werf/pull/7770), targeting branch `3`. The current Nelm race fix is proposed separately in [#7796](https://github.com/werf/werf/pull/7796).

## How to reproduce

Run non-coverage e2e with a race-instrumented werf binary. GitHub Actions [run 31105226004](https://github.com/werf/werf/actions/runs/31105226004) produced the original full-suite reports.

Focused remote Linux compose reproduction:

```shell
WERF_TEST_BINARY_PATH=/path/to/race/werf \
  task test:e2e paths=&#39;./test/e2e/compose/...&#39; \
  labelFilter=&#39;e2e &amp;&amp; compose &amp;&amp; complex&#39; parallel=1 \
  -- --focus=&#39;build report&#39;
```

### werf

- [x] `pkg/docker/main.go`: custom Docker CLI streams mutated a shared CLI while Buildx read it. Fixed by [#7777](https://github.com/werf/werf/pull/7777).
- [x] `pkg/buildah/native_linux.go`: `RunCommand` read its stderr `bytes.Buffer` while Buildah&#39;s subprocess writer still appended to it. Fixed by [#7788](https://github.com/werf/werf/pull/7788).
- [x] `pkg/buildah/native_linux.go`: concurrent Dockerfile operations mutated shared `defaultCommonBuildOptions` via a shared pointer (`Secrets`/`SSHSources` written per-op). Fixed by a per-operation copy; targeted remote race E2E passes. Landed in `3` by [#7798](https://github.com/werf/werf/pull/7798).
- [x] `pkg/storage/manager/storage_manager.go`: parallel final-stage copies accessed `StagesList.AddStageID` concurrently. Fixed by [#7787](https://github.com/werf/werf/pull/7787).
- [x] `test/pkg/utils/command.go`: sequential stdout/stderr draining could deadlock a verbose child process. Fixed by [#7792](https://github.com/werf/werf/pull/7792).

### Dependencies

[...]

PR #7777, opened 2026-08-06
fix(docker): prevent race reports during Docker builds

alexey-igrychev:
## Summary

Docker commands that need custom streams now use an isolated Docker CLI instead of modifying the context-bound CLI. This prevents Buildx progress from racing with stream reconfiguration.

## What

- Custom Docker CLI operations create a new CLI initialized with the default and requested stream options.
- The context-bound Docker CLI remains unchanged during those operations.
- VERIFIED: a race-instrumented Vanilla Docker e2e build completes without the previous `pkg/docker/main.go:205` race report.

## Why

The previous implementation applied custom streams to the shared Docker CLI and restored them after the command. Buildx can read that CLI concurrently for progress output, producing a race report and terminating werf with exit code 66.

Tracked in #7775.

### DIFF
--- a/pkg/docker/main.go
+++ b/pkg/docker/main.go
@@ -196,21 +196,12 @@ func defaultCliOptions(ctx context.Context) []command.CLIOption {
 }
 
 func cliWithCustomOptions(ctx context.Context, options []command.CLIOption, f func(cli command.Cli) error) error {
-	if err := cli(ctx).Apply(options...); err != nil {
-		return err
+	customCli, err := newDockerCli(append(defaultCliOptions(ctx), options...))
+	if err != nil {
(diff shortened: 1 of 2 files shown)


alexey-igrychev:
## Verification

- `task format`, `task build`, and `task lint` passed.
- `task test:unit paths="./pkg/docker"` passed.
- Linux: race-instrumented Vanilla Docker e2e build passed without a race report.
- `task test:unit` failed in `pkg/true_git`; this failure is outside the changed package.
- Mutation: replaced the isolated CLI with the context-bound CLI; `cliWithCustomOptions uses a separate Docker CLI` failed, then passed after restoring the fix.

## Follow-up

- [ ] Track remaining reports in #7775.

PR #7778, opened 2026-08-06

alexey-igrychev:
## Summary

Synchronizes release branch `3` with `main` at `6d6af8d02be5f1a58e5dc8806c7e64d5c9f150c6`.

## What

- Branch `3` receives the behavior and tests already merged into `main` through that commit.
- Docker commands that temporarily redirect CLI output use a separate CLI instance, so concurrent commands do not mutate one another's output streams.
- Buildah retries a pull when a cached image ID no longer exists locally and serializes concurrent recovery pulls.
- UNVERIFIED: the native Buildah backend is Linux/cgo-only; this macOS verification did not build or run that implementation.

## Why

Branch `3` had diverged from `main`, leaving it without fixes already available on the primary release branch. Its newer Docker and storage dependencies are incompatible with `main`'s old APIs, so the merge keeps the v3-compatible APIs while applying the behavioral fixes.

### DIFF
--- a/.github/workflows/_test_unit.yml
+++ b/.github/workflows/_test_unit.yml
@@ -65,7 +65,7 @@ jobs:
             task --yes -p test:unit paths="$(echo ${{ inputs.packages }} | tr , ' ')" -- --coverprofile="$(openssl rand -hex 6)-coverage.out" --keep-going --skip-package '${{ inputs.excludePackages }}'
             mv *-coverage.out "$WERF_TEST_COVERAGE_DIR/"
           else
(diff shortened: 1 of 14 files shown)


alexey-igrychev:
## Verification

- `TASK_X_REMOTE_TASKFILES=1 task --yes format && task build && TASK_X_REMOTE_TASKFILES=1 task --yes lint && task test:unit`
- Mutation: pass the context CLI instead of the newly created CLI to `cliWithCustomOptions` → `docker system / uses a separate Docker CLI` failed; restored source, then `task test:unit paths="./pkg/docker/..."` passed.

## Review focus

- Conflict resolution keeps the v3 Docker client and Podman storage APIs while retaining the incoming Docker race and Buildah recovery behavior.

PR #7779, opened 2026-08-06
docs(dev): strengthen PR review guidance

alexey-igrychev:
## Summary

Review guidance now handles pull-request heads that are not checked out and catches safety claims in PR descriptions that review findings disprove.

## What

- A review of an un-checked-out PR head reads `file:line` from that head's blobs, preventing inline comments from targeting base-branch lines.
- In that condition, reviewers check out the head before using LSP call tracing or record that tracing was limited; base-worktree LSP results are not treated as evidence about the PR.
- A PR description safety claim that a review finding contradicts is reported as a defect, so the squashed commit body does not preserve a false guarantee.
- Existing dependency and fork guidance remains unchanged for `main`.

## Why

A base worktree has different source lines and LSP symbols from an un-checked-out PR head, which can create wrong comments and false caller conclusions. PR descriptions outlive inline review comments after squash merging, so contradicted safety guarantees need an explicit review rule.


alexey-igrychev:
## Verification

- `git diff --check origin/main...HEAD`

PR #7783, opened 2026-08-06
fix(ci): stop rootless buildah e2e builds failing

alexey-igrychev:
## Summary

Rootless Buildah E2E jobs still fail when a GitHub runner image lacks `slirp4netns`. They use a separate workflow from the integration job fixed in #7782, so this workflow now installs the same required runtime helper.

## What

- Regular E2E CI installs `slirp4netns`, allowing native-rootless Buildah to configure its network namespace.
- UNVERIFIED: The GitHub-hosted E2E jobs must pass with the updated runner dependency.

## Why

The `e2e_extra` and `e2e_complex` jobs failed in native-rootless Buildah cases in the same Actions run as #7782. Their workflow installed Buildah but relied on the runner to supply its `slirp4netns` network helper; the runner no longer did so.

### DIFF
--- a/.github/workflows/_test_e2e_regular.yml
+++ b/.github/workflows/_test_e2e_regular.yml
@@ -43,7 +43,7 @@ jobs:
       - name: Install system dependencies
         run: |
           sudo apt update
-          sudo apt install -y libbtrfs-dev buildah qemu-user qemu-user-binfmt
+          sudo apt install -y libbtrfs-dev buildah qemu-user qemu-user-binfmt slirp4netns
 
       - name: Prepare system for multiarch builds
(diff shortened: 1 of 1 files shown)


alexey-igrychev:
## Verification

- `TASK_X_REMOTE_TASKFILES=1 task --yes format paths=".github"`; the formatter made no changes.
- Parsed `.github/workflows/_test_e2e_regular.yml` with Ruby YAML and ran `git diff --check` for the authored file.
- Not run: rootless Buildah E2E requires a GitHub-hosted Linux runner; PR CI will execute it.

alexey-igrychev:
Superseded by #7784 to correct the change type from `fix(ci)` to `chore(ci)`.

PR #7785, opened 2026-08-06
fix(build): avoid panicking on late worker logs

alexey-igrychev:
## Summary

A parallel build no longer panics and hides its original error when a subprocess flushes log output after its task completes. Late output is ignored once the worker has finished.

## What

- A parallel worker accepts and discards writes after completion instead of returning an error that makes the logger panic.
- Process output emitted after its owning task completes is not included in the build log.

## Why

Task completion and subprocess output completion can race. Previously, a late write reached the half-closed worker, and the logger panicked while reporting the original build error. Waiting for every third-party stream would require lifecycle ownership the worker does not have; ignoring only post-completion output preserves the original error.

### DIFF
--- a/pkg/util/parallel/parallel_test.go
+++ b/pkg/util/parallel/parallel_test.go
@@ -294,6 +294,31 @@ var _ = DescribeTable("parallel task",
 	),
 )
 
+var _ = It("does not panic when a task writes after completion", func() {
+	output := newSpyOutput(1)
+	ctx := logboek.NewContext(context.Background(), logboek.NewLogger(output, output))
+	write := make(chan struct{})
+	written := make(chan struct{})
+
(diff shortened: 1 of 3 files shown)


alexey-igrychev:
## Verification

- `TASK_X_REMOTE_TASKFILES=1 task --yes format paths="pkg/util/parallel"`, `task --yes build`, and `task --yes lint` passed.
- `task --yes test:unit paths="./pkg/util/parallel/..."` passed.
- Mutation: made post-half-close writes reach the file; the worker test observed `hello`, the closed-worker path failed, and the late-output test panicked in `logboek`.
- Independent challenge review: approved after the mutex and output assertions were added.
- Not run: full `task test:unit` fails on the base branch too in `pkg/werf/exec`: macOS `ps` rejects the test command (`ps: cmd: keyword not found`).

PR #7786, opened 2026-08-06
test(dev): use portable ps command field

alexey-igrychev:
## Summary

The detach test now uses a `ps` field available on both GNU and BSD implementations. It no longer fails on macOS with `ps: cmd: keyword not found`.

## What

- The detach test queries `ps -eo pid,command` to locate the detached test process.

## Why

BSD `ps` does not support the GNU-only `cmd` field used by the test. The `command` field provides the same full command line on both platforms.

### DIFF
--- a/pkg/werf/exec/detach_test.go
+++ b/pkg/werf/exec/detach_test.go
@@ -77,7 +77,7 @@ func findProcessLineByCommand(ctx context.Context, command string) (string, erro
 	b := backoff.NewConstantBackOff(time.Millisecond * 10)
 
 	operation := func() (string, error) {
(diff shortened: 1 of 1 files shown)


alexey-igrychev:
## Verification

- `TASK_X_REMOTE_TASKFILES=1 task --yes format paths="pkg/werf/exec"`, `task --yes build`, and `task --yes lint` passed.
- `task --yes test:unit paths="./pkg/werf/exec/..."` passed on macOS.
- Mutation: replaced the fixture's `exec.Detach` call with a no-op; the test failed with `line not found`, then passed again after restoration.
- Not run: full `task test:unit` is blocked by an unrelated, intermittent `pkg/true_git` failure in this worktree.

PR #7787, opened 2026-08-06
fix(storage): prevent concurrent final-stage list access

alexey-igrychev:
## Summary

Parallel final-stage copies could race with debug logging while updating the shared final-stage list cache. Stage-count reads now use the cache mutex.

## What

- Concurrent final-stage copy and stage-count logging access the cached list under its mutex.
- Final-stage cache contents and duplicate-stage handling remain unchanged.
- The storage-manager race test concurrently adds stages and reads their count.

## Why

Debug logging read the cache slice directly while another copy appended to it. The race detector reported this as `StagesList.AddStageID` concurrent access.

### DIFF
--- a/pkg/storage/manager/storage_manager.go
+++ b/pkg/storage/manager/storage_manager.go
@@ -168,6 +168,13 @@ func (stages *StagesList) AddStageID(stageID image.StageID) {
 	stages.StageIDs = append(stages.StageIDs, stageID)
 }
 
+func (stages *StagesList) Len() int {
+	stages.Mux.Lock()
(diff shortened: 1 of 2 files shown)


alexey-igrychev:
## Verification

- `task --yes test:unit paths="./pkg/storage/manager" parallel=false -- --race --focus="read stage count while adding stages"`.
- Mutation: removed `StagesList.Len` locking; race test failed with concurrent `AddStageID` write and `Len` read. Restored locking; test passed.

## Review focus

- Confirm every final-stage cache count uses `StagesList.Len` rather than reading `StageIDs` directly.

PR #7788, opened 2026-08-06
fix(buildah): prevent concurrent stderr access

alexey-igrychev:
## Summary

Buildah can continue writing command stderr while werf formats a returned error. The stderr collector now serializes those writes and reads.

## What

- Buildah command stderr uses a mutex-protected buffer.
- Error messages retain collected stderr output.
- A Linux race test reads the collected stderr while another goroutine writes it.

## Why

A plain `bytes.Buffer` was read after `Builder.Run` returned while its subprocess output writer could still append. The race detector reported concurrent buffer access.

### DIFF
--- a/pkg/buildah/native_linux.go
+++ b/pkg/buildah/native_linux.go
@@ -17,6 +17,7 @@ import (
 	"sort"
 	"strconv"
 	"strings"
+	"sync"
 	"syscall"
 	"text/template"
 	"time"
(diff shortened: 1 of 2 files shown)


alexey-igrychev:
## Verification

- Linux: `task --yes build`.
- Linux: `task --yes test:unit paths="./pkg/buildah" parallel=false -- --race --focus="read stderr while it is written"`.
- Mutation: removed `lockedBuffer.String` locking; race test failed with concurrent `Write` and `String`. Restored locking; test passed.

## Review focus

- Confirm `RunCommand` still includes all stderr written before error formatting.

PR #7789, opened 2026-08-06
fix(release): preserve v3 release version state

alexey-igrychev:
## Summary

After `main` is merged into `3`, release-please can mistake a newer v2 release commit for the v3 release baseline and generate a v2 version. The v3 branch now carries its own manifest state, seeded at `3.1.0`, while shared release policy lives in one root config.

## What

- On pushes to `3`, release-please loads shared `release-please-config.json` and `.release-please/3/manifest.json` rather than constructing its release state from generic `release-type` action input.
- The manifest pins the current v3 release to `3.1.0`; a merged `v2.*` release no longer selects the version baseline for the next v3 release.
- Generated release tags continue to use the `v` prefix, and external release creation remains unchanged (`skip-github-release: true`).
- The shared config contains strategy only; version state remains branch-local, so merging branches does not replace v3's manifest.
- Replaces incorrectly generated #7772 (`chore(3): release 2.75.5`).

## Why

The simple release-please mode searched merged history and, after #7778 imported `chore(main): release 2.75.4`, selected that release as branch `3`'s baseline. A branch-local manifest supplies the expected version explicitly; keeping unchanged policy in a root config avoids duplicate configuration.


alexey-igrychev:
## Verification

- Parsed workflow YAML and both branch-3 JSON files; confirmed `v3.1.0` is reachable from branch `3`; checked authored diff for whitespace errors.
- Not run: release-please action against GitHub API. It fetches config from target branch, so the final behavior is observable only after this PR lands on `3` and its next push runs the workflow.

## Review focus

- Confirm release-please-action@v5 enters manifest mode when action-level `release-type` is absent and fetches both files from `github.ref_name`.

alexey-igrychev:
## Verification update

Shared config is now root `release-please-config.json`; only `.release-please/3/manifest.json` remains branch-local. Parsed both JSON files and workflow YAML again; manifest stays `3.1.0`.

PR #7790, opened 2026-08-06
fix(release): preserve main release version state

alexey-igrychev:
## Summary

Release-please now tracks main's current release explicitly instead of inferring it from every release commit reachable through merged history. This gives main independent state while retaining one shared release configuration for all release lines.

## What

- On pushes to `main`, release-please loads the root `release-please-config.json` and `.release-please/main/manifest.json`.
- The main manifest pins the current release to `2.75.4`; releases from another line cannot select the next main version baseline.
- Generated release tags continue to use the `v` prefix, and external release creation remains unchanged (`skip-github-release: true`).
- The common config contains release strategy only; version state remains branch-local, so merging branches does not replace another line's manifest.

## Why

Simple release-please mode derives its baseline from merged history. That is unsafe for long-lived branches which exchange commits, because a reachable release from another line can become the apparent latest release. The manifest supplies the expected version directly, while a shared config avoids duplicating unchanged release policy.


alexey-igrychev:
## Verification

- Parsed workflow YAML and both JSON files; confirmed `v2.75.4` is reachable from `main`; checked authored diff for whitespace errors.
- Not run: release-please action against GitHub API. It fetches configuration from target branch, so final behavior is observable only after this PR lands on `main` and its next push runs the workflow.

## Review focus

- Confirm release-please-action@v5 enters manifest mode when action-level `release-type` is absent and resolves the main manifest from `github.ref_name`.

PR #7792, opened 2026-08-07
test: prevent command output deadlocks

alexey-igrychev:
## Summary

E2E command capture no longer blocks a verbose child process when stderr fills before stdout closes. Stdout and stderr now share a synchronized collector.

## What

- Test command execution drains stdout and stderr concurrently.
- `CancelOnOutput` continues to detect its marker in combined command output.
- The command helper regression test writes 128 KiB to stderr before stdout and completes before its context deadline.
- VERIFIED: on Linux, the previously hanging race-instrumented compose build-report spec completed; it then reported the separately tracked Nelm race.

## Why

`io.MultiReader` drained stdout to EOF before stderr. A child that filled stderr while keeping stdout open blocked in `pipe_write`, while the test harness waited for stdout EOF. The resulting deadlock hid race reports and exhausted E2E job timeouts.

Tracked in #7775.

### DIFF
--- a/test/pkg/utils/command.go
+++ b/test/pkg/utils/command.go
@@ -1,7 +1,6 @@
 package utils
 
 import (
-	"bufio"
 	"bytes"
 	"context"
 	"encoding/base64"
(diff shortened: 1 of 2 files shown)


alexey-igrychev:
## Verification

- `task --yes test:unit paths="./test/pkg/utils" parallel=false`.
- Mutation: restored sequential `io.MultiReader` draining; regression test hit its 2-second deadline and failed with `signal: terminated`. Restored concurrent capture; test passed.
- Linux: focused race compose build-report spec completed in 5.535s; it reported four known Nelm `Configuration.Init` races instead of hanging.

## Review focus

- Confirm `CancelOnOutput` still cancels after a marker written to either output stream.

PR #7794, opened 2026-08-07
test(dev): unify test provenance

alexey-igrychev:
## Summary

Tests no longer carry labels based on their origin. Review instructions now choose challenge depth from the change's risk and test surface, not its author.

## What

- Test files and identifiers formerly marked with AI provenance now use ordinary test names while preserving Go and Ginkgo discovery.
- Test and verification diffs invoke `challenge-review` regardless of who authored them.
- Test assertions, runner filters, build tags, and production behavior do not change.
- UNVERIFIED: full local `task test:unit` exits non-zero because `TestGetContainerRegistryMirror_MergesWerfEnvMirrorsBeforeBuildahConfigMirrors` depends on host registry-mirror configuration; a host with the expected Buildah configuration settles this check.

## Why

Origin labels made equivalent tests appear to have different status and applied review policy inconsistently. Use behavior and risk as the only basis for test and review treatment.


alexey-igrychev:
## Verification

- `task format`, `task build`, `task lint` — passed.
- `task test:unit` — runner exits non-zero while reporting `pkg/true_git` after 51/51 passing specs; no failing spec is emitted.
- `task test:unit parallel=false` — fails unrelated `cmd/werf/common` registry-mirror assertion: expected two mirrors, host exposes one.
- Mutation: not run — this change renames existing tests and leaves assertions and production behavior unchanged.

## Review focus

- Two independent reviews checked test discovery, marker removal, and policy triggers. The review findings were incorporated before rebase; current base replaces the former agent-specific skill with `challenge-review`.

PR #7795, opened 2026-08-07
test(dev): unify test provenance

alexey-igrychev:
## Summary

Tests on the 3 branch no longer carry labels based on their origin. Review instructions now choose challenge depth from the change's risk and test surface, not its author.

## What

- All test files and identifiers formerly marked with AI provenance now use ordinary test names while preserving Go and Ginkgo discovery.
- The two tests previously gated by the `ai_tests` build tag now run with the normal unit suite.
- Test and verification diffs invoke `challenge-review` regardless of who authored them.
- Test assertions, runner filters, and production behavior do not change.

## Why

Origin labels made equivalent tests appear to have different status and kept two test files outside the regular suite. Use behavior and risk as the only basis for test and review treatment.


alexey-igrychev:
## Verification

- `task format`, `task build`, `task lint`, `task test:unit` — passed.
- Mutation: not run — this change alters test discovery names and build tags only; test assertions and production behavior are unchanged.

## Review focus

- Confirm all renamed test files retain the intended package and Go/Ginkgo discovery; the two former `ai_tests`-gated files now run in the ordinary unit suite.

PR #7798, opened 2026-08-09
fix(buildah): prevent concurrent Dockerfile build races

alexey-igrychev:
## Summary

Concurrent native Buildah Dockerfile builds no longer mutate one shared Buildah options object. Parallel image builds no longer produce a race-detector failure in this path.

## What

- Create a per-operation copy of native Buildah common build options before Dockerfile build and builder creation.
- Preserve parallel Dockerfile build execution; no user-facing flag or configuration changes.
- VERIFIED: focused remote Linux race E2E `Build with dockerfile outside context.*Native Buildah with chroot isolation` passed.

## Why

`BuildFromDockerfile` passed a pointer to shared `defaultCommonBuildOptions`, then assigned operation-specific secrets and SSH sources through that pointer. Concurrent builds therefore wrote the same object. Each operation now owns its options copy.

Tracked in #7775.

### DIFF
--- a/pkg/buildah/native_linux.go
+++ b/pkg/buildah/native_linux.go
@@ -418,6 +418,7 @@ func (b *NativeBuildah) BuildFromDockerfile(ctx context.Context, dockerfile stri
 		return "", err
 	}
 
+	commonBuildOpts := b.defaultCommonBuildOptions
 	buildOpts := define.BuildOptions{
 		Isolation:               define.Isolation(b.Isolation),
 		Args:                    opts.BuildArgs,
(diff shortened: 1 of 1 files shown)


alexey-igrychev:
## Verification

- Rebased onto current `3` (`f759becbb`), which carries the non-serializing Buildx fork (#7803) and the command-drain deadlock fix (#7792). Only the 2-line `pkg/buildah/native_linux.go` change remains on top.
- `TASK_X_REMOTE_TASKFILES=1 task --yes format && task --yes build && task --yes lint && task --yes test:unit` passed locally.
- Mutation: reverting the per-operation copy back to `&b.defaultCommonBuildOptions` reintroduces the `-race` failure in the focused remote Linux Buildah Dockerfile-build E2E (concurrent `Secrets`/`SSHSources` writes through the shared pointer); the copy makes it pass.

## Review focus

- The copy is shallow, which is sufficient here: after the copy only `Secrets` and `SSHSources` are written, both as whole-slice-header assignments on the local value (never appends into shared backing arrays); `Ulimit` is set once at construction and never mutated.

## Follow-up

- [ ] The earlier red `e2e_simple` (1h0m21s, run 31314346419) was NOT this diff: that run sat on the old base carrying the serializing Buildx fork `ec8211f6fecb`, and the hang matches its documented mutex-backpressure deadlock (whole-build serialization). `e2e_complex` (native Buildah, no Buildx) passed in the same run. The rebase base no longer contains that fork, so re-running this scope on the rebased branch is the confirmation.

PR #7797, opened 2026-08-07
test(buildx): use race-fix fork

alexey-igrychev:
## Summary

Race-instrumented Docker builds temporarily use `werf/3p-buildx` while the upstream Buildx log-output race is pending review. This lets the bounded compose race scope finish and exposes later reports instead of exiting on the known Buildx race.

## What

- Replace `github.com/docker/buildx` with `github.com/werf/3p-buildx` at the `docker/buildx#4003` fix commit.
- Keep an in-line `go.mod` removal condition: remove the replacement after upstream PR `#4003` merges.
- VERIFIED: local `task format && task build && task lint && task test:unit` passed.
- VERIFIED: focused remote Linux race compose build-report scope passed with the replacement.
- No werf API, flag, configuration, or non-race build behavior changes.

## Why

Buildx progress printers mutate one process-global logrus output and race during concurrent Dockerfile builds. The upstream patch serializes those pauses; until it lands in a released dependency, this fork prevents that known report from masking subsequent race investigation.

Depends on [docker/buildx#4003](https://github.com/docker/buildx/pull/4003).

PR #7800, opened 2026-08-10
fix(buildx): restore concurrent Dockerfile builds

alexey-igrychev:
## Summary

Restore concurrent Dockerfile builds by removing the temporary Buildx replacement from branch `3`.

## What

- Remove the `werf/3p-buildx` replacement introduced by #7797.
- Restore the upstream Buildx `v0.33.0` dependency used before the replacement.
- No CLI, configuration, or persisted-data changes.

## Why

The temporary fork serializes Buildx log pauses with a process-global mutex. Its pause lifetime spans the whole Buildx progress display, so parallel Dockerfile builds can serialize for their full duration. Remove it until docker/buildx has a concurrency-safe upstream fix.

Tracked in #7775.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -456,7 +456,6 @@ require (
 )
 
 replace (
-	github.com/docker/buildx => github.com/werf/3p-buildx v0.0.0-20260807135054-ec8211f6fecb // temporary race fix; remove after docker/buildx#4003 merges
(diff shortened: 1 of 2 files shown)


alexey-igrychev:
## Verification

- `TASK_X_REMOTE_TASKFILES=1 task --yes format && task --yes build && task --yes lint && task --yes test:unit` passed locally.

## Follow-up

- [ ] Keep the Buildx workaround only on the diagnostic race branch until a concurrency-safe upstream fix is available.

PR #7802, opened 2026-08-10
fix(logboek): prevent concurrent stream races

alexey-igrychev:
## Summary

Concurrent build worker output no longer races while logboek temporarily changes stream indentation.

## What

- Update `github.com/werf/logboek` from `v0.7.1` to the merged race fix revision.
- VERIFIED: integration race scopes no longer report the logboek `StateAndModes` race on the test branch.
- No werf CLI, flag, or configuration surface changes.

## Why

Logboek temporarily reset shared indentation during callbacks while concurrent stream formatting read it. The upstream fix uses atomic indentation access without holding a callback-spanning mutex.

Tracked in #7775.

PR #7803, opened 2026-08-10
fix(build): keep concurrent Dockerfile builds parallel under race fix

alexey-igrychev:
## Summary

Branch `3` currently serializes concurrent Dockerfile builds: the `werf/3p-buildx`
replacement in `go.mod` fixes a data race in `docker/buildx`'s `logutil.Pause` with a
global mutex that is held for a build's entire progress-display lifetime, so two
Dockerfile builds in one process cannot run at the same time. This points the
replacement at a fork commit that fixes the same race without a global lock, restoring
concurrency.

## What

- Concurrent Dockerfile builds in a single werf process run in parallel again; the
  previous fork serialized them.
- The `-race`-detected data race on `logrus.StandardLogger()` in `logutil.Pause` stays
  fixed: pauses are multiplexed through one refcounted per-logger writer, and `Pause`
  never blocks on another active pause.
- Only the `go.mod`/`go.sum` `docker/buildx => werf/3p-buildx` pin changes; no CLI,
  configuration, or persisted-data changes.

## Why

#7800 removed this replacement precisely because its global mutex serialized concurrent
builds, and said to keep it out until an upstream-safe fix existed. #7802 was branched
before that revert and reintroduced the old pin on merge, so `3` regressed to the
serializing fork. The new fork commit (backport of docker/buildx#4007) replaces the
global mutex with a per-logger refcounting writer: the first `Pause` wraps the logger
output once, each `Pause` bumps a counter, output is buffered while any pause is active
and flushed in order when the last one resumes. Remove this replacement once the fix
lands in an upstream buildx release.

Fixes: f80e827577896 ("fix(logboek): prevent concurrent stream races (#7802)")

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -456,7 +456,7 @@ require (
 
 replace (
 	github.com/deislabs/oras => github.com/werf/3p-oras v0.9.1-0.20260408144000-3b8c77eb09e8 // used by bundles, not maintained
-	github.com/docker/buildx => github.com/werf/3p-buildx v0.0.0-20260807135054-ec8211f6fecb // temporary race fix; remove after docker/buildx#4003 merges
+	github.com/docker/buildx => github.com/werf/3p-buildx v0.0.0-20260810132822-84b2b5a524c2 // temporary race fix; remove after docker/buildx#4007 merges
 	github.com/spf13/cobra => github.com/werf/3p-cobra v0.0.0-20260403075225-552c82797324 // adds EnableErrorOnUnknownSubcommand, not yet in upstream
 	oras.land/oras-go => github.com/werf/3p-oras-go v1.2.8-0.20260408140625-72dd516ce0aa // used by bundles, not maintained
 )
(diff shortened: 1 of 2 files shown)


alexey-igrychev:
## Verification

- `task build` (macOS, non-CGO) with the new pin: succeeds, binary produced.
- Mutation (in the fork, `util/logutil`): reverted `pause.go` to upstream → `TestPauseInitializationConcurrent` fails with the `-race` report on `logrus.SetOutput`/`l.Out`; simulated the old global-mutex fork by holding the writer mutex across resume → `TestPauseConcurrent` times out; dropped the `pauses == 0` flush guard → `TestPauseBuffersOverlappingOutput` fails; removed the resume `sync.Once` → strengthened `TestPauseResumeIsIdempotent` fails. Each reverts clean.
- Not run: `-race` e2e Dockerfile-build suite against this pin on Linux — pending; the unit suite proves the pause layer, not the werf-side concurrent build path.

## Review focus

- `go.mod`/`go.sum` pin only. Confirm the pseudo-version `v0.0.0-20260810132822-84b2b5a524c2` resolves to `werf/3p-buildx` HEAD and that no stale `ec8211f6fecb` hash remains in `go.sum`.

## Follow-up

- [ ] Drop this replacement and bump to a stock `docker/buildx` release once docker/buildx#4007 (or an equivalent) merges and ships.

PR #7770, opened 2026-08-06
test(ci): run the race detector across all daily tests only

alexey-igrychev:
## Summary

Enable Go's race detector across the full daily test matrix instead of on every pull request. Per-PR CI keeps race only on `unit` (which already builds with `-race`); all e2e and integration scopes now run against a `-race` werf binary on the daily schedule, where a 1.5–3× slowdown is acceptable.

## What

- Per-PR `tests.yml` runs no race-instrumented e2e/integration jobs; `unit` stays `-race` as before.
- Every `test_daily.yml` test job runs against a `-race` binary: e2e per-k8s-version (simple/complex/extra), integration_main, integration_git, and the per-k8s / per-container-registry / per-k8s-and-cr matrices.
- The `set-up-prebuilt-werf-test-binary` action and the `build-with-coverage` task forward `raceDetectorEnabled`, so daily coverage builds are race-instrumented (`-race` and `-cover` combine in one `go build`; VERIFIED: `go version -m` on the built binary shows `-race=true -cover=true`).
- `raceDetectorEnabled` input added to `_test_e2e_per-k8s-version.yml`, `_test_integration_per-k8s-version.yml`, `_test_integration_per-container-registry.yml`, `_test_integration_per-k8s-version-and-container-registry.yml`; it already existed on the regular workflows.
- Integration workflow job timeouts raised from 60 to 120 minutes to absorb race overhead.
- No product code, CLI, or persisted-data changes; workflow and Taskfile only.

## Why

Race coverage found and drove fixes for several concurrency bugs (buildx log pause, Buildah build options, go-containerregistry writer, logboek state, nelm progress printer), but a full `-race` run costs 1.5× on most scopes and ~3× on `e2e_complex` (51m vs 17m). Paying that on every PR slows the feedback loop for no proportional benefit once the known races are fixed; the daily schedule keeps ongoing race coverage without taxing per-PR CI. `unit` is cheap under race and already always instrumented, so it stays on the PR path.

PR #7805, opened 2026-08-10
fix(deploy): prevent progress printer race during release tracking

alexey-igrychev:
## Summary

Bump `werf/nelm` to `06096861c106` so concurrent release tracking no longer data-races on the progress table printer. The race was reproduced under the race detector in `e2e_complex` during release uninstall, where `ProgressTablesPrinter.Stop` read state that the `Start` worker was still writing.

## What

- Deploy no longer triggers a data race in `ProgressTablesPrinter.Start`/`Stop` during concurrent release tracking (werf/nelm#690).
- Deploy now matches Kubernetes resource condition types case-insensitively during tracking (werf/nelm feat + werf/kubedog bump to `e8f61a4bc90a`).
- The nelm range `8b420d326cbc..06096861c106` also carries werf/nelm#688 (concurrent Helm configuration race); nelm branch `2` was rebased between the two pins, so the range is `diverged` rather than strictly linear.
- VERIFIED: `task build` (macOS, non-CGO) succeeds against the new pins.

## Why

werf/nelm#690 moves printer state initialization before the tracking goroutine starts, so `Stop` no longer reads a field the worker is concurrently assigning. It is the last dependency race left open in `e2e_complex`. nelm is an ordinary versioned dependency, so the fix lands as a nelm PR plus this bump; the intermediate commits on branch `2` come along with the tip.


alexey-igrychev:
## Verification

- `TASK_X_REMOTE_TASKFILES=1 task --yes build` succeeds with nelm `06096861c106` and kubedog `e8f61a4bc90a`; `go mod tidy` clean.
- Mutation lives upstream: werf/nelm#690 added `TestAI_ProgressTablesPrinterStartInitializesBeforeReturn`, which fails when the fix (init before goroutine start) is reverted.
- Range read from `gh api repos/werf/nelm/compare/8b420d326cbc...06096861c106` and `repos/werf/kubedog/compare/85087f97ed99...e8f61a4bc90a`, not from release notes.

## Follow-up

- [ ] Rebase the race CI branch (#7770) to drop its temporary nelm pin `48790ffcba47` once this lands on `3`.

PR #7806, opened 2026-08-10
feat(cleanup): add registry-side cleanup report

alexey-igrychev:
## Summary

`werf cleanup` and `werf purge` can now write a machine-readable JSON report of every registry object they kept and deleted. Until now the only way to find out what a cleanup did was to grep ANSI colour codes out of its log — which the documentation actually recommends — so generating a keep-list or auditing a deletion was unusable anywhere colour is stripped.

Ported by hand from #7737, registry half only.

### Flags and output

- `--save-cleanup-report` (default `false`, `$WERF_SAVE_CLEANUP_REPORT`) on `cleanup` and `purge` writes the report; without it no file is created and no behaviour changes.
- `--cleanup-report-path` (default `.werf-cleanup-report.json`, `$WERF_CLEANUP_REPORT_PATH`) sets the path; an extensionless path gains `.json`, and any extension other than `.json` fails with `invalid --cleanup-report-path "x.yaml": extension must be either .json or unspecified`.
- The report holds `command`, `dryRun`, `repo`, `finalRepo` when a final repository is configured, and always-present `kept` and `deleted` arrays — never `null`.
- Item types are `stage`, `finalStage`, `customTag`, `rejectedStage`, `rejectedStageMarker`, `imageMetadata`, `managedImage`, `importMetadata`; the set is extensible, so a consumer filters by the types it knows.
- An `imageMetadata` item carries `imageName`, except for metadata whose image is absent from `werf.yaml`, where no name is recoverable and the item carries the internal `id` instead — `werf purge` never consults `werf.yaml`, so its reports use `id` throughout.
- `werf purge` reports an empty `kept`, because it deletes everything; only `werf cleanup` populates it.
- `kept` holds protected stages and final stages, each with the `reason` it was kept (`used in Kubernetes`, `git policy`, …), plus custom tags that were retained; `deleted` holds every registry object actually removed. A keep-list is `jq -r '.kept[] | select(.type == "stage") | .tag'`.
- `kept` and `deleted` are unordered, because deletions run in parallel.
- The documented keep-list workflow changes: the recipes now read the report with `jq` instead of grepping ANSI escape codes out of the cleanup log, which produced an empty file wherever colour is stripped. The tag colour legend is dropped with them, so cleanup output is no longer presented as something to parse. Both language versions updated.

### Guarantees

- `--dry-run --save-cleanup-report` records exactly the set a real run would have deleted. VERIFIED: same input through both paths produces equal `deleted` sets, in unit tests per funnel.
- A deletion that fails is absent from the report, together with any follow-up it cancelled — a rejected-stage marker is kept for the next cleanup to retry, and neither it nor the failed tag is reported.
- An unwritable report path — including a path that is an existing directory — fails the command before the first deletion, so cleanup never deletes objects it cannot record.
- A run that deletes and then fails still writes the report and still exits non-zero.
- A reader never observes a partially written report; the file is written to a temp sibling and renamed. UNVERIFIED: this is atomic visibility, not crash durability — there is no `fsync`, and no test covers a host crash mid-write.

### Unchanged

- Cleanup log output is byte-identical — no existing log statement's text, style or position changed, and colours are still emitted.
- `werf host purge` is untouched. It runs a full registry purge when given `--project-name`, so the report would apply, but the command documents itself as host-only and never mentions a registry; wiring a registry report into it is a separate decision from this one.
- The host-cleanup report from #7737 is not ported: no `--save-host-cleanup-report`, no `.werf-host-cleanup-report.json`, no `spaceReclaimed`, and no change under `pkg/host_cleaning`.

## Why

Cleanup deletes irreversibly and left no machine-readable trace, so the documented way to build a keep-list was to parse coloured terminal output — fragile, and empty in CI that strips colour. The same gap makes `--dry-run` far less useful than it should be: the planned deletion set exists only as prose in a log.

The writer is deliberately stricter than the sibling `--save-build-report`, which is a plain `os.WriteFile` with no pre-check: a build report can be regenerated by rebuilding, whereas a lost cleanup report is the only record of deletions that cannot be undone. That is also why the writability check runs before the first deletion rather than at save time.

Two alternatives were rejected. An `apiVersion`/`kind` envelope, as in #7737, would make this the only werf report carrying one — neither the build nor the deploy report has one, and with the host report out of scope `kind` discriminates nothing. A per-item `repo` field was dropped because `repo` and `finalRepo` are already top-level and an item's repository follows from its `type`.


alexey-igrychev:
## Verification

- `werf purge --save-cleanup-report` run end to end against a throwaway registry, with the report's deleted stage tags cross-checked against the command output. A stock `registry:2` rejects `DELETE`, so the e2e suite needs `REGISTRY_STORAGE_DELETE_ENABLED=true` or the purge fails for reasons unrelated to the report.
- `doc:gen` is idempotent on this branch and regenerates exactly the three command references.
- Mutation: recorded the metadata key unconditionally as `imageName` again → `TestDeleteImageMetadata_ReportUsesIDForUnresolvableMetadata` failed.
- Mutation: pointed `GetFinalProtectedStageDescSetByReason` at the non-final stage set → `TestGetFinalProtectedStageDescSetByReason` failed.
- Mutation: deleted the kept-recording line in `cleanupFinalStages` → `TestCleanupFinalStages_ReportRecordsKeptFinalStageWithReason` failed.
- Not run: a cleanup against a registry holding stages that are still in use, which is what would cover the `kept` recording for unused stages; the final-stage `kept` path is covered by a unit test instead.

## Review focus

- `deleteImageMetadata` and its two identity fields. The classification rests on `isNonexistentImage` having exactly one assignment, from the not-managed map returned by `groupImageMetadataTagsByImageName`; if a second assignment appears, items would silently report a hash as `imageName` again.
- The recording sites sit inside `parallel.DoTasks` success callbacks. Moving one out of its callback would over-report a partially failed batch.
- Confirm the writability pre-check is genuinely ahead of every destructive path in all three commands, not just the one each command obviously takes.

## Follow-up

- [ ] Decide whether `artifact` should be reported once orphaned-artifact cleanup exists on `main` — this port dropped that item type because `deleteOrphanedArtifacts` is not present here.
- [ ] Decide separately whether `werf host purge --project-name` should write a report. It performs a full registry purge, so the flags would apply, but its own help describes it as host-only and never mentions a registry — worth settling that documentation mismatch first.

PR #7808, opened 2026-08-10
chore(git): merge main into 3

alexey-igrychev:
## Summary

Brings the six commits `main` gained since the last sync into `3`. No user-visible behavior of `3` changes except the registry-side cleanup report, which arrives from `main` and is wired onto `3`'s cleanup code.

## What

- `werf cleanup` and `werf purge` on `3` accept `--save-cleanup-report` (default `false`) and `--cleanup-report-path` (default `$WERF_CLEANUP_REPORT_PATH` or `.werf-cleanup-report.json`), and record deleted and kept stages, custom tags, image metadata and managed images.
- The cleanup report records nothing for import metadata: `3` no longer stores import metadata at all, so `main`'s import-metadata report items and their two tests are dropped.
- The build report on `3` keeps `Operations` and `StageCache` under `--log-debug`, minus the `docker daemon API` operation: `3`'s `docker/cli` hands back a `moby/moby/client`, which exposes no accessor for the `*http.Client` `main`'s round-tripper wraps, so `pkg/docker/timing.go` is not carried over. Every other opstats measurement (image build/pull/push, registry API, git, stapel, context add files, lock waits) is unchanged.
- No `main` change to a subsystem `3` has removed is resurrected: the synchronization lock managers, import metadata, nameless images (`WERF_DOCKER_IMAGE_NAME` for an unnamed image), `--virtual-merge`, the in-cluster namespaces scan and `CalculateDependencyImportChecksum` stay absent, and the `main`-side tests asserting them are dropped with them.
- `pkg/true_git` keeps `3`'s ginkgo suite for `include.path` forwarding and submodule object-store reuse; `main` renamed its testify tests over the same ground and those are dropped, so coverage of that behavior is unchanged.
- Dependencies stay at `3`'s pins, which are newer throughout and no longer carry the `kubedog-for-werf-helm` and `nelm-for-werf-helm` forks.
- The generated CLI reference and the JSON-report documentation match the merged code: `task doc:gen` produces no diff.

## Why

`3` is 213 commits ahead of the last sync point and has removed whole subsystems (synchronization, import metadata, nameless images, virtual merge) that the new `main` commits instrument or extend, so 19 files conflicted and every conflict is the same question: keep `3`'s removal or take `main`'s edit. Letting the gap grow makes each later sync harder, and the cleanup report is a feature users of `3` would otherwise lose until the next merge.

Dropping `pkg/docker/timing.go` was chosen over rebuilding the moby client by hand: `main` wraps the transport by reconstructing the client from its own `HTTPClient()`, and reproducing the endpoint's TLS and dialer setup without that accessor would trade a debug-only metric for a hand-maintained client constructor.


alexey-igrychev:
## Verification

- `task format`, `task build`, `task lint` (0 issues, including `./test/...`), `task test:unit` (49 suites), `task doc:gen` (no diff) and `./bin/werf cleanup --help` (both report flags present) all run on this merge commit locally on macOS.
- Not run: `task test:e2e` and `task test:integration`, including the `purge_report` e2e suite that arrives from `main` — they need Linux with Docker and kind. The Buildah backend is not compiled by a macOS build either, so `pkg/container_backend/buildah_backend.go` typechecks here only through the linter.
- `task mock:check` fails on this branch, and equally on `origin/3`: mockgen emits an import grouping that `task format` then regroups in three already-committed mock files, none of which this merge touches.

## Review focus

- Whether every conflict resolved toward `3`'s removal is right, in particular that the cleanup report genuinely has nothing to say about import metadata on `3`.
- `pkg/cleaning/cleanup_test.go`: `main`'s new report tests were rebased onto `3`'s split-storage fake (`newFakeStorageManagerWithSplitStorages`, `UnregisterStageCustomTag` routed to the meta storage), so the fake now serves both sets of tests.
- The dropped `pkg/docker/timing.go` — confirm nobody wants the daemon round-trip metric on `3` badly enough to rebuild the moby client.

## Follow-up

- [ ] Run the `purge_report` e2e suite from `main` against `3` on a Linux host with `REGISTRY_STORAGE_DELETE_ENABLED=true`.
- [ ] Decide whether the `docker daemon API` opstats operation should return on `3` (needs a transport hook for `moby/moby/client`).

PR #7817, opened 2026-08-27
fix(deploy): apply .helmignore when reading the chart

### LINKED ISSUE #6747: [giterminism] Read the helm chart taking into account the .helmignore file

dmmordvi:
## Summary

werf read the chart with its own file reader and pulled in every file of the chart directory, so `.helmignore` was ignored entirely: on the same chart werf and helm disagreed, and files helm would have dropped reached the render and the published bundle.

Repro:

```yaml
# werf.yaml
project: helmignore-probe
configVersion: 1
```

```
# .helm/.helmignore
templates/ignored.yaml
```

`.helm/templates/ignored.yaml` and `.helm/templates/kept.yaml` each render a ConfigMap. `werf render` emitted both; only `kept` is correct.

Fixes #6747.

### Breaking

- BREAKING: a chart carrying a `.helmignore` that werf silently ignored now excludes the files it lists, so the rendered manifests and the published bundle lose them; a chart that relied on werf's old behavior has to drop the rule.
  The way out is editing `.helmignore`; a release note is the only mitigation.
- BREAKING: helm's default rules apply even when the chart has no `.helmignore`, so a dotfile directly under `templates/` (e.g. `templates/.gitkeep`) is dropped. Chosen for full helm compatibility over a narrower blast radius.

### Chart contents

- The whole chart tree is filtered by the root `.helmignore`: a rule from the root chart's file applies to vendored subcharts as well, and a subchart's own `.helmignore` is not applied — it stays an ordinary chart file. The default `templates/.?*` rule is path-structural, so it does not reach `charts/<name>/templates/`.
  VERIFIED: `helm package` (helm 3.20.1) on such a chart drops `notes.txt` and `charts/sub/notes.txt` by the root rule, keeps `charts/sub/templates/kept.yaml` despite the subchart's `.helmignore`, and keeps `charts/sub/templates/.gitkeep`.
- The effective chart tree is assembled before filtering, so files delivered through werf includes are filtered too, and a path excluded by `.helmignore` cannot come back from an include. The local `.helmignore` wins over one delivered by an include; a chart delivered entirely through an include is filtered by the `.helmignore` that include carries.
- A directory rule drops the whole subtree, not just the directory's direct children.
- Rule classes behave as in helm: exact path, basename at any depth, glob, root-anchored `/`, trailing-slash directory, directory without a slash.
  VERIFIED: file set compared against helm 3.20.1 on a chart exercising every class, in the default and `--loose-giterminism` modes.
- A `**` rule fails the render with `double-star (**) syntax is not supported`, as in helm.
- A negated rule (`!pattern`) inverts the whole rule set instead of re-including a file, and also drops `Chart.yaml`. Unsupported in helm too, where such a chart fails with `Chart.yaml file is missing`; already documented as a deviation. When a rule set leaves no files, werf now names `.helmignore` as the reason instead of reporting the chart directory as not found.

### Giterminism

- An ignored path is not checked against giterminism at all, so `.helmignore` can exclude an uncommitted file — the advice `docs/pages_en/usage/project_configuration/giterminism.md` already gives. A file that stays part of the chart is unaffected.
- The rules are matched against the chart-relative path with its symlink parts kept, so they still apply when `templates/` or the chart directory itself is a symlink, and both giterminism modes produce the same file set for such a chart.
- A symlink escaping the project inside an ignored directory no longer fails the render under `--loose-giterminism` while succeeding by default: helm skips such a directory outright.
  VERIFIED: symlink matrix checked against helm — escaping and dangling targets, each both ignored and not ignored; paths that stay in the chart are still rejected by giterminism in both modes.

### Unchanged

- `.helmignore` filters chart directories only: `werf.yaml` templates and the files Go-template rendering reads are unaffected by it, and a rule matching one of their paths does not exclude it.
- No documentation changes: `docs/pages_en/usage/distribute/bundles.md` already documents `.helmignore` as excluding files from the published chart, along with the format's deviations. This makes the code match the docs.

## Why

`FileReader.loadChartDir` carried a `// TODO helmignore support` marker and walked the chart directory with werf's own reader, which knows about giterminism but not about `.helmignore`. Leaving it meant werf and helm disagreeing on the contents of the same chart, and two documentation pages promising behavior the code never had.

The filtering had to go into `FileManager.LoadChartDir`, not only `FileReader`: `FileManager` is the chart reader registered at runtime, and it merges the local chart directory with the files delivered through werf includes. Filtering only the local source left the imported half unfiltered and, because an excluded path was indistinguishable from an absent one, let an include add an excluded local path back with different content.

The rules are parsed by `github.com/werf/nelm/pkg/helm/pkg/ignore` rather than reimplemented, so the semantics cannot drift from helm. The rejected alternative was matching in the walk callback instead of inside the walk: it reads simpler, but an ignored path would still be resolved and checked against giterminism first, which breaks excluding uncommitted files and makes a symlink inside an ignored directory behave differently depending on `--loose-giterminism`.


dmmordvi:
## Verification

- Compared the resulting file set against helm 3.20.1 on a chart exercising every rule class, and on the symlink matrix (escaping and dangling targets, each both ignored and not ignored), in the default and `--loose-giterminism` modes. Also ran `werf render` on a hand-built project with a binary from this branch: before the fix the ignored ConfigMap rendered, after it only the kept one did. No cluster involved.
- Mutation, caught: unwiring the skip predicate (8 specs); dropping `AddDefaults()` (3); forcing `isDir` false (2); disabling the flat commit-list filter (2); matching the resolved path instead of the logical one (2, now including an enforced-mode spec); dropping the parent-directory loop, and the same loop reduced to the immediate parent (1 each); swallowing the `.helmignore` parse error (1); reading the rules file outside giterminism (2); resolving it only on disk (1); restricting the flat filter to paths containing a slash (2).
- Mutation, initially uncaught and now covered: reading `.helmignore` outside giterminism, resolving it only on disk, and restricting the flat filter to paths with a slash all left the suite green — the enforced mode had specs for directory rules only, and none for a root-level file, a root-anchored rule, the helm defaults, or where the rules file comes from. Also uncaught: dropping the parent-directory loop, since every directory-rule spec ran in the loose mode. Specs added in e01891bac and 76f803c9c; each mutation now fails.
- Not run: `test/e2e/converge` and `test/e2e/bundle-publish-apply` need Linux with kind and a registry. Their assertions expected `not-deployed-because-in-helm-ignore` while the fixtures render `not-deployed-because-in-helmignore`, so they asserted the absence of a resource no fixture ever created and passed under any behavior; this branch fixes the name. The pre-fix binary renders exactly that resource, which is what made the typo detectable.

## Review focus

- `skipConfigurationPathFunc` in `configuration.go` and the `skipPathFunc` threading in `fs.go`: the predicate has to see the path with its symlink parts kept, while the giterminism checks keep operating on the resolved one. Getting these two crossed is the bug the last commit fixes, and it is invisible unless a chart path goes through a symlink.
- `github.com/werf/nelm/pkg/helm/pkg/ignore` is a deep import. On branch `3` this is already how werf imports nelm; on `main` it is the first one.
- Confirm the breaking claims are acceptable as-is, in particular dropping dotfiles under `templates/` in charts that carry no `.helmignore` at all.
- The two giterminism modes reach the same outcome by different code: the loose one filters inside the walk, the enforced one filters the flat commit file list where a directory rule cannot match a file. Both paths now have a directory-rule spec; a further rule class added to one path needs its counterpart on the other.

- The `ListCommitFilesWithGlob` mock now mirrors `(*git_repo.Base).WalkCommitFiles` — it resolves the walked directory and reverse-resolves each entry. The previous shape resolved nothing, which made the enforced mode structurally unable to express a symlinked chart and hid every walk-side path bug behind the flat filter.

## Follow-up

- [ ] Release note for both breaking claims — it is the only mitigation for charts relying on werf's old behavior.
- [ ] Same fix on branch `3`: identical patch with `nelmcommon.BufferedFile` instead of `file.ChartExtenderBufferedFile`, separate PR.

PR #7818, opened 2026-08-31
feat(deploy,secrets): write authenticated secret values

alexey-igrychev:
## Summary

Werf v3 now uses common-go branch `1`, adding authenticated AES-GCM secret encryption while retaining read support for existing AES-CBC secrets. The v2-to-v3 migration guide now has a dedicated secret-migration section with the upgrade order and data-handling limits users need before re-encrypting a repository.

Related: https://github.com/werf/nelm/issues/677

### Breaking

- BREAKING: secrets newly written by werf v3 use AES-GCM formats that earlier werf or nelm releases cannot read. Upgrade every CI job and saved deploy-plan consumer before re-encrypting a repository.
- Existing AES-CBC encrypted values remain readable, but new writes use authenticated formats.

### Secret values

- Whole-blob secrets use version 2 and YAML scalar secrets use version 3, preventing whole-blob values from being interpreted as scalar metadata.
- YAML scalar tags, block styles, and value comments survive a secret-values edit cycle. Existing values that lost their original type remain strings until re-entered.
- Whole-blob ciphertext can be used as a scalar in an additional secret-values file without frame corruption.
- A dedicated v2-to-v3 migration section, in English and Russian, documents the upgrade order, mixed-version CI and deploy-plan risk, absent downgrade writer, irrecoverable historic scalar types, and cleartext value comments.

## Why

The previous common-go AES-CBC format accepted tampered data and wrong keys without reliable detection, while secret YAML encryption discarded scalar metadata. Werf v3 needs the common-go v1 implementation to participate in the coordinated authenticated-secret rollout with nelm v2, and users need migration guidance before their writers and readers diverge.


alexey-igrychev:
## Verification

- Remote Linux (Go 1.25.5): `task format`, `task build`, and `task lint` passed against the dependency update.
- `task test:unit` has one failure, identical on origin/3: `pkg/buildah/native_linux_test.go:354` observes host `/etc/containers/registries.conf.d/mirrors.conf`; 34/35 specs in that suite pass. The dependency bump does not touch Buildah.
- The authored v2-to-v3 migration guide now has a dedicated secret-migration section in both English and Russian; no generated CLI reference is affected.

## Review focus

- This pin consumes the versioned AES-GCM secret format: version 2 whole blobs and version 3 YAML scalar frames. Version 16 CBC remains readable only.

## Follow-up

- [ ] Document the minimum nelm v2 and werf v3 releases needed to read AES-GCM secret values before release.
- [ ] Close https://github.com/werf/nelm/issues/677 manually after the release documentation is in place.

PR #7819, opened 2026-08-31
fix(build, stapel): stop panicking when the base image is gone at commit

alexey-igrychev:
## Summary

Building a stapel stage with the Docker Server backend aborted with `panic: runtime error: info must be set for image "<ref>"` when the base image reference disappeared from the local backend while the stage's build container was running. The panic needs a pre-existing host state — something outside the werf process removing local images mid-build, e.g. a prune on a container backend shared with other jobs — so it does not reproduce from a clean checkout.

## What

- A stage whose base image tag is removed after its build container has started now commits normally instead of failing. VERIFIED: on a Linux host, `docker rmi -f <base stage tag>` during the stage's run panics before this change and builds through after it.
- A base image missing before the container starts still fails the build, now with the `docker run` error rather than a panic.
- The base image is inspected once per stage, before the container runs, by the same reference `docker run` receives — the committed image id when the stage was built in this process, the stage name otherwise. Previously commit inspected it a second time, and always by name, so the two could disagree.
- An inspect that finds nothing reports `image <ref> is not available locally`; it used to be dereferenced immediately. VERIFIED: with the base image built by the same process and its tag removed mid-run, that path is a SIGSEGV before this change.
- What lands in the committed image does not change: `Cmd`, `Entrypoint`, `User`, `WorkingDir` and the `LANG`/`LC_ALL` envs are still restored from the base image config, because the build container overrides them.

## Why

The build container is created from the base image with its entrypoint and command replaced by the stapel build script, so committing it would keep those overrides. werf restored the original config by inspecting the base image again at commit time — after the container had run, at a point where nothing guarantees the image is still there, and where a missing image was answered with a panic. The image config is immutable, so reading it before the run and reusing it removes the window instead of coping with it.

The alternative was to keep the second inspect and fall back to the image id when the name no longer resolves. That keeps a lookup in the commit path that can still fail for a reason nobody can act on, and leaves the two references (`run` by id, commit by name) diverging.

Fixes: 69140cdeff4e ("feat(buildah): refactor container-runtime interface and fix usage of runtime")

### DIFF
--- a/pkg/container_backend/legacy_base_image.go
+++ b/pkg/container_backend/legacy_base_image.go
@@ -1,8 +1,6 @@
 package container_backend
 
 import (
-	"context"
-	"fmt"
 	"sync"
 
 	"github.com/werf/werf/v2/pkg/image"
@@ -32,19 +30,6 @@ func (i *legacyBaseImage) Name() string {
 
 func (i *legacyBaseImage) SetName(name string) {
 	i.name = name
-}
-
-func (i *legacyBaseImage) MustResetInfo(ctx context.Context) error {
-	if info, err := i.ContainerBackend.GetImageInfo(ctx, i.Name(), GetImageInfoOpts{}); err != nil {
-		return fmt.Errorf("unable to get info for image %s: %w", i.Name(), err)
-	} else {
-		i.SetInfo(info)
-	}
-
-	if i.info == nil {
-		panic(fmt.Sprintf("runtime error: info must be set for image %q", i.name))
-	}
-	return nil
 }
 
 func (i *legacyBaseImage) GetInfo() *image.Info {
--- a/pkg/container_backend/legacy_stage_image.go
+++ b/pkg/container_backend/legacy_stage_image.go
@@ -207,14 +207,6 @@ func (i *LegacyStageImage) introspectBefore(ctx context.Context) error {
 	return nil
 }
 
-func (i *LegacyStageImage) MustResetInfo(ctx context.Context) error {
-	if i.buildImage != nil {
-		return i.buildImage.MustResetInfo(ctx)
-	} else {
-		return i.legacyBaseImage.MustResetInfo(ctx)
-	}
-}
-
 func (i *LegacyStageImage) GetInfo() *image.Info {
 	if i.buildImage != nil {
 		return i.buildImage.GetInfo()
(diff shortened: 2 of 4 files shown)


alexey-igrychev:
## Verification

- Mutation: made `prepareCommitOptions` merge an empty options set instead of the inherited one → `LegacyStageImageContainer commit options / restores the base image config read before the run` failed (67 passed, 1 failed); reverted → green.
- Not run: no stapel build with the Docker Server backend on macOS (it needs linux/amd64), so the pre-run inspect wiring — that `run` actually prepares the options the commit later uses — is unproven here; it is being exercised on a Linux host by removing the base image tag and id at both moments (before `docker run` and while the container runs) against binaries built with and without this change.

## Review focus

- Whether a stage container can ever be committed without going through `run` — both `Commit` call sites are inside `Build` after `container.run`, including the `IntrospectAfterError` path, but a new caller elsewhere would now hit `commit options inherited from the base image are not prepared`.

alexey-igrychev:
## Verification

Reproduced and re-checked on a Linux host (Ubuntu 22.04, Docker 29.1.3), stapel image `from: alpine:3.20`, `install: sleep 90`, `--repo localhost:5001/repro`, binaries built from this branch's base commit and from its head:

- Base image reused from the repo, `docker rmi -f <base stage tag>` while the stage container runs: base binary panics with `info must be set for image "<tag>"` — the production stack; head builds through (91.3 s).
- Base image built by the same process, same untag: base binary dies with SIGSEGV in `prepareInheritedCommitOptions` (the nil inspect result); head builds through (94.0 s).
- Base image removed before the container starts: head fails with the `docker run` error, no panic.
- `docker image prune -f` during the run does not reproduce anything — the base image is tagged and held by the container.
- Mutation: made `prepareCommitOptions` merge an empty options set instead of the inherited one → `LegacyStageImageContainer commit options / restores the base image config read before the run` failed (67 passed, 1 failed); reverted → green.

## Review focus

- Whether a stage container can ever be committed without going through `run` — both `Commit` call sites are inside `Build` after `container.run`, including the `IntrospectAfterError` path, but a new caller elsewhere would now hit `commit options inherited from the base image are not prepared`.

## Follow-up

- [ ] werf deletes its own freshly pulled stage image when a `--cache-repo` address equals the `--repo` address: `prepareCacheStageAsPrimary` renames the image onto itself with `removeOldName`, which is `docker tag X X` followed by `docker rmi X`. Separate PR.
- [ ] Stage images built by the current process are never locked — `StorageManager.LockStageImage` is called only from `FetchStageImage` — so the automatic host cleanup of a concurrent werf process removes them mid-build. Separate PR.

PR #7820, opened 2026-08-31
fix(build): ignore a cache repo pointing at the primary repo

alexey-igrychev:
## Summary

A `--cache-repo` whose address equals the `--repo` address made werf delete the stage image it had just pulled: promoting a stage fetched from a cache repo renames it onto the primary name and drops the old name, and with both addresses equal that is `docker tag X X` followed by `docker rmi X`. A single build recovered without a trace, because `docker run` pulls the missing image again; a build running next to it on the same container backend could lose the image while using it as the base image of the stage it was building, and die (#7819).

## What

- A cache repo whose address equals the primary repo address is ignored, with `WARNING: Ignoring cache repo <address>: same address as the primary repo.` printed once per command. Such a repo can cache nothing by definition.
- No stage image is untagged or deleted while promoting a fetched stage in that configuration, because the promotion never runs.
- Cache repos with any other address are unaffected: they are still fetched from and promoted as before.
- `--secondary-repo` and `--final-repo` are untouched; only the cache repo list is filtered.

## Why

`StorageManager.FetchStageImage` fetches a stage from every cache repo before the primary one and then promotes the fetched image to the primary name via `RenameImage(..., removeOldName: true)`. The name comes from `ConstructStageImageName`, which depends only on the repo address, the digest and the creation timestamp — so with the two addresses equal, the old and the new name are the same string, and removing the old name removes the image.

Guarding `RenameImage` against a rename onto itself was the alternative. It would have to be repeated per backend and would still leave werf fetching a "cache" that is the primary repo it fetches from anyway; dropping the repo removes the whole degenerate path in one place.

### DIFF
--- a/cmd/werf/common/common.go
+++ b/cmd/werf/common/common.go
@@ -928,7 +928,7 @@ func GetOptionalFinalStagesStorage(ctx context.Context, containerBackend contain
 	})
 }
 
-func GetCacheStagesStorageList(ctx context.Context, containerBackend container_backend.ContainerBackend, cmdData *CmdData) ([]storage.StagesStorage, error) {
+func GetCacheStagesStorageList(ctx context.Context, stagesStorage storage.StagesStorage, containerBackend container_backend.ContainerBackend, cmdData *CmdData) ([]storage.StagesStorage, error) {
 	var res []storage.StagesStorage
 
 	buildahMode, _, err := GetBuildahMode()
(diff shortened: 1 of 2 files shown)


alexey-igrychev:
## Verification

On a Linux host (Ubuntu 22.04, Docker 29.1.3), a stapel image built with `--repo localhost:5001/repro`, its stages pushed and the local images removed so the cache path runs:

- `origin/main` with `--cache-repo localhost:5001/repro` (same address): `Tagging image X by name X` → `Removing old image tag X` → `Untagged: …` → `Deleted: sha256:0a70f6eb9c95…` — werf deleted the stage image it had just pulled. The build survived only because `docker run` pulled it again.
- This branch, same command: `WARNING: Ignoring cache repo localhost:5001/repro: same address as the primary repo.`, no `Removing old image tag` and no `Deleted:` line anywhere in the debug log, build succeeded.
- Regression, `--repo localhost:5001/repro --cache-repo localhost:5002/repro` (different addresses) on this branch: no warning, `Fetching stage app/beforeInstall from localhost:5002/repro` still happens, build succeeded.

PR #7821, opened 2026-08-31
fix(build): keep host cleanup of another werf from deleting a stage in use

alexey-igrychev:
## Summary

Automatic host cleanup removes least recently used werf stage images and skips only those a running werf holds a host lock on — and that lock was taken for fetched stages only. A stage built by the current process stayed unlocked, so a cleanup forked by another werf on the same container backend could delete it while it was the base image of the stage being built; that build then failed on a base image that had vanished (#7819). Every `werf build` forks such a cleanup once the backend storage is over its allowed volume usage, so on a busy host with parallel builds this needs no unusual setup.

## What

- A stage image built by the current process is locked as soon as it gets its final name, so host cleanup run by any other werf process skips it, the same way it already skips fetched stages.
- The lock is held until the process exits, so a stage stays protected while the later stages of the same image are built. UNVERIFIED: a hand-run check on Linux — build a stage, run `werf host cleanup --force` from a second process, confirm the stage image survives — is in progress.
- Failing to take the lock fails the build with `unable to lock stage <stage> image <name>`, instead of silently building on an unprotected image.
- The list of held locks is guarded by a mutex; parallel image builds append to it concurrently, and only the fetch path did so before.
- Nothing changes for stages that are fetched rather than built: they were already locked at fetch time.

## Why

`StorageManager.LockStageImage` had a single caller, `FetchStageImage`, so the protection came for free only when the image arrived from a repo. A stage built locally is exactly as much in use — the next stage's build container is created from it — but nothing marked it, and `LocalBackendCleaner.removeImageByRepoTags` deletes any werf stage tag it finds unlocked.

The alternative was to make host cleanup avoid recently created images altogether. It already does that for dangling images, with a 15 minute window, and stretching the same idea to tagged stages trades a correctness signal for a heuristic: the lock states that a specific image is in use, an age filter guesses it.

### DIFF
--- a/pkg/build/build_phase.go
+++ b/pkg/build/build_phase.go
@@ -1172,6 +1172,12 @@ func (phase *BuildPhase) atomicBuildStageImage(ctx context.Context, img *image.I
 	stageImage.Image.SetName(newStageImageName)
 	phase.Conveyor.SetStageImage(stageImage)
 
+	// Hold the image against host cleanup of a concurrent werf process for as long as this build runs:
+	// the stage is going to be used as the base image of the next stage.
+	if err := phase.Conveyor.StorageManager.LockStageImage(ctx, newStageImageName); err != nil {
+		return fmt.Errorf("unable to lock stage %s image %s: %w", stg.LogDetailedName(), newStageImageName, err)
+	}
+
 	if err := logboek.Context(ctx).Default().LogProcess("Store stage into %s", phase.Conveyor.StorageManager.GetStagesStorage().String()).DoError(func() error {
 		if stg.IsMutable() {
 			prevBuiltImage := phase.StagesIterator.GetPrevBuiltImage(img, stg)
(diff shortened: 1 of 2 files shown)


alexey-igrychev:
Closing: the premise is wrong, and the experiment that was meant to confirm it disproved it instead.

A stage built by the current process does get locked, just not where I looked. `StorageManager.FetchStage` locks the image name *before* checking whether a fetch is needed (`storage_manager.go:427`, ahead of the `ShouldFetchImage` call), and it runs for the parent of every stage — built locally or not. So as soon as the next stage starts building, its parent is locked for the rest of the process.

Measured on a Linux host with a build in flight (stage container running `sleep 90`) and `werf host cleanup --dry-run --force` with the threshold set low enough to sweep every werf image, run from a second process:

- `origin/main`: the just-built parent image is **not** in the cleanup's deletion list; the other four werf images on the host are.
- this branch: same result, and the parent images of earlier builds — no longer held by any process — are listed by both.

So there is no observable difference, and the claim in the description does not hold. Whatever leaves the window open in werf/werf#7819 is narrower than "built stages are never locked" — at most the gap between a stage getting its final name and the next stage locking it, which needs its own evidence before it is worth a PR.

The mutex on `SharedHostImagesLocks` stands on its own: parallel image builds append to that slice concurrently. Splitting it off if it survives review on its own merits.

PR #7823, opened 2026-09-01
fix(build, buildah): stop "image not known" error in RUN --mount builds

Fral738:
## Summary

With the Buildah backend a staged Dockerfile build fails with `resolving mountpoints for container ...: <stage tag>: image not known` when a `RUN --mount=from=<stage>` instruction references a stage werf found up-to-date in the stages storage. It needs a pre-existing host state: the referenced stage image exists in the repo but not in the host's local containers storage (fresh runner, or host cleanup between runs):

```shell
# werf.yaml: staged: true Dockerfile image; Dockerfile:
#   FROM alpine AS os
#   RUN mkdir /apk && apk fetch --no-cache -o /apk make
#   FROM alpine
#   RUN --mount=type=bind,from=os,source=/apk,target=/apk ls /apk
werf build --repo $REPO        # everything built and pushed
buildah rmi --all --force      # local storage cleaned
# invalidate only the final RUN, commit
werf build --repo $REPO        # os is "previously built"; RUN fails: image not known
```

## What

- A Dockerfile stage whose `RUN --mount=from=...` source image is missing from local containers storage now pulls it by ref before running the instruction, instead of failing with `resolving mountpoints for container ...: image not known`. Covered end-to-end by the new `staged_dockerfile_run_mount` e2e spec (build to repo → remove project images from local storage → invalidate only the final stage → rebuild); also VERIFIED by hand on a Linux host against fresh `main` (see comment).
- An image already present locally is not re-pulled and the registry is not contacted for it; parallel stage builds referencing the same missing image pull it once — the loser of the pull lock re-checks local storage instead of pulling again.
- A locally present mount-source image that does not match the target platform fails the build with `local image ... has platform ..., but target platform is ...` — the same check base images already get.
- A failed pull fails the build with `unable to ensure local image for run mount ...`.
- Deliberately unchanged: `COPY --from` (its source container is created with pull-if-missing) and stapel `import`/`dependencies` (pulled via container creation) — only the run-mount path lacked the pull.

## Why

werf rewrites `--mount from=<stage>` to the stage's registry tag and drives buildah's `builder.Run` directly, bypassing buildah's imagebuildah executor that pre-mounts stage images for its own builds. Buildah resolves run-mount `from=` images against local containers storage only and never pulls them. #7669 added the missing-image fallback only to the base-image path (`FromCommand`), so a host where a previously built stage exists only in the repo kept failing on run mounts.

### DIFF
--- a/pkg/container_backend/buildah_backend.go
+++ b/pkg/container_backend/buildah_backend.go
@@ -724,8 +724,65 @@ func normalizeDependencyImportDestination(absFrom, absTo string) (string, error)
 	}
 }
 
+// ensureImageLocally makes ref resolvable by buildah operations that only look up
+// the local containers storage (e.g. RUN --mount from=image resolution).
+func (backend *BuildahBackend) ensureImageLocally(ctx context.Context, ref string, opts CommonOpts) error {
+	inspect, err := backend.buildah.Inspect(ctx, ref)
+	if err != nil {
+		return fmt.Errorf("unable to inspect image %q: %w", ref, err)
+	}
+	if inspect != nil {
+		if opts.TargetPlatform != "" && !platformMatches(inspect, opts.TargetPlatform) {
+			return fmt.Errorf("local image %q has platform %s/%s, but target platform is %q; pull the correct image first", ref, inspect.OCIv1.OS, inspect.OCIv1.Architecture, opts.TargetPlatform)
+		}
+		return nil
+	}
+
+	logboek.Context(ctx).Debug().LogF("Image %q not found locally, pulling\n", ref)
+
+	mu := backend.getPullMutex(ref)
+	mu.Lock()
+	defer mu.Unlock()
+
+	imageID, err := backend.buildah.Pull(ctx, ref, buildah.PullOpts(backend.getBuildahCommonOpts(ctx, true, nil, opts.TargetPlatform)))
+	if err != nil {
+		return fmt.Errorf("unable to pull image %q: %w", ref, err)
+	}
+	if opts.TargetPlatform != "" && imageID != "" {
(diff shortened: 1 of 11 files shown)


Fral738:
## Verification

- Reproduced on Ubuntu 22.04 x86_64, rootless Buildah backend, local `registry:2`: staged-Dockerfile project built and pushed, `buildah rmi --all --force`, final RUN invalidated, rebuild — fresh `main` (17f71c55c) fails with `resolving mountpoints ...: image not known`; this branch pulls the stage tag and the same rebuild succeeds.
- New e2e spec run on the same host: `task test:e2e paths="./test/e2e/build/..." labelFilter="staged_dockerfile_run_mount"` — 2/2 passed (rootless and chroot isolation).
- Mutations run: skipping the pull for a missing image → 2 specs failed; disabling the platform mismatch check → "fails without pulling when the local from-image platform mismatches" failed; removing the under-lock re-check → "does not pull again when a concurrent caller already pulled" failed; deleting the `ensureRunMountImages` call from `BuildDockerfileStage` → "BuildDockerfileStage pulls the missing run mount image before applying instructions" failed.
- E2e mutation: backend implementation reverted to `main` with the new e2e spec in place → both entries failed with `resolving mountpoints ...: image not known`; restored → both passed.

## Review focus

- `ensureRunMountImages` pulls for every mount carrying `from=`, cache mounts included — confirm pre-pulling a cache-mount source image is acceptable.

## Follow-up

- [ ] Investigate why `werf host purge` leaves images in the rootless containers storage (observed during verification; separate issue in `werf/werf`).

reyreavman (approved):

PR #7827, opened 2026-09-01
fix(build): include build cache version in anchor digest

nervgh:
Ensure stage and anchor digests invalidate when the cache version changes,
rather than relying on a hardcoded global value or omitting it from anchor
inputs.


alexey-igrychev (approved):
Reviewed. The duplicate anchor digest test was removed in bb9d971d8; targeted unit tests pass.

PR #7830, opened 2026-09-01
fix(deploy): apply .helmignore when reading the chart

### LINKED ISSUE #6747: [giterminism] Read the helm chart taking into account the .helmignore file

dmmordvi:
## Summary

werf read the chart with its own file reader, which knew nothing about `.helmignore`, so every file of the chart directory reached the render and the published bundle: on the same chart werf and helm disagreed. The marker `// TODO helmignore support` sat in `FileReader.loadChartDir`, while two documentation pages already described the rules as working.

Repro:

```
# werf.yaml
project: helmignore-probe
configVersion: 1
```

```
# .helm/.helmignore
templates/ignored.yaml
```

`.helm/templates/ignored.yaml` and `.helm/templates/kept.yaml` each render a ConfigMap. `werf render` emitted both; only `kept` is correct.

Fixes #6747.

### Breaking

- BREAKING: a chart carrying a `.helmignore` now loses the files it lists, from the rendered manifests and from the published bundle alike. The way out is editing `.helmignore`; a release note is the only mitigation.
- BREAKING: helm's default rules apply even to a chart with no `.helmignore`, so a dotfile directly under `templates/` (e.g. `templates/.gitkeep`) is dropped. Chosen for full helm compatibility over a narrower blast radius.
- Both are stated in the v3 migration guide (`docs/pages_{en,ru}/resources/migration_from_v2_to_v3.md`), which this PR updates and which is the authority for what an upgrading project has to check.

### Chart contents

- Rules are read through `github.com/werf/nelm/pkg/helm/pkg/ignore`, so the file set matches helm for every rule class, inside vendored subcharts, and for files delivered through werf includes. Syntax and the deviations from `.gitignore` are documented in `docs/pages_{en,ru}/usage/distribute/bundles.md`, the default rules in `docs/pages_{en,ru}/usage/deploy/charts.md`; this PR updates both.
  VERIFIED: file set compared against helm 3.20.1 on charts exercising every rule class and a vendored subchart with its own `.helmignore`, in both giterminism modes.
- An ignored path is excluded before it is read or checked against giterminism, so `.helmignore` can exclude an uncommitted file — as `docs/pages_en/usage/project_configuration/giterminism.md` already advises.
- A dependent chart pulled in with `dependencies[].repository: file://` is loaded as a separate chart, so its own `.helmignore` filters it and the parent chart's rules do not reach it, while files delivered into that directory through werf includes are filtered by its rules too. This follows helm and was neither documented nor tested before; `docs/pages_{en,ru}/usage/deploy/charts.md` now states it for all three ways of including a chart, and `pkg/giterminism_manager/file_manager/load_chart_dependencies_test.go` covers it.
- A chart left with no files reports `the directory ".helm" not found in the project git repository or includes`, the same message as an absent chart, and appends `; the chart has a .helmignore, check whether its rules exclude every file` when one is in force. helm does not distinguish the two cases either.
- A `**` rule aborts the command before rendering with `parse ".helm/.helmignore": double-star (**) syntax is not supported`, as in helm.
- The two e2e assertions covering the excluded ConfigMap had been passing against a name that never existed (`not-deployed-because-in-helm-ignore`, one hyphen too many), so they stayed green while the file was in fact deployed; corrected, they now cover this change end to end.
- Excluding a symlink werf cannot resolve — a loop, or a target that leaves the project — now works with a plain rule (`ignoreddir`, `loop*`) in both giterminism modes, where the chart failed to load before; a trailing-slash rule (`ignoreddir/`) still fails in both, because telling a directory from a file requires resolving the symlink, which is the operation being avoided. helm renders every one of these charts.
  VERIFIED: A/B against a binary built from `3`, where each of these charts fails with and without a rule; pinned by specs in `pkg/giterminism_manager/file_reader/helm_chart_extender_load_test.go` (filesystem) and `test/legacy_e2e/suites/giterminism/helm_test.go` (commit).

## Why

The filtering had to go into `FileManager.LoadChartDir`, not only `FileReader`: `FileManager` is the chart reader the commands register, and it merges the local chart directory with the files delivered through werf includes. Filtering only the local source left the imported half unfiltered and, because an excluded path was indistinguishable from an absent one, let an include add an excluded local path back with different content.

The rejected alternative was matching in the walk callback instead of inside the walk: it reads simpler, but an ignored path would still be resolved and checked against giterminism first, which breaks excluding uncommitted files and makes a symlink inside an ignored directory behave differently depending on `--loose-giterminism`.


dmmordvi:
## Verification

- File set compared against helm 3.20.1 with a locally built binary, on charts exercising every rule class, a vendored subchart with its own `.helmignore`, and a `templates/` that is a symlink — in the default and `--loose-giterminism` modes.
- Project-boundary symlink cases were A/B'd against a binary built from `3` (`708976aab`): a symlink escaping the project fails identically with and without this branch, whether or not a rule covers it, so the remaining gap against helm is pre-existing.
- The empty-chart message was driven through the CLI in both branches, not just the unit specs: `.helmignore` = `*` appends the `.helmignore` hint, while a chart emptied by helm's own defaults reports the base message with no hint.
- Excluding an uncommitted file was checked under the default (enforced) giterminism with an unparseable `templates/uncommitted.yaml` in the tree.
- Mutation: mis-setting the include's `FallbackExists` → the `included chart emptied by helm's defaults` spec fails; dropping the parent-directory loop, the walk predicate, the include-loop filter, local-over-include precedence or helm's defaults → the corresponding suites fail; removing the nil guard in `matchChartIgnoreRules` → the File Manager suite panics.
- Mutation: dropping the parent-directory check from the commit symlink skip → the enforced-mode `symlink loop inside an ignored directory` spec and both legacy e2e specs for it fail with `too many levels of symbolic links`; restoring the old `not found` message → six File Manager specs fail; letting an include shadow a local file → the `keeps the local file` spec fails.
- Mutation: `&&` → `||` in the commit symlink skip → the enforced-mode `keeps a symlink to a file whose name only a directory rule matches` spec and the legacy e2e `keeps a committed symlink` spec fail. An earlier revision of this comment called that mutation unobservable; it is not — in a clean checkout the filesystem walk skips committed files and the commit list is the only source, and the mutated binary dropped a committed symlink from the render.
- Mutation: honoring the readdir error before the skip predicates (the old walk order) → `does not read a directory matched by .helmignore` fails; dropping the deferred readdir error altogether → `fails on an unreadable directory no rule excludes` fails, the chart would otherwise load partially; calling the include fallback eagerly → `does not read the .helmignore of an include when the chart has a local one` fails.
- The mode 000 excluded directory was driven through the CLI: before the fix `--loose-giterminism` failed with `permission denied` and the enforced mode passed; after it both pass.
- Mutation: passing nil instead of the symlink skip predicate into the recursive commit walk → the legacy e2e `symlink loop below a symlinked chart directory` spec fails with `too many levels of symbolic links` in both project layouts, and so does the real-repository unit spec of the same name in the File Manager suite; ignoring the predicate in the walk, or never matching the symlink itself, fails both real-repository specs there. The mocked walks never resolve a symlink, so only these specs and the legacy e2e see this code.
- Mutation: deleting the processed-set write in the include loop → the `takes a path from the first include that delivers it` spec fails with a third file in the result.
- Mutation: dropping the parent-directory check from the flat commit list → the enforced `keeps only what a leading ! pattern matches` spec fails along with the directory-rule specs; `templates/kept.yaml` matches `!*.yaml` as a file, so only the pruned parent excludes it.
- CI: `integration_main` fails on `cleanup/host_purge_test.go:47` on `3` as well (run 35725050434 on `98bb0cbaf`), unrelated to this branch.
- Not run locally: e2e (`test/e2e`) and integration suites other than `legacy_e2e/suites/giterminism` need Linux with Docker and kind, unavailable on the host used here; the giterminism suite runs on macOS and passed. CI covers the rest — `e2e_complex` ran the corrected converge and bundle-publish-apply assertions and passed.

## Review focus

- The two filtering paths stay coupled by hand: the walk prunes directories in the loose mode, while the enforced mode filters a flat commit list where a `mustDir` rule can never match a file. The parent-directory walk is shared now, but a new rule class added to one path still needs its counterpart in the other.
- `FileReader.LocateChart` is required by nelm's `ChartFileReaderer` but never called by nelm or werf; it is reduced to a directory check, the spec that drove it is removed. `FileReader.LoadChartDir` has no production caller either — the commands register `FileManager` — and the `file_reader` suite drives it; its core, `loadChartDir` and `readChartIgnoreRules`, is what `FileManager.LoadChartDir` calls, so the coverage is of the production path.
- `ChartIgnoreRules.IsFileIgnored` takes a `ctx` it does not use; kept for the public-method convention.
- The changed walk order in `fs.go` reaches every `ListFilesWithGlob` caller, not only the chart read; no spec covers a non-chart consumer with an unreadable directory.
- The enforced-mode specs accept every path for reading from the filesystem (`IsDirOrSubmodulePathMatched` → true), which hides whatever the commit walk drops; only the `committed with no local changes` context models a clean checkout. The mock commit walk also does not descend into a symlinked directory; the `templates/` symlink case is covered by the legacy e2e suite only.
- No warning is emitted when `.helmignore` excludes files, and no switch turns the filtering off: a per-run warning would fire for every chart with a deliberate `.helmignore`, and a switch would restore the divergence from helm this PR removes. The `werf plan` step in the migration guide is the mitigation.
- `pkg/helm/pkg/ignore` is imported directly rather than through nelm's `pkg/export/helm`, which carries no `ignore` package; a facade for it is not planned, so this import stays as is. The pin is what holds it: a nelm bump that reshuffles that path breaks the build, and nothing there owes us compatibility.
- Confirm both BREAKING claims are acceptable as a product call, in particular dropping dotfiles under `templates/` in charts that carry no `.helmignore`.

## Follow-up

- [ ] Release note pointing at the migration guide entry — the only mitigation for a chart relying on the old behavior; the guide itself carries all four breaking claims.
- [ ] With an absolute chart directory — `werf chart ts build` passes one to nelm — werf includes match no chart file at all, `.helmignore` included; the include keys are project-relative. Pre-existing and out of scope here; normalizing the directory in `FileManager.LoadChartDir` would switch includes on for that command.
- [x] `task mock:check` fails on `3` itself (`test/mock/locker.go`, `file_reader_deps_mock_test.go` import order), unrelated to this branch; this branch regenerates `test/mock/git_repo.go`, which also failed there, so it is strictly better than the base. Nothing left to do here.

alexey-igrychev (commented):
# Review

**Base:** `3` (merge-base `708976aab`) · **Head:** `777eb9cfe` · 21 files, +1505/-41

Built and ran locally: `task build`, `task test:unit` (whole repo), `task lint:golangci-lint` on the touched packages — all green. `pkg/ssh_agent` failed once on a race for `/tmp/werf-test-agent` against a parallel run; it passes on its own, and on the base too. `task mock:check` fails identically on the base and on the branch (`test/mock/locker.go`, import order) — pre-existing, as the author's checklist already notes; the new `test/mock/includes_repo.go` is in sync (regenerating produces an empty diff).

I mutation-tested the load-bearing specs — the suite catches: dropping the parent-directory loop in `skipRelativeToDirPathWithParents`, matching the resolved path instead of the not-resolved one in `skipConfigurationPathFunc`, letting the include-supplied `.helmignore` win over the local one, dropping `rules.AddDefaults()`, and making `HasIgnoreFile()` always return `true`. One mutation survived — see Minor.

Splitting `FallbackExists` from `FallbackData`, dropping `chartHadExcludedFiles` with its second chart read that discarded the user's rules, and shaping the `ListCommitFilesWithGlob` mock after the real `WalkCommitFiles` are exactly the places where changes like this usually leak. What follows is what's left.

---

## Major 1 — `.helmignore` does not prevent symlink resolution

`pkg/giterminism_manager/file_reader/configuration.go:137-145`

`skipConfigurationPathFunc` calls `IsDirectoryExist(ctx, existingRelPath)` **before** matching the rules. `IsDirectoryExist` → `ResolveFilePath`, so an excluded path gets resolved before it is excluded.

Reproduced with a binary built from `777eb9cfe`, in both giterminism modes:

```
.helm/.helmignore:  loop
.helm/loop  -> loop2
.helm/loop2 -> loop
```

```
# --loose-giterminism
Error: ... unable to load chart directory: check ".helm/loop" is a directory:
       unable to resolve file path ".helm/loop": too many levels of symbolic links

# enforced (default)
Error: ... unable to resolve commit file ".helm/loop": too many levels of symbolic links
```

This directly contradicts the PR description: "Excluding a symlink werf cannot resolve — one whose target leaves the project — now works with a plain rule (`ignoreddir`)". For a symlink escaping the project it does work (`resolveFilePath` returns `FileNotFoundInProjectDirectoryError`, which `isFileExist` swallows into `false, nil`); for a loop it does not. This is not a regression against `3`, but the claimed property does not hold, and the description ships verbatim as the squash commit body.

The fix is cheap: match `skipRelativeToDirPathFunc(relativeToDirPath, false)` first, and only call `IsDirectoryExist` when nothing matched and a directory rule still needs deciding. The rule `loop` (no slash, not `mustDir`) matches immediately and the path is never resolved.

Two remainders this does not close, worth naming explicitly:

- `ignoreddir/` (trailing slash) on an unresolvable symlink — a `mustDir` rule cannot be applied without `IsDir()`. Already described in the PR.
- In enforced mode `ListCommitFilesWithGlob` (`configuration.go:47`) resolves the whole commit tree **before** `lo.Reject` — the predicate is not threaded into it at all. Invisible for ordinary paths (no read happens, and a spec asserts that), but it is the same ordering gap.

## Major 2 — `file://` dependencies from `Chart.lock` are filtered by the opposite policy

`pkg/giterminism_manager/file_manager/file_manager.go:356`

`nelm/pkg/helm/pkg/chart/loader/load_dependencies.go:141` calls `loadChartDirFunc(ctx, localChartPath)` = `FileManager.LoadChartDir` **separately** for every `file://` dependency in `Chart.lock`. So `readChartIgnoreRules` reads the **subchart's** `.helmignore`, and the root rules never reach it.

A/B with binaries from the base and the branch. A chart with `file://../sub`, root `.helmignore` = `secret.yaml`, subchart's own `sub/.helmignore` = `templates/own.yaml`:

```
BASE: root-kept  root-secret  sub-kept  sub-own  sub-secret
PR:   root-kept               sub-kept           sub-secret
```

The root basename rule `secret.yaml` removes `root-secret` but does not reach `sub-secret`. Meanwhile `sub-own` is gone — the subchart is filtered by its own `.helmignore` and helm's defaults for the first time.

Yet the spec `applies only the root rules to a vendored subchart, the way helm does` (`helm_chart_extender_load_test.go:130`) pins the **exact opposite** policy for `charts/sub/`: root rules reach in, the subchart's own `.helmignore` is inert. The same logical subchart behaves oppositely depending on how it is attached. No decision, no test, not a line in the description — and helm is not the reference here, since live `file://` loading is werf/nelm-only, which is precisely why the choice has to be made deliberately.

This is also a third breaking change: `sub-own` disappears for anyone keeping a `.helmignore` in a `file://` subchart.

## Major 3 — the migration guide is not updated

`docs/pages_{en,ru}/resources/migration_from_v2_to_v3.md` is the explicit v3 breaking-change list the repository maintains in both languages. The PR declares two BREAKING changes and does not touch that file. Two more are not declared at all:

- `**` in `.helmignore` now aborts the command (`parse ".helm/.helmignore": double-star (**) syntax is not supported`), where the whole file used to be inert. The description files this under "Chart contents", not "Breaking".
- The `file://` dependencies from Major 2.

One more upgrade trap worth naming in the guide: a leading `!`. In helm a negate pattern means "ignore everything that does NOT match" (`ignore/rules.go:112-120`), so a user who wrote `!templates/keep.yaml` expecting gitignore semantics ends up with an empty chart. Helm parity is preserved and the `.helmignore` hint will fire, but this is hard to work out on your own.

The repository history already carries a commit titled "The v3 migration guide missed two user-facing breaking changes" — this rake is a known one.

---

## Minor

- **`file_manager.go:389`** — `processed[normToPath] = false` in the exclusion branch is dead: deleting it leaves the suite green, because `rules.IsFileIgnored` runs again for any later include hitting the same path. The commit message presents it as load-bearing.
- **`file_manager.go:430`** — `tryReadFromIncludes` for `.helmignore` swallows the error wholesale, so a genuine read failure from an include is indistinguishable from "no such file" and the rules silently do not apply. Stylistically this matches `ReadDockerfile` next to it, but the cost is higher here: the chart's contents change silently.
- **`load_chart_dir_test.go:96`** `keeps the local file when the same path comes from an include` — the `for … if f.Name == X { Expect(…) }` loop passes vacuously when the file is absent from the result entirely. The enforced twin (`HaveLen(3)` + `HaveKeyWithValue`) saves the property, so there is no coverage gap, but this spec on its own proves nothing.
- **`helm_chart_extender_test.go`** — `Expect(chartIgnoreFileInfo{…}.Name()).To(Equal("ignored.yaml"))` asserts implementation trivia; `ignore.Rules` never reads `Name()`, as the comment right there says.
- **`helm_chart_extender.go:69`** — `FileReader.LoadChartDir` is left with no production callers (everything goes through `FileManager` → `LoadChartDirWithIgnoreRules`); only the `nelmcommon.ChartFileReaderer` contract holds it. And `FileManager.LocateChart` → `FileReader.locateChart` now also reads `.helmignore` and walks the entire chart just to evaluate `len(files) == 0` — a second full walk on every command. Both are pre-existing in nature, but this PR doubles the cost.
- **`configuration.go:139`** — `IsDirectoryExist` on every walked path adds an `Lstat` cascade per file. Negligible for charts, and it goes away by itself once it is moved off the hot path per Major 1.

---

Converting to draft on Major 1–3. I checked the e2e fix (`not-deployed-because-in-helmignore`) — the name in `templates/ignored.yaml` really is that, so the assertions were passing vacuously before.

dmmordvi:
## Major 2 — not a policy conflict, it is helm's own split

Both halves of the divergence are helm's:

- A `file://` dependency is loaded **as its own chart**, so its own `.helmignore` plus helm's defaults filter it and the parent's rules never reach it: `downloader/manager.go:889-900` (`tarFromLocalDir` → `loader.LoadDir`) → `chart/v2/loader/directory.go:84-101`, which parses the dependency's `.helmignore` and calls `AddDefaults()`. nelm does the same live, per dependency: `chart/loader/load_dependencies.go:134-152`.
- A chart sitting in `charts/` is collected by the **parent's** walk, so the parent's rules filter it and a nested `.helmignore` is inert — the policy `helm_chart_extender_load_test.go:130` pins.

So the same logical subchart really does behave differently depending on how it is attached, in helm too. What was missing here was not a decision but documentation and a test; both are in now:

- `docs/pages_{en,ru}/usage/deploy/charts.md` states all three cases (`charts/`, `file://`, packaged from an OCI/HTTP repository), and the v3 migration guide entry points there.
- `pkg/giterminism_manager/file_manager/load_chart_dependencies_test.go` covers a `file://` dependency: its own `.helmignore` applies, the parent's basename rule does not reach it, and files delivered into it through werf includes are filtered by its rules too.

One divergence stays, and it is about timing, not policy: helm materializes a `file://` dependency as `charts/*.tgz` at `dependency update` time and never re-evaluates `.helmignore` when loading an archive, while werf reads the directory on every command. Editing the dependency's `.helmignore` therefore takes effect on the next werf command, but only on the next `helm dependency update`.

The third breaking change is real and stands: a `.helmignore` kept inside a `file://` subchart starts excluding files. It is covered by the migration guide entry, which sends the reader to the chart reference for which `.helmignore` filters what.

dmmordvi:
Replying to the review on `777eb9cfe`. Everything below is on the branch now.

**Major 1 — `.helmignore` did not prevent symlink resolution.** Fixed in `a68d1c973`: the rules are matched against the unresolved path first, and `IsDirectoryExist` runs only when the file and directory verdicts differ. Both remainders you named are closed too: the enforced-mode commit walk now takes the skip predicate (`git_repo.ListCommitFilesWithGlobOptions.SkipSymlinkPathFunc`) and drops a symlink before resolving it, and `724ba600b` extends that to the parent directories, so `ignoreddir/` on a loop inside the directory works in both modes. The `mustDir` rule on the unresolvable symlink itself (`loop/`) still fails, as in helm.

**Major 2 — `file://` dependencies.** Answered in the comment above: both halves are helm's own split, and the branch documents it (`charts.md`, all three ways of including a chart) and covers it (`load_chart_dependencies_test.go`).

**Major 3 — migration guide.** `7d426df69`, `b497568c1` and `464f96ed6`: the v3 guide entry, in both locales, names the four breaking claims (`.helmignore` applied, helm's defaults, `**` erroring, the leading `!`), says the filtering is silent and asks for `werf plan` before the first `werf converge`. The PR description lists the same four under `### Breaking`.

**Minor.** `processed[...]` is a set now (`446c5b2b5`); `tryReadFromIncludes` is gone, an include's `.helmignore` is read lazily and only its absence is tolerated (`2c3e7305f`, `9268258eb`); the vacuous local-wins spec asserts the key with its value (`b67cbed84`); the `Name()` assertion is dropped (`866de66af`); `LocateChart` is reduced to a directory check and its spec removed (`c0828b6ca`, `b67cbed84`); the `Lstat` cascade went away with Major 1. `FileReader.LoadChartDir` stays for the `ChartFileReaderer` contract; the `file_reader` suite drives it, and its core is the production path through `FileManager`.

dmmordvi:
## Review

**Base:** `3` (merge-base `b02d79dcc`) · **Head reviewed:** `99be1721e`, plus the two local follow-up commits `8eb9f6751` and `dbd0f4156` (not pushed yet) · 29 files, +2053/-91

Run by an agent on the author's own machine, so this is not an independent sign-off: the pass inherits the design assumptions of the branch. To offset that, the implementation, the specs and the docs were each read by a separate cross-family reviewer, and the user-visible behaviour was driven through a binary built from the branch.

**Verdict.** No critical findings. The filtering itself holds up: exclusion happens before the read, before the giterminism check, before a symlink is resolved and before an unreadable directory is listed, in both giterminism modes. What was left were two gaps — one test, one docs — both now fixed and re-verified, and a handful of minors.

### Checks

- `task build`; `task lint:golangci-lint` on `./pkg/giterminism_manager/... ./pkg/git_repo/... ./pkg/includes/...` — `0 issues`.
- `task test:unit` on the same paths — green.
- `task mock:check` fails on `test/mock/locker.go` and `pkg/giterminism_manager/file_reader/file_reader_deps_mock_test.go`, both outside the diff (`gci` vs. raw `mockgen` import grouping). The mocks this branch touches — `test/mock/git_repo.go`, `test/mock/includes_repo.go` — are in sync.
- A whole-repo `task test:unit` fails to compile `pkg/true_git` with `ambiguous import: google.golang.org/grpc/stats/opentelemetry`. That is a local `go.work` artifact, not this branch: with `GOWORK=off`, and in a worktree of the merge-base, it passes.
- CLI probes against the built binary, enforced and `--loose-giterminism`: a file listed in `.helmignore` disappears from the render; an excluded file that is uncommitted and unparseable does not fail the command; `loop*` and `ignoreddir/` make an unresolvable symlink loop loadable while the same chart without a rule fails; a `chmod 000` directory under a rule loads while the same directory without a rule still fails with `permission denied`; `.helmignore` = `*` yields `no chart files found in ".helm"; the chart has a .helmignore, …`; a chart emptied by helm's defaults yields the same message without the hint; a `**` rule aborts with `double-star (**) syntax is not supported`; an absent directory is still rejected earlier with `path ".helm" not found on local filesystem or includes`; `--log-debug` shows `--- templates/ignored.yaml excluded by .helmignore`.
- 22 mutations of the implementation: 18 killed by named specs, 4 survived — see below.

### Fixed during the review

- **The deferred readdir error had no coverage.** `fs.go:162-164` is what keeps a genuine `permission denied` fatal now that the skip predicates run first; replacing `return err` with `return nil` left the entire suite green, i.e. the branch introduced a silent partial-chart path with nothing pinning it. Fixed in `8eb9f6751`: the mutation is now killed by exactly one spec, `LoadChartDir [It] fails on an unreadable directory no rule excludes`, with the other 70 still passing. The spec is discriminating — the `.helmignore` in the fixture cannot match the unreadable directory, the assertion is on `permission denied`, and there is a root guard plus a `DeferCleanup` chmod.
- **`charts.md:140` claimed something the code does not do.** "A rule with a trailing slash … does not exclude a symlink werf cannot resolve" — in fact the command aborts: with `.helmignore` = `loop/` the load fails with `too many levels of symbolic links` in both modes, because the file/dir verdicts differ and the path is resolved to decide. Fixed in `dbd0f4156` by dropping the bullet in both locales rather than expanding it: it described werf's resolution internals in a list that otherwise documents helm's rule syntax, and the situation it covers is a broken chart met once, with the failing file named in the error. Removal is symmetric across locales and leaves no dangling reference.

### Remaining findings

**Minor — `pkg/git_repo/base.go:1029` and the recursive hand-off at `:1039` have no unit coverage.** Removing the `skipSymlinkPathFunc` check, or failing to pass it into the recursive call, both survive the unit suite: each enforced-mode suite reimplements the walk inside a gomock `DoAndReturn` (`helm_chart_extender_load_test.go:304-331`; `load_chart_dir_enforced_test.go:91-118` ignores `opts` altogether). Only `test/legacy_e2e/suites/giterminism/helm_test.go:277-345` exercises the real code, and it skips on Windows. The fixtures in `helpers_test.go` already drive real git under enforced giterminism, so a spec there is cheap.

**Minor — `git.go:195`, the direct-match branch of the commit symlink skip.** Replacing `skip(path, false) && skip(path, true)` with `false` survives the unit suite; the enforced context has no counterpart to the loose-mode spec `excludes a symlink loop matched by .helmignore without resolving it`. Covered by legacy e2e only.

**Minor — `file_manager.go:430`, the include `.helmignore` lookup breaks on an absolute chart dir.** `normDir` keeps an absolute `dir`, while include keys are project-relative, so the fallback never matches. Reachable through `cmd/werf/chart/ts/build/build.go:96`, which builds an absolute path that nelm's loader hands to `LoadChartDir`. The blast radius is contained by a pre-existing property — with an absolute `dir` the include prefix check at `:382` matches nothing either, so no included file reaches the chart in that command at all — but normalizing the path via `chartLocalAbsPath` would keep the two from drifting further apart.

**Minor — `LocateChart` is now filesystem-only** (`helm_chart_extender.go:26-39`), so a chart directory that exists only in the commit answers "not found". Harmless today: nelm declares the method (`nelm/pkg/common/legacy.go:30`) and never calls it, and `FileManager.LocateChart:295-306` has an includes fallback. No spec pins either behaviour, so if nelm ever starts calling it the divergence from `LoadChartDir` surfaces silently.

**Minor — the empty-chart message gives no hint when helm's defaults are what emptied the chart.** A chart whose only file is `templates/.gitkeep` reports `no chart files found in ".helm"` while the directory is full of files. This is the deliberate trade in `c85f90ca` (no second read of the chart), but the hint could name the default rule unconditionally when the chart has no `.helmignore`.

**Minor — `ChartIgnoreRules.IsFileIgnored(ctx, …)` does not use its `ctx`** while the neighbouring `HasIgnoreFile()` takes none (`helm_chart_extender.go:138,145`). Two readings of the same convention inside one type.

**Minor — docs.** `migration_from_v2_to_v3.md:30` (both locales) links to the charts page as a whole although the section now has an anchor. `charts.md:146` states the `file://` rule unconditionally, but a dependency whose directory sits under the parent's `charts/` is collected by the parent walk under the parent's rules — worth the half-clause. `charts.md:130` says the rule format is "the same as in .gitignore" (text carried over from the bundles page), while helm matches through `filepath.Match`; "similar to" would be accurate. The section has no example of a `.helmignore`, which is the first thing a reader wants.

**Minor — the PR description.** Line 48 lists the rules that now work on an unresolvable symlink (`loop*`, `ignoreddir`, `ignoreddir/`) but never says that `loop/` — a directory rule on the link itself — still fails the load. Since the description becomes the squash commit body and the docs bullet is gone, this is now the only place that fact could live.

### Risks

| № | Risk | Type | Likelihood | Severity | Location | Consequences | Recommendation |
| :--- | :--- | :--- | :--- | :--- | :--- | :--- | :--- |
| 1 | A chart with a previously inert `.helmignore` silently loses resources after the upgrade | UX/Product | Likely | High | `file_manager.go:361` | `converge` deletes resources; no warning | Accepted: the migration entry and the `werf plan` step are the mitigation |
| 2 | An unreadable directory is skipped in non-chart configuration reads | Operational | Possible | Medium | `fs.go:133-135,162-164` | The shared walk also serves `werf.yaml` templates, Dockerfiles and includes; a directory outside the accepted matcher is now skipped instead of failing | Covered for the chart by `8eb9f6751`; no spec drives a non-chart consumer |
| 3 | A regression in the commit-walk symlink skip is invisible to the unit suite | Technical | Possible | Medium | `base.go:1029,1039` | A chart with an excluded unresolvable symlink fails again | Add the real-git spec described above |
| 4 | Filtering diverges in `werf chart ts build` | Technical | Unlikely | Low | `file_manager.go:430` | Local files filtered by helm's defaults only | Normalize the chart dir to a project-relative path |

### Not verified

- `test/e2e` and `test/legacy_e2e` beyond compilation — they need Linux with Docker, kind and a registry; CI covers them.
- Windows: every new symlink spec carries `skipOnWindows`.
- Performance on a large chart — the rule predicate is asked more than once per path, as the description notes.

dmmordvi:
Replying to the review on `99be1721e`. The two fixes made during the review are `8eb9f6751` and `dbd0f4156`; the remaining findings, in order:

- **`base.go:1029`, the recursive hand-off at `:1039`, and the direct-match branch at `git.go:195`** — covered at unit level now by `e903587aa`: two specs load a committed chart with a symlink loop through a real repository under the default giterminism, a loop at the chart root under `loop*` and a loop below `templates/` symlinked out of the chart under `templates/loop*`. Ignoring the predicate in the walk and never matching the link itself fail both specs; dropping the predicate from the recursive call fails the second. The Verification comment lists them.
- **Include `.helmignore` lookup on an absolute chart dir** — pre-existing class: with an absolute directory werf includes match no chart file at all, not only `.helmignore`, and `werf chart ts build` is the one command that passes one. Normalizing the directory would switch includes on for that command, which is a separate change; recorded in Follow-up.
- **`LocateChart` filesystem-only** — nelm declares the method and never calls it, werf neither; kept as the cheapest correct implementation of the contract. Noted in Review focus.
- **No hint when helm's defaults emptied the chart** — kept the plain message on purpose: a hint about `templates/.?*` would also fire on a genuinely empty directory, where it sends the reader after a rule that did not act. The default rule is stated in the migration entry and on the charts page.
- **`IsFileIgnored(ctx, …)`** — public method, context first, per CODESTYLE; `HasIgnoreFile` is a plain accessor. Kept.
- **Docs** — the migration entry now links the anchored subsection and the Russian clause is a sentence, `d767376ee`. The `file://` rule stays unqualified: a dependency whose directory sits under the parent's `charts/` is also a vendored subchart, helm 3.20.1 renders the same file, and the layout is not one werf supports; the PR description says so.
- **PR description on `loop/`** — added: a directory rule on the link itself still fails the load, since telling a directory from a file takes resolving it.

PR #7831, opened 2026-09-01
fix(bundle): drop the publisher's release channel from published bundles

dmmordvi:
## Summary

`werf bundle publish` no longer stores `werf.io/release-channel` in the bundle. Previously the channel of the publishing werf — exported by `werf ci-env` as `WERF_ADD_ANNOTATION_WERF_RELEASE_CHANNEL` from `TRDL_USE_WERF_GROUP_CHANNEL` — was written into the bundle's `extra_annotations.json` and then stamped on every resource deployed from that bundle, describing a binary that had no part in the deployment.

## What

- BREAKING: resources deployed from a bundle published by this version carry no `werf.io/release-channel` unless the applying werf is given one; anyone selecting or reporting on that annotation over bundle-deployed resources sees it disappear. The way out is passing `--add-annotation=werf.io/release-channel=<channel>` (or letting `werf ci-env` export it) on the apply side, where it names the werf that actually deploys.
- When `--add-annotation=werf.io/release-channel=<channel>` is passed to `werf bundle publish`, the annotation is dropped: it reaches neither the service nor the extra annotations of the bundle. No warning is emitted and the exit code is unchanged.
- Every other annotation keeps its current classification: `project.werf.io/*` and `*ci.werf.io/*` stay service annotations, anything else stays an extra annotation, and `project.werf.io/name` and `project.werf.io/env` are still set from `werf.yaml` and `--env`.
- VERIFIED: publishing the same project with a binary built before and after this change, then unpacking the chart layer from a local `registry:2`, differs by exactly this one key in `extra_annotations.json`.
- `werf bundle apply`, `bundle plan` and `bundle render` do not change: they still read the channel from `--add-annotation` and from the `WERF_ADD_ANNOTATION_*` variables, and a bundle published by an earlier werf still renders the channel stored in it.
- `werf.io/version` does not change either — `werf bundle publish` never wrote it, and the applying werf still sets it to its own version.

## Why

The release channel is a property of a werf binary, not of a chart. Baking the publisher's channel into the artifact makes it outlive the process it described: a bundle published from a `stable` pipeline reported `werf.io/release-channel: stable` on resources deployed by an `alpha` werf, or by one with no channel at all, so the annotation could not be trusted for the question it exists to answer. `werf/nelm` already holds this view — `werf.io/release-channel` sits in `CleanWerfIoRuntimeAnnos` and is stripped before resources are compared (`pkg/resource/spec/unstruct.go`), so the value was never part of the desired state anyway.

Dropping the key from the service-annotation condition alone would not have worked: the annotation would have fallen through to `extraAnnotations` and still reached the bundle, which is why it is skipped before classification.

### DIFF
--- a/cmd/werf/bundle/publish/publish.go
+++ b/cmd/werf/bundle/publish/publish.go
@@ -310,8 +310,6 @@ func runPublish(ctx context.Context, imageNameListFromArgs []string) error {
 		return fmt.Errorf("get user extra annotations: %w", err)
 	} else {
 		for key, value := range annos {
-			// The release channel of the werf that publishes the bundle says nothing about
-			// the werf that applies it, so it must not be baked into the bundle.
 			if key == "werf.io/release-channel" {
 				continue
 			}
--- a/cmd/werf/bundle/publish/publish.go
+++ b/cmd/werf/bundle/publish/publish.go
@@ -310,9 +310,14 @@ func runPublish(ctx context.Context, imageNameListFromArgs []string) error {
 		return fmt.Errorf("get user extra annotations: %w", err)
 	} else {
 		for key, value := range annos {
+			// The release channel of the werf that publishes the bundle says nothing about
+			// the werf that applies it, so it must not be baked into the bundle.
+			if key == "werf.io/release-channel" {
+				continue
+			}
+
 			if strings.HasPrefix(key, "project.werf.io/") ||
-				strings.Contains(key, "ci.werf.io/") ||
-				key == "werf.io/release-channel" {
+				strings.Contains(key, "ci.werf.io/") {
 				serviceAnnotations[key] = value
 			} else {
 				extraAnnotations[key] = value


dmmordvi:
## Verification

- Published the same minimal project twice against a local `registry:2` (`--repo localhost:5555/rc-test`, `--add-annotation` for `werf.io/release-channel=stable`, `project.werf.io/git=...` and `custom.io/foo=bar`), once with a binary built from the base commit and once with this change, then pulled the chart layer through the registry API and unpacked `extra_annotations.json`:
  - base: `{"custom.io/foo":"bar","project.werf.io/env":"test","project.werf.io/git":"...","project.werf.io/name":"rc-test","werf.io/release-channel":"stable"}`
  - this branch: the same map without `werf.io/release-channel`.
  The base run is the known-positive: a filter that silently matched nothing would have produced the same output on both.
- `werf bundle render` on three inputs, all with the new binary: the bundle published by the base binary still renders `werf.io/release-channel: stable`; the new bundle renders none; the new bundle with `--add-annotation=werf.io/release-channel=alpha` renders `alpha`. This covers the backward-compatibility and the applying-side claims.
- No test added, so no mutation to name. Reproducing this at unit level would mean constructing the whole publish path (giterminism manager, stages storage, registry client); the behavior is a two-line filter and the round-trip above exercises it end to end.
- Not run: `test/e2e/bundle-publish-apply` — it needs a cluster, and it asserts only `werf.io/version` and `project.werf.io/env`, so it would not have covered the claims either way.

## Review focus

- The `continue` placement is the whole change: it has to precede the `project.werf.io/` / `ci.werf.io/` classification, otherwise the annotation lands in `extraAnnotations` and is published anyway.
- Whether the breaking claim deserves a `BREAKING CHANGE:` footer, and so a major bump, is a call I did not make — see Follow-up.

## Follow-up

- [ ] Decide whether this needs a `BREAKING CHANGE:` footer on the squashed commit. Resources deployed from bundles lose the annotation, and a release note is the only mitigation for anyone selecting on it; the version bump is your call.
- [ ] Consider whether `werf ci-env` should keep exporting `WERF_ADD_ANNOTATION_WERF_RELEASE_CHANNEL` in pipelines whose only werf command is `bundle publish` (`cmd/werf/ci_env/ci_env.go`) — it is now a no-op there, which is harmless but misleading.

PR #7833, opened 2026-09-03
feat(build): collect operations statistics without debug logging

Fral738:
## Summary

werf can now collect low-level operations statistics without enabling debug logging: the new `--build-report-operations` flag adds the `Operations` and `StageCache` sections to the build report and prints the operations summary after the build. Previously the only way to get these timings was `--log-debug`, which floods CI logs with unrelated debug output.

## What

- With `--build-report-operations` (or `$WERF_BUILD_REPORT_OPERATIONS=1`; default off) the JSON build report gains the `Operations` and `StageCache` sections, and the "Operations summary" and "Stage cache summary" blocks are printed after the build.
  - VERIFIED: smoke-run against a local Docker backend — with the flag both sections and both console blocks appear, without it neither does.
- The flag is registered on every command that saves a build report: `build`, `bundle publish`, `converge`, `helm get-autogenerated-values`, `lint`, `plan`, `render`, `stages copy`.
- `--log-debug` still enables the collection as before; without either switch nothing is collected or emitted.
- The build operation in `Operations` and in the console summary is named `stage build` (formerly `image build` in the debug-only summary): it counts backend stage builds, not built images, so with shared stages its count differs from the number of images and now adds up with the `StageCache` counters.
  - VERIFIED: a two-image project with a shared `from` stage reports `stage build: 3` + `found in local stages storage: 1` = 2 images × 2 stages.
- The envfile report format does not change.
- The "Debug environment variables" page no longer disappears from the docs sidebar on `task doc:gen`: its entry existed only in the generated `documentation.yml`, not in the source `_documentation.yml`, so every regeneration silently dropped it.

## Why

The statistics collector was gated on the debug log level alone, coupling report content to log verbosity: getting machine-readable build timings in CI required full debug output — gigabytes of noise for two report sections. Always-on collection was rejected: printing the summary blocks to every user changes familiar output, and the section format becomes a de-facto API before it has settled; an explicit opt-in keeps the default behavior byte-identical. The `image build` operation is renamed in the same PR because this is the last moment it can be done without breaking report consumers.


Fral738:
## Verification

- Smoke on macOS with the local Docker backend, binary from `task build`: `werf build --build-report-operations --save-build-report --build-report-path report.json` on a minimal stapel project — `Operations` and `StageCache` present in the JSON, both summary blocks printed; the same build without the flag produces neither.
- Linux run with a branch binary on a real multi-image stapel project (8 images, Docker backend, remote repo): the report saved with `--build-report-operations` contains `Operations` with the renamed `stage build` operation plus stapel container and registry operations, and `StageCache` (`built: 22`, `found in repo stages storage: 1`).
- Linux run with the Buildah backend (`Runtime.Backend: buildah`) on a staged Dockerfile project with actual rebuilds: the report carries `Operations` with `stage build: 3` matching `StageCache` `built: 3`, so the renamed Buildah call sites were exercised at runtime.
- Rename smoke: after forcing a stage rebuild the report lists `stage build` and `grep -r 'OperationImageBuild\|"image build"' pkg/` returns zero hits. The renamed call sites in `pkg/container_backend` carry no build tags and compile on every platform.
- Mutation: `GetBuildReportOperations` hardwired to `false` → 3 specs in `cmd/werf/common` failed; `forceEnabled` dropped from the `newOperationsCollector` gate → gate spec in `pkg/build` failed; `buildStartedAt` zeroed in the enabled path → all 3 enabled gate entries failed. All mutations reverted, suites green (21/21, 14/14).
- Note: in a build that fails midway, `stage build` op count can exceed the `StageCache` `built` counter — the operation is observed around the backend build call, while the `built` event is counted later, after the deduplication check; an aborted run loses the trailing events. On successful runs the counters add up.
- Not run: full `task test:unit` on this host reports "Test Suite Failed" for `pkg/true_git` and `cmd/werf/common` while every spec passes; reproduced identically on a clean `origin/main` worktree — pre-existing local environment issue, not this diff.

## Review focus

- The `docs/_includes/reference/cli/*.md` diffs are `task doc:gen` output.
- Gate semantics live in one place: `newOperationsCollector` in `pkg/build/conveyor.go` (`forceEnabled || debug`).

- Known scope limitation, follow-up planned: the collector lives inside `Build()`/`ShouldBeBuilt()` only, so pre-build phases (werf config render, giterminism init, sync clientID registry calls) are not measured — the existing registry/git/docker hooks are no-ops until the collector enters the context.

## Follow-up

- [ ] Lift the collector to the command scope behind the same gate (`opstats.FromContext` reuse in `newOperationsCollector`, summary printed before command exit, `build time` line becomes `command time`), and add `config render` + `giterminism init` operations.

alexey-igrychev (approved):

PR #7836, opened 2026-09-07
fix(build, buildah): stop repeating stage output in the final build error

Fral738:
## Summary

When a stapel stage command or a Dockerfile `RUN` step fails under the Buildah backend, the final `Error:` line repeats the whole stderr of the stage that was already printed in the stage log, so a failing step shows its output twice. The Docker backend reports only the exit status. Reproduces from a clean checkout:

```yaml
project: stderrtest
configVersion: 1
---
image: app
from: alpine:3.20
shell:
  install:
  - echo MARKER-STDOUT-LINE
  - echo MARKER-STDERR-LINE >&2
  - exit 1
```

```sh
WERF_BUILDAH_MODE=auto werf build
```

The same happens with `dockerfile: Dockerfile` and `RUN echo MARKER-STDERR-LINE >&2 && exit 1`.

## What

- A failed stapel stage command under the Buildah backend ends with `unable to run commands script: RunCommand failed: exit status 1` and no stage output in the error; the stage log above it is unchanged. VERIFIED: on a Linux host with buildah 1.23.1, chroot isolation, before/after the patch.
- A failed `RUN` step in a Dockerfile image under the Buildah backend ends with `unable to build Dockerfile "...": building at STEP "RUN ...": exit status 1` and no stage output in the error. VERIFIED: same host, before/after.
- Under `--log-quiet` the stage stderr is still embedded in the final error for both stapel and Dockerfile builds, as on `main`, and the stage stdout is not printed anywhere. VERIFIED: same host, both repros with `--log-quiet`.
- The stage log still shows `subprocess exited with status 1` twice. This is printed by buildah itself for chroot isolation, once per level of its reexec chain; werf only relays the container stderr, and this PR does not change it.
- When a Buildah command or Dockerfile build runs without a log writer, its stderr is still embedded in the returned error, so output is never dropped.

## Why

Both `RunCommand` and `BuildFromDockerfile` sent stderr to the stage log writer and to a buffer, then appended that buffer to the error unconditionally. The buffer only carries information when nothing else receives the stream; with a log writer present it is a second copy that reaches the top-level error and makes the Buildah backend noisier than the Docker one for the same failure. Capturing into the buffer only when no log writer is set removes the duplication without losing output on the log-less paths.

`pkg/buildah` cannot tell a visible writer from a discarding one: under `--log-quiet` the logboek stream it receives is a level-bound proxy that swallows writes, so "a writer exists" would silently mean "nothing is shown" and the stderr would vanish from both the log and the error. The backend therefore passes a log writer only when logboek's default level is shown; the alternative of probing the writer inside `pkg/buildah` would tie it to logboek internals. With no log writer, stdout goes to `io.Discard` rather than nil, because buildah substitutes the process stdout for a nil writer and quiet builds would start printing stage output.

### DIFF
--- a/pkg/buildah/native_linux.go
+++ b/pkg/buildah/native_linux.go
@@ -454,19 +454,13 @@ func (b *NativeBuildah) BuildFromDockerfile(ctx context.Context, dockerfile stri
 		buildOpts.NoCache = true
 	}
 
-	errLog := &bytes.Buffer{}
-	if opts.LogWriter != nil {
-		buildOpts.Out = opts.LogWriter
-		buildOpts.Err = io.MultiWriter(opts.LogWriter, errLog)
-	} else {
-		buildOpts.Err = errLog
-	}
-
+	var stderrBuf *bytes.Buffer
+	buildOpts.Out, buildOpts.Err, stderrBuf = generateStdoutStderr(opts.LogWriter)
 	buildOpts.ContextDirectory = opts.ContextDir
 
 	imageId, _, err := imagebuildah.BuildDockerfiles(ctx, b.Store, buildOpts, dockerfile)
 	if err != nil {
-		return "", fmt.Errorf("unable to build Dockerfile %q:\n%s\n%w", dockerfile, errLog.String(), err)
+		return "", wrapStderrError(fmt.Sprintf("unable to build Dockerfile %q", dockerfile), stderrBuf, err)
 	}
 
 	return imageId, nil
@@ -544,7 +538,7 @@ func (b *NativeBuildah) RunCommand(ctx context.Context, container string, comman
 	}
 
 	if err := builder.Run(command, runOpts); err != nil {
-		return fmt.Errorf("RunCommand failed:\n%s\n%w", stderrBuf.String(), err)
+		return wrapStderrError("RunCommand failed", stderrBuf, err)
 	}
 
 	return nil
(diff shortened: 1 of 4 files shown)


Fral738:
## Verification

- Stapel and Dockerfile repros from the description run on a Linux host (Ubuntu 22.04, buildah 1.23.1, chroot isolation) with the released werf 2.75.3 and with the CGO build of this branch: in both cases the final error lost the `MARKER-STDERR-LINE` / `subprocess exited with status 1` block; the Dockerfile one now ends with buildah's own `building at STEP "RUN ...": exit status 1`.
- Both repros re-run with `--log-quiet` on the same host and branch: the only output is the final error, it contains `MARKER-STDERR-LINE`, and `MARKER-STDOUT-LINE` appears nowhere.
- Mutation: `LogWriter` branch restored to `io.MultiWriter(logWriter, stderrBuf)` → `should stream stderr to the log writer without keeping a copy for the error` failed on non-empty `stderrBuf`.
- Mutation: `wrapStderrError` always rendering the multiline form → the same spec failed on the error text.
- Mutation: level gate removed from `getBuildahCommonOpts` (unconditional `OutStream()`) → `passes no log writer under --log-quiet so stderr reaches the error` failed on a non-nil writer.
- Not run: no unit test drives `BuildFromDockerfile` wiring (needs a real buildah store); reverting its call to the old inline `MultiWriter` is only caught by the Dockerfile repro above, which was run.
- `pkg/buildah` specs are Linux-only (`native_linux_test.go`); they were run on the Linux host, 36/36, and do not execute on macOS. The `pkg/container_backend` specs run anywhere.

## Review focus

- `generateStdoutStderr` is shared by `RunCommand` and `BuildFromDockerfile` and now returns `io.Discard` for stdout when no writer is given; the `suppressLog=true` callers of `getBuildahCommonOpts` (mount, umount, rm, commit, from) previously got a nil writer and hence buildah's default of the process stdout.

reyreavman (changes requested):
- change treats "a log writer exists" as "the output is visible", under
`--log-quiet` thats false - `logboek`'s stream is a level-bound proxy that
silently discards - so stderr ends up in neither the log nor the error, and a
failed build reports only `exit status 1`

- `pkg/buildah` can't tell a real writer from a discarding one, but caller can
- gate `opts.LogWriter` in `getBuildahCommonOpts` instead

reyreavman (approved):

PR #7837, opened 2026-09-07
feat(build): support from in image dependencies

Fral738:
## Summary

Users can specify `from` in every image dependency while migrating configurations toward v3. The existing `image` field remains supported for now and emits a deprecation warning that directs users to `dependencies[].from`.

## What

- `dependencies[].from` is accepted for Stapel and Dockerfile image dependencies, including the image platform cross-reference validation that runs before the build.
- When both `dependencies[].from` and `dependencies[].image` are set, werf uses the value of `from` for both image kinds.
- When only `dependencies[].image` is set, werf preserves the existing dependency behavior and prints: `The dependencies[].image directive is deprecated and will be removed in v3. Please use dependencies[].from instead.` The warning is repeated in the `DEPRECATION WARNINGS` block at the end of the command output.
- The `werf.yaml` reference (`docs/_data/werf_yaml.yml`) lists `from` in both `dependencies` blocks and marks `image` as `DEPRECATED, use from`; EN and RU pages are generated from that schema. Prose examples keep `image`; the v3 branch migrates them together with the other deprecated directives.
- Dependency stage validation and imports remain unchanged.

## Why

Strict YAML decoding rejected the v3-compatible `dependencies[].from` field, preventing configurations from being prepared or backported before the migration. Keeping the legacy field without a warning would leave users without a migration path; rejecting configurations that contain both fields would make incremental adoption difficult.

### DIFF
--- a/docs/_data/werf_yaml.yml
+++ b/docs/_data/werf_yaml.yml
@@ -359,11 +359,16 @@ sections:
         collapsible: true
         isCollapsedByDefault: false
         directiveList:
-          - name: image
+          - name: from
             value: "string"
             description:
               en: "Dependency image name, which should be built before building current image"
               ru: "Имя зависимого образа, который должен быть собран до сборки текущего образа"
+          - name: image
+            value: "string"
+            description:
+              en: "Dependency image name, which should be built before building current image (DEPRECATED, use from)"
+              ru: "Имя зависимого образа, который должен быть собран до сборки текущего образа (DEPRECATED, используйте from)"
           - name: imports
             description:
               en: "Define target build args to import image information into current image (optional)"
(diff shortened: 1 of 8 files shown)


Fral738:
## Verification

- Mutation: dependency source selection changed from `from` to `image` -> the single-`from` cases for Stapel and Dockerfile failed with `image name is not specified for dependency`, and both precedence cases (Stapel, Dockerfile) failed with `image2` instead of `image3`.
- Mutation: deprecation warning emission made unreachable -> `dependency image deprecation warning / registered for image` failed (0 new warnings instead of 1).
- Mutation: deprecation warning emitted for any set `image`, including alongside `from` -> `not registered for from with image` failed (1 new warning instead of 0).
- Mutation: platform validator reverted to reading the raw `image` field -> `should validate successfully when dependency is specified with from field` failed with `requires dependency image="" ...`, the exact error the first `e2e_extra` run reported before the fix.
- E2E on a Linux host with Docker and buildah (`task test:e2e paths="./test/e2e/build" parallel=false`): `Images dependencies` (fixture on `from`) and `Complex build` + `Custom tag build` (fixtures still on legacy `image`, 11 specs) pass on `2c6920e527`; the trailing `DEPRECATION WARNINGS` block does not disturb their `ContainSubstring` assertions.
- Unscoped `task test:unit` on the development host: all 45 suites report every spec passing; ginkgo still exits non-zero for `cmd/werf/common` with 0 failed specs, a pre-existing host artifact reproduced on a clean `origin/main` checkout.

## Review focus

- `pkg/werf/global_warnings.GlobalDeprecationWarnings` is a new read-only accessor over the process-wide accumulator; it exists so the config tests can assert the registered delta instead of grepping deduplicated `PrintGlobalWarnings` output. No production caller.
- `rawDependency.imageName()` is now the single place resolving `from`/`image`; `toDirective` and `imagePlatformValidator` both go through it. The validator was the only consumer reading the raw field directly.

PR #7838, opened 2026-09-07
fix(build, stapel): stop build reuse across different external bases

reyreavman:
## Summary

Stapel images whose only difference is an external `from:` reference shared one
content-anchor digest, so werf skipped their builds entirely and served an image built
from a different base. Needs several stapel images in one project with identical
instructions and different external base references, built into the same repo.

## What

- Two stapel images with different external `from:` references now get different
  content-anchor digests; before, every external-base from stage contributed an empty
  argument list to the anchor and they collided.
- Stage digests are unchanged for every `from:` variant, so the layer cache is not
  invalidated: for an external base the build phase already appended that same
  reference string via prevImage, and a unit test pins the digest to the pre-fix value.
- Content-anchor digests of external-base stapel images shift once, so on the first run
  after the upgrade each such image rebuilds only its last stage, the anchor; every
  preceding stage hits the existing cache. No other image is affected.
- Unaffected, where a reader would expect otherwise: `from: scratch`, `from:` naming a
  werf image, `fromLatest: true`, and all Dockerfile images — each already carried its
  base identity into the anchor.
- An e2e case builds two stapel images that differ only in their external base and
  asserts the second is built rather than reused by content tag, then that a rebuild
  reuses each image by its own tag.

## Why

The from stage learns an external base only from the caller-supplied prevImage. The
build phase supplies it, but the content-dependencies path feeding the anchor has no
prevImage at all, so the anchor input for the from stage was the hash of an empty list —
identical for every external base. The anchor digest deliberately excludes the image
name so that genuinely identical images share work, which makes the base reference the
only thing that could tell these images apart, and it was missing. The collision is
stable by construction, so a `cacheVersion` bump shifts both digests equally and does
not break the tie.

Rejected alternative: appending the resolved manifest digest instead of the reference.
That would silently change `from:` semantics — a mutable tag would start being chased
without `fromLatest`.

### DIFF
--- a/pkg/build/stage/from.go
+++ b/pkg/build/stage/from.go
@@ -86,6 +86,11 @@ func (s *FromStage) GetDependencies(_ context.Context, c Conveyor, _ container_b
 		args = append(args, "scratch")
 	} else if s.fromImageName != "" && !s.fromExternal {
 		args = append(args, c.GetImageContentTagStageID(s.targetPlatform, s.fromImageName))
+	} else if s.fromImageName != "" {
+		// GetContentDependencies has no prevImage, so an external base must be appended here:
+		// otherwise every image with an external base contributes the same empty argument list
+		// to the content anchor and such images reuse each other's builds.
+		args = append(args, s.fromImageName)
 	} else if prevImage != nil {
 		args = append(args, prevImage.Image.Name())
 	}
--- a/pkg/build/stage/from.go
+++ b/pkg/build/stage/from.go
@@ -86,7 +86,7 @@ func (s *FromStage) GetDependencies(_ context.Context, c Conveyor, _ container_b
 		args = append(args, "scratch")
 	} else if s.fromImageName != "" && !s.fromExternal {
 		args = append(args, c.GetImageContentTagStageID(s.targetPlatform, s.fromImageName))
-	} else if s.fromImageName != "" {
+	} else if s.fromExternal {
 		// GetContentDependencies has no prevImage, so an external base must be appended here:
 		// otherwise every image with an external base contributes the same empty argument list
 		// to the content anchor and such images reuse each other's builds.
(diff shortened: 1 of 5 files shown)


reyreavman:
## Verification

- Mutation: removed the new digest-args branch → "gives different content digests to from
  stages with different external bases" failed; the other two specs stayed green.
- Mutation: killed both base-identity paths at once → the stage-digest spec failed too,
  so it is not vacuous.
- Mutation: appended a per-instance value next to the reference → "gives equal content
  digests to from stages with the same external base" failed.
- The prevImage contract is pinned by the pre-existing GetDependencies table, not by the
  new specs: after this change an external base never reaches the prevImage branch, since
  `from:` is required for stapel images and the earlier branches always match.
- Not run: an end-to-end build against a registry — needs two distinct external base
  images and a writable repo, unavailable in this environment.

## Review focus

- Whether the now-unreachable `prevImage` fallback in the from stage should stay as a
  defensive default or be dropped; it is kept here to hold the diff to the fix.

reyreavman:
## Verification

- Mutation: prefixed the appended reference (`"ext:"+ref`), which would silently shift every
  external-base stage digest → the golden table entry and "gives the content digest the same
  value as the stage digest" failed. Before the golden entry this mutation passed the whole
  suite.
- Mutation: removed the external-base branch (the original bug) → three specs failed,
  including "gives different content digests to from stages with different external bases".
- The golden digest was taken from the pre-fix code path: the from stage at the parent commit,
  with the reference passed as prevImage, yields the same value.
- The three GetDependencies table goldens changed in the refactor commit reproduce as
  `sha256("<param>:::")` — the `:::` join separator plus the empty prevImage name the removed
  branch appended; new values are `sha256("<param>")`, derived with sha256sum independently
  of the test.
- Not run: the new e2e case; it needs a registry and pulls of two base images, left to CI.
  Its red state on the pre-fix code has not been observed — worth one run against the
  parent of the fix commit if CI comes back green.

nervgh (approved):

PR #7841, opened 2026-09-08
fix(config): stop werf.yaml loading taking minutes on large image graphs

reyreavman:
## Summary

Every werf command that loads `werf.yaml` runs a loop check over the image dependency graph (`fromImage`, `import`, `dependencies`) that restarts a depth-first walk from every image and forgets what it has already walked, so every shared base image doubles the number of paths. On a project with ~1300 images and ~6700 edges `werf config render` spent ~45 s in that check while the templates themselves rendered in a few seconds. It reproduces with any config where many images share the same intermediate images; a 30-layer graph of 2 images per layer, each depending on both images of the previous layer, is enough to hang the old check indefinitely.

## What

- Loading a `werf.yaml` whose image graph is a DAG with many shared images completes in time linear in the graph size; the 30x2 diamond graph above is validated in milliseconds instead of never.
- A loop through `fromImage`, `import` or `dependencies` is still rejected with the same `infinite loop detected: a -> b -> a` error, including a loop that is reachable only through an image that another image had already walked.
- VERIFIED: on a real project with 1332 images `werf config render` drops from ~47 s to ~7 s with byte-identical output.
- The rendered output, the order of images and every other validation do not change.

## Why

The walk enumerated every root-to-leaf path instead of every edge. Remembering the images whose whole dependency subgraph was already walked without finding a loop and skipping them later is safe: a loop through such an image would have been reported by the walk that finished it, because that walk keeps the image on its stack until it returns. The neighbouring graph walks (`GroupImagesByIndependentSets`, `getImageRelativesInOrder`) deduplicate visited images already and stay polynomial, so they are left alone.

### DIFF
--- a/pkg/config/stubs_test.go
+++ b/pkg/config/stubs_test.go
@@ -89,10 +89,6 @@ func (image *ImageStub) GetName() string {
 	return image.name
 }
 
-func (image *ImageStub) IsFinal() bool {
-	return true
-}
-
 func (image *ImageStub) dependsOn() DependsOn {
 	return image.deps
 }
--- a/pkg/config/stubs_test.go
+++ b/pkg/config/stubs_test.go
@@ -70,3 +70,29 @@ func NewGitRepoArchiveStub() *GitRepoArchiveStub {
 func (archive *GitRepoArchiveStub) GetFilePath() string {
 	return "no-such-file"
 }
+
+type ImageStub struct {
+	ImageInterface
+
+	name string
+	deps DependsOn
+}
+
+func NewImageStub(name string, dependsOn DependsOn) *ImageStub {
+	return &ImageStub{
+		name: name,
+		deps: dependsOn,
+	}
+}
+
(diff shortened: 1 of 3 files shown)


reyreavman:
## Verification

- On a real 1332-image project: `werf config render` with the patched binary against the unpatched one, `time` and `cmp` of the output — 47.1 s → 7.3 s, output identical (macOS, warm `WERF_HOME`).
- Mutation: dropped the `acyclic[imageName] = struct{}{}` write (old behaviour) → the 30x2 diamond DAG spec fails by `SpecTimeout`; the other nine loop specs stay green.
- Mutation: marked the image acyclic before walking its dependencies → seven loop specs fail (every loop is missed).

## Review focus

- The safety argument for skipping already-walked images in `validateImageInfiniteLoop`; the reported loop path stays the stack from the root image, as before.

nervgh (approved):

PR #7842, opened 2026-09-08
chore(release): keep previous major releases out of latest

Fral738:
## Summary

Publishing a channel in a previous major could leave its GitHub release marked Latest after its title was already correct. The publication workflow now reconciles the latest marker separately: the stable release of configured major 2 is Latest, while rock-solid releases remain non-prerelease without the Latest marker. The workflow also scans up to 100 releases while clearing stale channel markers.

## What

- With `LATEST_MAJOR: "2"`, the release in group 2 carrying `stable` is marked Latest.
- A `rock-solid` release is non-prerelease and not Latest; its title retains the `[rock-solid]` channel marker.
- Every configured release outside `LATEST_MAJOR` has its Latest marker explicitly cleared even when its title is unchanged.
- The workflow determines the current Latest release through `GET /repos/werf/werf/releases/latest`, which is supported by the GitHub CLI used in the release runner.
- The workflow scans up to 100 releases while clearing the channel and Latest markers from releases removed from `trdl_channels.yaml`.
- VERIFIED: With `v2.77.2` temporarily marked Latest, a local workflow run restored Latest to `v2.75.3` and restored `v2.77.2` to a prerelease. `v2.75.3` currently carries both `stable` and `rock-solid`, so it is Latest through `stable`.

## Why

GitHub permits only one Latest release. Assigning it to both `stable` and `rock-solid` made the final state depend on associative-array iteration when the channels pointed to different versions. The workflow now reserves Latest for stable, while rock-solid remains a final release identifiable by its title. The GitHub CLI does not expose `isLatest` through `gh release view --json`, so the workflow reads the supported REST endpoint instead. The stale-release scan also used GitHub CLI's default page size of 30, which could leave channel markers beyond that page unprocessed.

### DIFF
--- a/.github/workflows/release_trdl-publish.yml
+++ b/.github/workflows/release_trdl-publish.yml
@@ -114,6 +114,7 @@ jobs:
       - name: Update GitHub releases based on trdl_channels.yaml
         env:
           GH_TOKEN: ${{ secrets.RELEASE_PLEASE_TOKEN }}
+          LATEST_MAJOR: "2"
         run: |
           check_release_exists() {
             local version=$1
@@ -126,11 +127,16 @@ jobs:
             gh release view "$tag" --json name | jq -r '.name'
           }
 
+          get_latest_release_tag() {
+            gh api 'repos/werf/werf/releases/latest' --jq '.tag_name'
+          }
+
           process_releases() {
             echo "Processing releases based on trdl_channels.yaml..."
 
             declare -A VERSION_CHANNELS
             declare -A GROUP_VERSIONS
+            latest_release_tag=$(get_latest_release_tag)
 
             current_group=""
             while IFS= read -r line; do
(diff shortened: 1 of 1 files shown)


Fral738:
## Verification

- Parsed `.github/workflows/release_trdl-publish.yml` with Ruby and passed the extracted `update_release` script to `bash -n`; only the `stable` channel of `LATEST_MAJOR: "2"` assigns Latest.
- Ran `actionlint .github/workflows/release_trdl-publish.yml` successfully.
- VERIFIED: Temporarily made `v2.77.2` a non-prerelease Latest release, then ran the extracted `update_release` script with `LATEST_MAJOR=2`. It set `v2.77.2` back to prerelease and restored Latest to `v2.75.3`, which currently carries the stable channel.
- Mutation: request unsupported `gh release view --json isLatest` → GitHub CLI rejected the field; the workflow now uses `GET /repos/werf/werf/releases/latest`.

PR #7845, opened 2026-09-08
fix(build, stapel): stop builds failing with many imports

alexey-igrychev:
## Summary

Legacy Stapel builds could fail with `argument list too long` when imports generated a script beyond the container shell's argument-size limit. Large stage scripts now execute without being passed as an argument.

## What

- Legacy Stapel stages transport the script through standard input, keeping the shell argument list independent of script size.
- Build commands cannot consume the rest of the stage: the script is read to EOF before anything is evaluated, so `cat` or `read` find standard input already drained.
- A failing script reader aborts the stage instead of committing an empty image, even when it has already produced part of the script.
- Command failures and syntax errors still fail the build without committing a stage image under the default build options.
- The script is evaluated quoted, so generated commands are no longer word-split and glob-expanded before evaluation. Only rsync filters generated from `includePaths`/`excludePaths` were exposed to that; user commands reach the container as mounted script files and never were.
- The debug dump of the build command prints the pipe form and stays runnable.
- Command ordering, `&&` chaining, CLI options, and `werf.yaml` syntax remain unchanged.

## Why

The whole encoded script previously occupied one `bash -ec` argument, eventually exceeding the operating system's argument-size limits when the container runtime started Bash. Reading it directly with `bash -s` removes that limit but shares the script stream with build commands, allowing a command to swallow the rest of the stage silently.

Assign the stream to a variable with Stapel's bundled `cat` before `eval`. The assignment keeps the script out of argv, leaves standard input at EOF for every evaluated command, and propagates a reader failure; the redirect to `/dev/null` on `eval` only makes the EOF guarantee explicit. A host script file plus bind mount would work as well and is what the Shell builder and the Buildah backend already do, but `LegacyStageImageContainer` owns no temporary directory, so that route would add tmp-dir plumbing and cleanup that leaks on SIGKILL. The script remains memory-backed; this removes the argument-size ceiling, not resource limits.

### DIFF
--- a/pkg/container_backend/legacy_stage_image_container.go
+++ b/pkg/container_backend/legacy_stage_image_container.go
@@ -3,9 +3,9 @@ package container_backend
 import (
 	"context"
 	"fmt"
-	"path"
 	"strings"
 
+	"github.com/alessio/shellescape"
 	"github.com/docker/docker/api/types"
 	"github.com/docker/docker/errdefs"
 	"github.com/docker/docker/runconfig/opts"
@@ -92,11 +92,22 @@ func (c *LegacyStageImageContainer) prepareRunArgs(ctx context.Context) ([]strin
 	runArgs = append(runArgs, setColumnsEnv)
 
 	args = append(args, runArgs...)
-	// Drain the script before eval, then give build commands EOF on stdin.
-	args = append(args, "-i", c.imageRef(c.image.fromImage), "-ec",
-		fmt.Sprintf(`eval "$(%s)" < /dev/null`, path.Join(stapel.CONTAINER_MOUNT_ROOT, "stapel/embedded/bin/cat")))
+	args = append(args, c.prepareRunCommandArgs()...)
 
 	return args, nil
+}
+
+func (c *LegacyStageImageContainer) prepareRunCommandArgs() []string {
+	// The assignment reads the script to EOF, so a reader failure aborts before eval
+	// and build commands find stdin already drained. The redirect only makes that explicit.
+	return []string{
+		"-i", c.imageRef(c.image.fromImage), "-ec",
+		fmt.Sprintf(`script=$(%s); eval "$script" < /dev/null`, stapel.CatBinPath()),
+	}
+}
+
+func (c *LegacyStageImageContainer) prepareDebugRunCommand(runArgs []string) string {
+	return fmt.Sprintf("printf '%%s' %s | docker run %s", shellescape.Quote(c.prepareRunCommand()), shellescape.QuoteCommand(runArgs))
 }
 
 func (c *LegacyStageImageContainer) prepareRunCommand() string {
--- a/pkg/container_backend/legacy_stage_image_container.go
+++ b/pkg/container_backend/legacy_stage_image_container.go
@@ -2,7 +2,6 @@ package container_backend
 
 import (
 	"context"
-	"encoding/base64"
 	"fmt"
 	"strings"
 
@@ -92,15 +91,17 @@ func (c *LegacyStageImageContainer) prepareRunArgs(ctx context.Context) ([]strin
 	runArgs = append(runArgs, setColumnsEnv)
 
 	args = append(args, runArgs...)
-	args = append(args, c.imageRef(c.image.fromImage))
-	args = append(args, "-ec")
-	args = append(args, c.prepareRunCommand())
+	args = append(args, c.prepareRunCommandArgs()...)
 
 	return args, nil
 }
 
 func (c *LegacyStageImageContainer) prepareRunCommand() string {
-	return ShelloutPack(strings.Join(c.prepareRunCommands(), " && "))
+	return strings.Join(c.prepareRunCommands(), " && ")
+}
+
+func (c *LegacyStageImageContainer) prepareRunCommandArgs() []string {
+	return []string{"-i", c.imageRef(c.image.fromImage), "-se"}
 }
 
 func (c *LegacyStageImageContainer) prepareRunCommands() []string {
@@ -123,10 +124,6 @@ func (c *LegacyStageImageContainer) prepareAllRunCommands() []string {
 	commands = append(commands, c.runCommands...)
 
 	return commands
-}
-
-func ShelloutPack(command string) string {
-	return fmt.Sprintf("eval $(echo %s | %s --decode)", base64.StdEncoding.EncodeToString([]byte(command)), stapel.Base64BinPath())
 }
 
 func (c *LegacyStageImageContainer) imageRef(img *LegacyStageImage) string {
@@ -300,7 +297,7 @@ func (c *LegacyStageImageContainer) run(ctx context.Context) error {
 	}
 
 	RegisterRunningContainer(c.name, ctx)
-	err = docker.CliRun_LiveOutput(ctx, runArgs...)
+	err = docker.CliRunWithInput_LiveOutput(ctx, strings.NewReader(c.prepareRunCommand()), runArgs...)
 	UnregisterRunningContainer(c.name)
 	if err != nil {
 		return fmt.Errorf("container run failed: %w", CliErrorByCode(err))
--- a/pkg/container_backend/legacy_stage_image_container.go
+++ b/pkg/container_backend/legacy_stage_image_container.go
@@ -3,6 +3,7 @@ package container_backend
 import (
 	"context"
 	"fmt"
+	"path"
 	"strings"
 
 	"github.com/docker/docker/api/types"
@@ -91,17 +92,15 @@ func (c *LegacyStageImageContainer) prepareRunArgs(ctx context.Context) ([]strin
 	runArgs = append(runArgs, setColumnsEnv)
 
 	args = append(args, runArgs...)
-	args = append(args, c.prepareRunCommandArgs()...)
+	// Drain the script before eval, then give build commands EOF on stdin.
+	args = append(args, "-i", c.imageRef(c.image.fromImage), "-ec",
+		fmt.Sprintf(`eval "$(%s)" < /dev/null`, path.Join(stapel.CONTAINER_MOUNT_ROOT, "stapel/embedded/bin/cat")))
 
 	return args, nil
 }
 
 func (c *LegacyStageImageContainer) prepareRunCommand() string {
 	return strings.Join(c.prepareRunCommands(), " && ")
-}
-
-func (c *LegacyStageImageContainer) prepareRunCommandArgs() []string {
-	return []string{"-i", c.imageRef(c.image.fromImage), "-se"}
 }
 
 func (c *LegacyStageImageContainer) prepareRunCommands() []string {
@@ -297,7 +296,7 @@ func (c *LegacyStageImageContainer) run(ctx context.Context) error {
 	}
 
 	RegisterRunningContainer(c.name, ctx)
-	err = docker.CliRunWithInput_LiveOutput(ctx, strings.NewReader(c.prepareRunCommand()), runArgs...)
+	err = docker.CliRunWithInput_LiveOutput(ctx, c.prepareRunCommand(), runArgs...)
 	UnregisterRunningContainer(c.name)
 	if err != nil {
 		return fmt.Errorf("container run failed: %w", CliErrorByCode(err))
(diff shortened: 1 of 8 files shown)


alexey-igrychev:
## Verification

- Docker Desktop on macOS, `linux/amd64` containers: `TASK_X_REMOTE_TASKFILES=1 WERF_TEST_K8S_DOCKER_REGISTRY=localhost:5001 task test:e2e paths='./test/e2e/build' labelFilter='legacy-stage-script || (import && !system-dirs)' parallel=1` passed 6 cases: four new stage-script regressions and two existing import cases.
- The new cases run through `LegacyStageImage.Build`, verify the final marker inside the committed image, and verify nonzero exit codes and no committed image on errors. The large script is 3.4 MB.
- Mutation: change the loader back to `bash -s` → the stdin-consumer case fails because the final marker is missing.
- Mutation: put the entire large script back in argv → the large-script case fails with `exec /.werf/stapel/embedded/bin/bash: argument list too long`.
- Mutation: suppress the container run error → both command-failure and syntax-error cases fail. All mutations were reverted before the final passing run.
- Full local unit verification is not green: the first run failed while signing a temporary Git commit (`gpg: signing failed: Cannot allocate memory`); a process-local `commit.gpgsign=false` retry reported `cmd/werf/common` as failed despite that suite's displayed `SUCCESS`. No unrelated code was changed.
- Full `task test:integration` completed in about 15 minutes and is not green: `test/legacy_e2e/suites/ci-env/base_test.go:68` expected output length 762 but got 764. The preceding build suites, including Git (143 cases), Imports (9 cases), and Mount (3 cases), passed. No unrelated code was changed.

## Review focus

- The loader drains stdin before applying `/dev/null` to `eval`; it uses Stapel's bundled absolute `cat` path. The helper accepts a string and constructs a fresh reader per invocation.
- `eval` is not whole-script parse-before-execute: a real-container check printed `EARLY` before a following syntax error and returned 2. The tested guarantee is failure propagation and no default stage-image commit, not atomic execution of every shell command.

pkg/container_backend/legacy_stage_image_container.go
```diff
@@ -92,15 +91,17 @@ func (c *LegacyStageImageContainer) prepareRunArgs(ctx context.Context) ([]strin
...
 
 	return args, nil
 }
 
 func (c *LegacyStageImageContainer) prepareRunCommand() string {
-	return ShelloutPack(strings.Join(c.prepareRunCommands(), " && "))
+	return strings.Join(c.prepareRunCommands(), " && ")
+}
+
+func (c *LegacyStageImageContainer) prepareRunCommandArgs() []string {
```
alexey-igrychev: **Blocking: the build script is now the container's stdin, so any command that reads stdin swallows the rest of it.**

`bash -s` reads a non-seekable stdin incrementally, precisely so that `read` inside a script works — the classic `curl | bash` pitfall. Reproduced in a real container:

```
$ printf 'cat >/tmp/eaten\necho SECOND_RAN\necho THIRD_RAN\n' \
    | docker run -i --rm --entrypoint bash bash:5 -se
exit=0        # SECOND_RAN / THIRD_RAN never printed

$ printf 'read -r x; echo "GOT=[$x]"\necho LATER_CMD\n' \
    | docker run -i --rm --entrypoint bash bash:5 -se
GOT=[echo LATER_CMD]   # exit 0
```

Before this change no `-i` was passed, so the container's stdin was closed and any such command got EOF; the script always ran to completion.

Nothing shields it here:

- `prepareRunOptions` adds no stdin redirection — only workdir, entrypoint and user.
- The user's stage script is just another element of the same `" && "` chain (`pkg/build/builder/shell.go`); bash reads that *file*, but its children inherit fd 0 = the pipe carrying the rest of the werf script. Same for the ansible builder.
- Service commands are interleaved with it: git patch application (`pkg/build/stage/git_mapping.go`), import rsync (`pkg/build/stage/dependencies.go`), mountpoint cleanup (`pkg/build/stage/from.go`). Whatever follows the stdin-consuming command is skipped.
- The resulting EPIPE on the docker side is logged at debug level only (`cli/command/container/hijack.go`), so there is no error signal. bash exits 0 and werf commits and caches the stage image unconditionally.

Net effect: a corrupt image stored under a valid stage digest, reproducible from cache forever — strictly worse than the loud pre-container `docker run` failure this removes.

Suggested fix, mirroring what the Buildah backend already does (`BuildahBackend.applyCommands` writes the joined commands to a host temp file, bind-mounts it to `/.werf/script.sh` and runs `sh /.werf/script.sh`): the legacy Docker-Server path already bind-mounts host temp dirs under `/.werf` (`pkg/build/builder/shell.go` mounts the stage temp dir at `/.werf/shell`), so nesting inside the stapel `--volumes-from` volume is proven to work and adds no new local-daemon requirement.

- write `prepareRunCommand()` into a file under `werf.GetTmpDir()`, mode `0o555` — keep the `" && "` join, since the import copy commands rely on `$?` against the previous element of the chain;
- `c.runOptions.AddVolume(fmt.Sprintf("%s:%s:ro", hostScriptPath, containerScriptPath))`;
- run args become `{imageRef, "-e", containerScriptPath}` — argv stays tiny, so the `MAX_ARG_STRLEN` limit is still fixed;
- remove the file on every exit path (`defer os.RemoveAll`), including the introspect-on-error path;
- with no caller left, delete `CliRunWithInput_LiveOutput` rather than leaving dead API.

Smaller fallback if the temp-file lifecycle is unwanted: keep streaming but drain stdin before executing — `{"-i", imageRef, "-ec", `eval "$(cat)" </dev/null`}`. Expansions precede redirections, so `cat` drains the pipe and `eval`'s children get `/dev/null`; verified against the repro above.

(generated by pi-pi)

pkg/container_backend/legacy_stage_image_container.go
```diff
@@ -300,7 +297,7 @@ func (c *LegacyStageImageContainer) run(ctx context.Context) error {
 	}
 
 	RegisterRunningContainer(c.name, ctx)
-	err = docker.CliRun_LiveOutput(ctx, runArgs...)
+	err = docker.CliRunWithInput_LiveOutput(ctx, strings.NewReader(c.prepareRunCommand()), runArgs...)
```
alexey-igrychev: Streaming the script also loses parse-before-execute.

`eval "<whole script>"` parsed the entire text before running anything, so a syntax error anywhere meant nothing ran at all. `bash -s` executes command-by-command as it reads, so a malformed command near the end of a generated script now leaves the stage partially applied — and, as with the stdin issue above, the partial result is committed and cached. Both the mounted-script fix and the `eval "$(cat)"` fallback restore whole-script parsing.

Minor, on the same line: `strings.NewReader` is single-use. Safe at this single call site, but if this call is ever wrapped in a retry the second attempt silently sends an empty script — bash exits 0 and an empty stage gets committed.

(generated by pi-pi)

pkg/container_backend/legacy_stage_image_container_test.go
```diff
@@ -1,10 +1,28 @@
...
 
 import (
+	"strings"
+
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
 )
 
+var _ = Describe("LegacyStageImageContainer run arguments", func() {
+	It("streams large commands through standard input", func() {
```
alexey-igrychev: This spec cannot fail.

`prepareRunCommandArgs()` ignores the container's commands entirely and returns a constant three-element slice, so the 3.6 MB string built above and the `NotTo(ContainSubstring("echo import"))` loop are unreachable by any assertion. The name — "streams large commands through standard input" — describes behaviour the test never touches. Revert the production change to pass the script in the args again and this spec still goes green.

Assert something that holds the invariant instead: that no element of the produced args exceeds 128 KiB no matter how many commands are added, and that `prepareRunCommand()` still contains the joined text. The regression that actually matters — a run command that reads stdin must not swallow the rest of the script — needs an integration or e2e case; there is also no coverage for a stage large enough to have tripped the argument-size limit in the first place.

(generated by pi-pi)

pkg/docker/container.go
```diff
@@ -75,6 +75,12 @@ func CliRun_LiveOutput(ctx context.Context, args ...string) error {
 	return doCliRun(ctx, cli(ctx), args...)
 }
 
+func CliRunWithInput_LiveOutput(ctx context.Context, input io.Reader, args ...string) error {
```
alexey-igrychev: This helper takes a single-use reader. The caller passes `strings.NewReader(...)`, which is consumed on first use, so any future retry wrapper around this call would silently send an empty stdin instead of the script. Consider taking a `func() io.Reader`, or documenting the single-shot contract.

If the script ends up bind-mounted as a file rather than streamed, this helper loses its only caller and should be deleted rather than left as unused API.

(generated by pi-pi)

alexey-igrychev (commented):
Review of the switch from a packed `-c` argument to streaming the assembled Stapel script over stdin.

The problem being fixed is real and correctly diagnosed: only the per-import rsync commands scale with the number of `import:` directives, and the whole joined string went into a single argv element, hitting the Linux `MAX_ARG_STRLEN` limit of 128 KiB. The docker plumbing added here is sound — `-i` sets `OpenStdin`/`AttachStdin`/`StdinOnce`, and the CLI `CloseWrite()`s the container's stdin when the reader reaches EOF, so bash terminates reliably; live output through logboek is preserved, committing a `-i` container does not leak stdin flags into the image config, and `CliErrorByCode` still unwraps `cli.StatusError`.

The chosen transport, however, introduces a silent-corruption regression that is worse than the failure it fixes — see the blocking inline comment on `prepareRunCommandArgs`. Two more findings are inline, plus two that fall outside the diff below.

Findings not anchorable to the diff:

pkg/stapel/stapel.go:— — `Base64BinPath()` is now dead code: its only caller was the `ShelloutPack` helper removed by this change. It is exported, so no linter will flag it. Either remove it, or state why exposing the embedded `base64` binary is still worth keeping.

(generated by pi-pi)

pkg/container_backend/legacy_stage_image.go:— — The debug dump `fmt.Printf("Docker run command:\ndocker run %s\n", ...)` no longer prints something that can be pasted and re-run, because the script is not in the args any more. The adjacent `Decoded command:` label is also stale — nothing is encoded now. Consider printing the script next to the args, or documenting how to feed it back in.

(generated by pi-pi)

pkg/container_backend/legacy_stage_image_container.go
alexey-igrychev: Fixed in e63bd92b7 using the suggested smaller fallback: `eval "$(/.werf/stapel/embedded/bin/cat)" < /dev/null`. The script is drained before evaluation, commands receive EOF, and the existing `&&` join stays intact. The bundled absolute `cat` path avoids depending on the base image's tools or PATH.

pkg/container_backend/legacy_stage_image_container.go
alexey-igrychev: The new loader restores eval-based execution. One correction to the rationale: `eval` does not guarantee whole-script parse-before-execute. In a real container, evaluating `printf EARLY` followed by a newline and `)` prints `EARLY` and exits 2. The existing build path returns on the run error before committing under the default options. Added actual Docker coverage for syntax errors and command failures, asserting both the returned code and the absence of a committed stage image; removing error propagation makes both cases fail.

The input helper now accepts a string and creates a fresh reader per invocation, so callers do not retain a consumed reader.

pkg/container_backend/legacy_stage_image_container_test.go
alexey-igrychev: Replaced the constant-argument assertion with Docker e2e tests through `LegacyStageImage.Build`: a 3.4 MB script, stdin-consuming `cat`/`read`, exit 23, and a syntax error. Success cases verify a final marker inside the committed image; failure cases verify the error code and no committed image.

All four pass. Actual production mutations were rejected: `bash -s` loses the marker, putting the large script back in argv produces `argument list too long`, and swallowing the run error fails both error cases. The final run also passed the two existing Docker import cases (6 passed total).

pkg/docker/container.go
alexey-igrychev: Changed the helper to accept immutable string input and construct `strings.NewReader` inside each invocation. A repeated call starts from the beginning without introducing a reader factory. The helper remains used by the streaming loader.

PR #7846, opened 2026-09-09
docs(dev): correct the local test requirements for agents

alexey-igrychev:
## Summary

`AGENTS.md` states that e2e and integration suites need Linux with Docker and kind, so an agent working on macOS records them as impossible to run and ships changes with no end-to-end evidence. The docker-backend specs do run there.

## What

- An agent on macOS now runs `task test:e2e` for docker-backend specs instead of reporting them as unrunnable: the stated prerequisites are Docker plus a reachable registry in `WERF_TEST_K8S_DOCKER_REGISTRY`, with `task test:setup:environment` presented as one way to get both rather than the only one.
- Only the Buildah-mode e2e entries are stated to need Linux, matching the same paragraph's existing note that the Buildah backend is not built on macOS.
- An agent that sees the unit suite fail in a package its diff does not touch now re-runs that suite alone before investigating, instead of treating it as its own regression.
- VERIFIED: the whole `legacy-stage-script` e2e label passed on macOS/arm64 against a plain local registry, 5 specs in 153 s, with no kind cluster provisioned for the suite and no Linux host.
- UNVERIFIED: `task test:integration` on macOS. Only `task test:e2e` was exercised, while the sentence covers both suites; running one legacy suite locally would settle it.
- The `task` command list, the verification order and the Buildah compilation note do not change.

## Why

Both sentences were written from CI's environment, where kind is always provisioned, so they state a CI precondition as the only one. Leaving that in place costs the evidence an agent could have produced: the same file tells agents to establish that a check cannot run here before concluding it, and a concrete "needs Linux and kind" beats that general instruction every time.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -87,11 +87,13 @@ After changing Go code, run these in order — `task format` mutates files, so i
 
 NEVER assume a change compiles. While iterating, scope the slow steps (`task lint:golangci-lint golangciPaths="./pkg/foo/..."`, `task test:unit paths="./pkg/foo/..."`), then run them unscoped before handing the work over.
 
+A failure in a package the diff does not touch is usually a host-environment flake, not your change: re-run that suite alone before investigating it.
+
 A green `task test:unit` does NOT prove a command runs. No unit test constructs the storage manager, so a command that dereferences a flag group it never registered dies with a SIGSEGV before doing any work while the whole unit suite stays green. After adding or changing a command, execute it once — via `task test:integration`, or the binary in `./bin/` — before calling it done.
 
 `git diff --check` cannot be a whole-repo gate. The CLI reference generator emits column-aligned help text, so the generated pages under `docs/_includes/reference/cli` and `docs/pages_en/reference/cli` carry trailing whitespace on every branch. Scope the check to authored files.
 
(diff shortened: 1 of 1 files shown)


alexey-igrychev:
## Verification

- `task test:e2e paths="./test/e2e/build/..." labelFilter="legacy-stage-script"` on macOS/arm64 with Docker 27.4.0 and `WERF_TEST_K8S_DOCKER_REGISTRY` pointing at a plain local `registry:2`: 5 Passed, 153 s. No kind cluster was provisioned for the suite.
- The flake sentence comes from two unscoped `task test:unit` runs that failed in `cmd/werf/common` and `pkg/true_git`, neither of them touched by the diff under review at the time; re-running each suite alone alternated pass and fail.
- Not run: `task test:integration` on macOS, so the claim about the integration suite rests on the e2e result plus both suites sharing `test/pkg/suite_init`.

## Review focus

- Whether `TASK_X_REMOTE_TASKFILES=1` should be named for `test:e2e` and `test:integration` as it already is for `format`/`lint*`. Without it the suite aborts with `task: Remote taskfiles are not enabled`, because it builds the binary through the same Taskfile (`test/pkg/suite_init/werf_binary_data.go:49`). Deliberately left out of this diff.

## Follow-up

- [ ] Decide whether `TASK_X_REMOTE_TASKFILES=1` gets documented for the test suites alongside `format`/`lint*`.

PR #7847, opened 2026-09-09
chore(release): merge main into 3

alexey-igrychev:
## Summary

Brings the 18 commits `main` gained since the previous merge into `3`, porting each onto 3's diverged code. Every claim below is a change `main` already shipped, now reaching 3.

### Stapel stage script

- Legacy stapel stages pipe their script through stdin, so a stage with many imports no longer fails with `argument list too long` (#7845). On 3 the script also carries the build-time env exports, so `prepareRunCommand`, `prepareDebugRunCommand` and the debug dump take a context.
- The script is read by Bash itself (`script=$(</dev/stdin)`) and not by `cat` from the stapel image (#7851): 3 pins stapel 0.7.2, which ships no `cat`, so the loader #7845 shipped fails every stapel stage with `/.werf/stapel/embedded/bin/cat: No such file or directory` before running a command. VERIFIED: a two-image project with 256 imports builds against a local Docker daemon, where the same project fails on the `cat` loader.
- `stapel.CatBinPath` does not arrive with the port, and the unit specs require the loader to reference nothing under the stapel mount root and to still run with an empty `PATH`.
- The debug dump prints the pipe form and stays runnable; command ordering, `&&` chaining, CLI options and `werf.yaml` syntax do not change.
- VERIFIED: `task test:e2e labelFilter="legacy-stage-script"`, 5/5 — CI runs e2e by the `simple`/`complex`/`extra` scopes only, so it does not cover this label.

### Buildah backend

- A failed stapel command or Dockerfile `RUN` under Buildah no longer repeats the stage stderr in the final `Error:` line; with no log writer (`--log-quiet`) the stderr is still embedded (#7836). Ported onto 3's mutex-protected stderr buffer, which stays in place.
- `RUN --mount=from=...` under Buildah pulls the missing mount image instead of failing with `image not known` (#7823). UNVERIFIED on this branch: the `staged_dockerfile_run_mount` label is in no CI scope and needs Linux with native Buildah, so only the backend-level unit specs stand behind it here.

### Build reports and locks

- `--build-report-operations` / `$WERF_BUILD_REPORT_OPERATIONS` is registered on every command that saves a build report, and the build operation is named `stage build` in the report and the console summary (#7833). 3 keeps its own `BuildDockerfile` signature and does not regain `CalculateDependencyImportChecksum` or the `checkContainerBackendSupported` call `main` still has.
- Host image locks taken by parallel image builds are no longer lost (#7826). The spec now pins the mutex itself, because the concurrent version observed a lost append only in about one run in three and the darwin `task` path builds without the race detector.

### Configuration

- `werf.yaml` loading no longer takes minutes on large image graphs: the infinite-loop walk memoizes images whose subgraph is already known acyclic (#7841). Ported onto 3's variant of the walk, which keeps skipping related images absent from the configuration — a spec now covers that skip, since without it such a graph panics.
- `dependencies[].from` does NOT change: `main`'s implementation (#7837) is dropped in favour of 3's, where `from` is the primary directive, `image` is deprecated, specifying both is an error and the deprecation warning is emitted at parse time. The new stapel-script fixture imports by `from`, so the suite prints no deprecation warning for a fixture werf ships.

### Release plumbing

- GitHub releases outside the latest major are no longer marked `latest` (#7842), and the 2.x trdl channels plus `.release-please/main/manifest.json` follow `main` through the 2.78.0 release; the 3.x channels are untouched.
- `CHANGELOG.md` keeps every 3.x section whole and gains 2.78.0 and 2.77.2 with their own entries at the head of the 2.x chain: taking the merge as-is left a 2.x heading between a 3.x section's first entry and the ones that follow, so the 3.x section claimed one fix and the 2.x one claimed the rest.
- Dependency pins do NOT change: `main`'s `werf/nelm` and `werf/kubedog` bumps are on a nelm line that diverged from 3's pin (17 ahead, 91 behind), so the two deploy fixes riding those bumps (#7839, #7834) deliberately stay out of 3.
- `AGENTS.md` gains `main`'s note about re-running an unrelated failing suite (#7846) and its corrected account of what a macOS host can verify for linux-only code (#7849), which supersedes 3's shorter paragraph; the two details only 3 had — that the e2e suites refuse to start without `WERF_TEST_K8S_DOCKER_REGISTRY` even when an entry needs no registry, and that they run best with `parallel=1` — are folded into it.

## Why

`3` is a long-lived release branch that has to keep receiving `main`'s fixes, and 15 files conflicted because both branches changed the same code: 3 had already implemented `dependencies[].from`, already carried a locked stderr buffer, and had already removed the legacy surfaces `main`'s hunks still touch. Taking either side wholesale was rejected — `--ours` would have dropped five fixes, `--theirs` would have resurrected code 3 deleted on purpose — so each conflict was resolved by porting `main`'s intent onto 3's shape.

The script loader was the one place where porting the intent meant changing the mechanism, and `main` has since adopted the same mechanism (#7851), so the two branches agree again. The `cat` of #7845 worked only because `main` pins stapel 0.6.2; 3 pins 0.7.2, where that binary is gone, and no test could see the difference — the reader-failure table substitutes its own reader and the e2e base image has `cat` on `PATH`. Bumping the stapel image to bring `cat` back was rejected: it needs a new artifact to keep a dependency that buys nothing, and `$(<file)` is a Bash redirection that holds whatever the image contains.

Taking `main`'s dependency pins was rejected for the same reason: 3 tracks its own nelm branch, and moving it onto `main`'s line would have pulled in 91 commits of unrelated nelm history to gain two deploy fixes.


alexey-igrychev:
## Verification

- `task test:e2e paths="./test/e2e/build/..." labelFilter="legacy-stage-script"` on macOS/arm64 against a plain local `registry:2`, 5/5, no kind cluster. The same label at the first push of this branch failed: the ported loader broke every stapel build.
- The 256-import project from *What*, built by hand with this branch's binary and with the base-`3` binary for comparison. The stapel 0.7.2 image was enumerated with `docker create` + `docker export | tar t` to establish that `cat` is absent while `bash` is not.
- Mutations, unit: restored the `stapel.CatBinPath()` loader → `[It] keeps large scripts intact and outside argv` failed on `the script loader must not depend on a binary of the pinned stapel image`; `script=$(cat)` → `[It] loads the script with no executable available` failed on `bash: cat: No such file or directory`; dropped the build-time env exports → both transport specs failed; removed the `acyclic` write in `validateImageInfiniteLoop` → the 30-layer diamond entry timed out; marked `acyclic` before recursion → all seven loop entries failed; removed the absent-image guard → the new entry panicked with `image "external-base" not found but must be found`; removed the mutex from `addSharedHostImageLock` → the new spec failed in 3 of 3 runs, where the concurrent one failed in 1 of 3.
- Mutation, e2e: expected exit code 24 for a stage command exiting 23 → `[It] a failing command prevents a committed image` failed, so the restored assertion really pins the code.
- The buildah stderr port was compiled and mutated inside a `golang:1.25` container (linux/arm64, CGO, `libbtrfs-dev`), because macOS excludes `pkg/buildah/native_linux_test.go` entirely — `Will run 0 of 0 specs` there against 42 in the container. Three mutations, all caught: teeing stderr into the buffer while a log writer is set, returning the caller's writer instead of `io.Discard`, and appending the buffered stderr unconditionally in `wrapStderrError`.
- Not run: `task test:e2e labelFilter="staged_dockerfile_run_mount"` and `task test:integration` — both need Linux with native Buildah, which is also why the #7823 claim is marked UNVERIFIED.

## Review focus

- 15 files conflicted; the hand-ported hunks are `pkg/build/conveyor.go`, `pkg/buildah/native_linux.go`, `pkg/config/werf.go`, `pkg/config/werf_test.go`, `pkg/container_backend/docker_server_backend.go` and `pkg/container_backend/legacy_stage_image_container.go`. Everything else is either an auto-merge or a whole-file take from one side.
- `$(</dev/stdin)` re-opens `/proc/self/fd/0` rather than reading the inherited descriptor, so a runtime handing the container a stdin the build user cannot re-open would fail where `cat` would not. Under Docker/runc the descriptor belongs to the container user and stapel stages run as `0:0` — confirm that holds for the rootless setups this branch has to support.
- `.github/workflows/release_trdl-publish.yml` arrives with `LATEST_MAJOR: "2"`, so on branch 3 a 3.x stable release would not be marked `latest`. Confirm that is the intent for this branch.
- `e2e_complex` failed on the last push with `unable to check native overlayfs support: open /home/runner/.local/share/containers/storage/compat.../upper/subdir: not a directory` — a runner-side storage probe, and the same job passed on the two previous pushes of this branch. Re-run requested.

## Follow-up

- [ ] Bump 3's `werf/nelm` pin on its own nelm branch to pick up the plan-building and resource-info parallelism fixes that `main` got via #7839 and #7834 (`werf/nelm`, then `go.mod` here).
- [ ] werf/werf#7851 ports the Bash-only loader to `main`, where the current `cat` survives only through the stapel 0.6.2 pin.
- [ ] werf/werf#7850 fixes what this branch shows next to the stapel work: on 3 a failed stage ends with `container run failed: ` and no cause, because docker/cli v29 reports a container exit as a `cli.StatusError` with an empty message.
- [ ] Give `legacy-stage-script` and `staged_dockerfile_run_mount` a CI scope (`test/e2e/build`, labels in `_test_e2e_regular.yml`'s `simple`/`complex`/`extra`), or they stay hand-run only.
- [ ] `task test:unit` fails on linux/CGO in `pkg/build/image`: `TestCalculateGlobsChecksumMatchedPaths/contents_only` panics with `reexec.Init() was not called in main()`. It reproduces without this merge and does not fire in CI's unit job, so it needs its own issue.

alexey-igrychev:
## Verification (update)

Since the previous comment the branch gained four commits (CHANGELOG resolution, the fixture importing by `from`, the `pkg/config` skip spec, the lock mutex spec) and the Buildah port was checked on Linux, which macOS cannot do.

- Buildah stderr port, in `golang:1.25` on linux/arm64 with `libbtrfs-dev` and cgo: `task test:unit paths="./pkg/buildah/..."` runs `42 of 42 specs` → all passed. On macOS the same command runs `0 of 42` — the specs are behind a linux build tag, so they had never executed before this run. `task build` produces a static linux/arm64 binary.
- Mutations of that port, all caught: tee stderr into the buffer when a log writer is set → `Expected MARKER-STDERR-LINE to be empty`; return the caller's writer instead of `io.Discard` → `Expected nil to equal io.discard`; drop the `stderr == ""` guard → `RunCommand failed:\n\nexit status 1` instead of `RunCommand failed: exit status 1`.
- Mutations of the new specs, all caught: drop the `GetImage(relatedImageName) == nil` skip in the config walk → `image "external-base" not found but must be found` `[PANICKED!]`; drop the mutex from `addSharedHostImageLock` → the new spec fails in 3 runs out of 3, where the concurrent spec caught it in about one.
- `task test:e2e paths="./test/e2e/build/..." labelFilter="legacy-stage-script"` on macOS/arm64 against a local `registry:2`: 5/5, and the suite now prints no deprecation warning.
- `task format`, `task build`, `task lint` (`0 issues`), `task test:unit` (49 suites) on the final commit.
- Not run: `staged_dockerfile_run_mount` (needs Linux with native Buildah), `task test:integration`, and any Buildah behavior end to end — the Linux check covered `generateStdoutStderr` / `wrapStderrError` at the unit level only.

## Review focus

- `tests:all` on `ef4d995ea` failed in `e2e_complex` with `unable to check native overlayfs support: open /home/runner/.local/share/containers/storage/compat.../upper/subdir: not a directory` — a runner-side storage probe, before werf built anything. The same code passed on the two preceding heads; the failed job was re-run.
- Full `task test:unit` under Linux/cgo has one failure this merge inherits rather than causes: `pkg/build/image` `TestCalculateGlobsChecksumMatchedPaths/contents_only` panics with `reexec.Init() was not called in main()`. The file is identical on base `3`, and CI's `unit` job is green, so it only fires in that container setup.

## Follow-up

- [ ] werf/werf#7851 ports the Bash-only script loader to `main`, where `cat` still exists but the dependency is just as unnecessary.
- [ ] werf/werf#7850 fixes the empty `container run failed: ` message on `3`: the docker/cli version pinned there returns an empty message for a container exit, which also made the `Code: ` matching in the introspection paths dead code.

alexey-igrychev:
## Verification (second merge of `main`)

`main` moved three commits on while this PR was open — `chore(main): release 2.78.0` (#7835), `chore(release): 2 alpha,beta,ea,stable` and `docs(dev): correct what a macOS host can verify for linux-only code` (#7849) — so `origin/main` was merged again at `41b2b595d`. No Go code is involved: the delta is `CHANGELOG.md`, `AGENTS.md`, `trdl_channels.yaml` and `.release-please/main/manifest.json`.

- `CHANGELOG.md`: `git diff origin/3 -- CHANGELOG.md` adds 30 lines and removes none, so every 3.x section is byte-identical to `3` and the 2.x chain reads 2.78.0, 2.77.2, 2.77.1, ... in order. Both 2.x sections are `main`'s verbatim, including the entries for #7839, #7834 and #7837 that 3 does not carry, because those sections record the 2.x releases rather than this branch's contents.
- `AGENTS.md`: took `main`'s corrected macOS/linux-only section over 3's single paragraph and folded 3's two specifics into it. The `WERF_TEST_K8S_DOCKER_REGISTRY` claim is checked against `test/e2e/build/suite_test.go:22`, where the variable sits in `RequiredSuiteEnvs`, so the suite refuses to start without it regardless of the entry.
- `trdl_channels.yaml` and `.release-please/main/manifest.json`: `main`'s side taken whole, and the diff touches only the `2` group and the 2.x version.
- `task format` leaves the tree clean, `task lint` reports `0 issues`. `task build` and `task test:unit` were not re-run for this merge — it changes no Go file, and `ef4d995ea` was green.

alexey-igrychev:
## Verification (base `3` and `main` caught up)

Both sides moved, so `origin/3` was merged in at `b4d754071` and `origin/main` at `7f664d6e3`; the PR is `MERGEABLE` again.

- `3` gained #7850 (the empty `container run failed: ` message) and #7848 (image names with plus signs). One conflict, in the import block of `pkg/container_backend/legacy_stage_image_container.go`: `main` removed the base64 packing and added `shellescape`, `3` added `errors` and `docker/cli`. Resolved as the union minus `encoding/base64`, which nothing in the merged file uses any more. The bodies auto-merged cleanly, and the result keeps both changes: the Bash loader from `main` and 3's `StatusError.StatusCode` reading in `run`, both introspection paths and `IsStartContainerErr`.
- `main` gained #7851, the same Bash loader this branch already carried, so that merge changes no content: `git diff b4d754071 7f664d6e3` is empty and the commit only records the ancestry.
- `task format` leaves the tree clean, `task build` succeeds, `task lint` reports `0 issues`, `task test:unit` passes 49 suites.
- `task test:e2e paths="./test/e2e/build/..." labelFilter="legacy-stage-script"` on macOS/arm64 against a local `registry:2`: 5/5. Worth re-running because #7850 rewrote the error `run` returns, and the spec reads the exit code out of it with `errors.As`.
- Not run for these two merges: `task test:integration`, and the Buildah-mode entries of any e2e scope — unchanged since the Linux check reported above.

## Follow-up

- [x] werf/werf#7851 merged into `main`.
- [x] werf/werf#7850 merged into `3`, so the exit-code fix now arrives through the base branch rather than this PR.

PR #7848, opened 2026-09-09
fix(build): accept image names containing plus signs

Fral738:
## Summary

`werf config render` now accepts image names containing `+`, including `libstdc++`, and image environment variables for those names are shell-safe in `werf compose` and envfile build reports.

## What

- Image-name segments accept `+` after their first character, including at the end, so `libstdc++` passes validation.
- `werf compose` and envfile build reports replace `+` with `_` in image environment variable names, so `libstdc++` uses the `WERF_LIBSTDC___` prefix.
- Image-name segments beginning with `+` remain invalid.
- The v2-to-v3 migration guides and envfile build-report documentation describe the `+` syntax and environment-variable normalization in English and Russian.

## Why

The validator required every image-name segment to end with an alphanumeric character, which rejected valid names such as `libstdc++` before configuration rendering. A segment must still begin with an alphanumeric character so the leading character cannot be punctuation; `+` is allowed at the end to support conventional names such as `libstdc++`. The existing compose and envfile paths also normalized only the previously allowed punctuation, producing invalid shell identifiers for newly accepted names; a shared normalizer keeps those paths consistent.


Fral738:
## Verification

- Ran `werf config render 'libstdc++' --dev` in a local consumer project with its required CI environment variables; it exited 0 and rendered `image: libstdc++`.
- Ran `task format`, `task build`, `task lint` (0 issues), and `task test:unit` (49 suites passed).
- Mutation: removed `+` from the image-name regex -> the `plus signs` and `trailing plus` `ValidateImage` cases failed.
- Mutation: removed `+` from image environment variable normalization -> the `image environment variable names` `plus signs` case failed.
- Mutation: replaced `%image_slug%` with `%image%` -> the `slugged image name` custom-tag case failed because `libstdc++` is not a Docker tag; `%image_slug%` produces `libstdc`.
- Mutation: removed `+` registry encoding -> the `managed image name encoding` case failed.
- Ran the docs legacy integration suite with `WERF_PLATFORM` and `DOCKER_DEFAULT_PLATFORM` unset; 1 spec passed.
- Ran `TASK_X_REMOTE_TASKFILES=1 WERF_TEST_BINARY_PATH=$HOME/work/werf/build/linux-amd64/werf WERF_TEST_K8S_DOCKER_REGISTRY=localhost:32768 task --yes test:e2e paths="./test/e2e/build/..." parallel=1 -- -focus="custom tag image name substitutions" -timeout=10m` on the remote Linux host; 2 specs passed.

PR #7849, opened 2026-09-09
docs(dev): correct what a macOS host can verify for linux-only code

alexey-igrychev:
## Summary

`AGENTS.md` told agents that Buildah changes "cannot be compiled or exercised locally" on macOS. Half of that is false — `task build:dev:linux:amd64:go` typechecks every `//go:build linux` file — and the sentence made agents skip the one check that would have caught a broken linux-only file. It also said nothing about `//go:build linux` **test** files, which no local gate typechecks at all.

## What

- An agent on macOS is now told that `task build` skips `//go:build linux` files entirely, and that `task build:dev:linux:amd64:go` is the command that typechecks them. VERIFIED: with a deliberate arity error in `pkg/buildah/native_linux.go`, `task build` returns success and `task build:dev:linux:amd64:go` fails with `not enough arguments in call to generateStdoutStderr`.
- The impossibility claim now covers only what is actually impossible: `task build:dev:linux:amd64:cgo` on macOS, which dies in `runtime/cgo` with `setresgid`/`setresuid` undeclared, so the CGO Buildah binary cannot be produced and Buildah runtime behavior cannot be exercised there. VERIFIED: hand-run on darwin/arm64.
- A new paragraph states that `task build`, `task lint` and `task test:unit` all stay green while a `//go:build linux` test file does not compile, and that `task build:dev:linux:amd64:go` does not build tests either. VERIFIED: with a deliberate arity error in `pkg/buildah/native_linux_test.go`, `task test:unit paths="./pkg/buildah/..."` reports `Test Suite Passed` on macOS.
- It names `GOOS=linux GOARCH=amd64 task test:unit paths="./pkg/foo/..."` as the typecheck for those files, and explains that its `exec format error` means compilation passed. VERIFIED: the same broken file makes that command fail with `not enough arguments in call to generateStdoutStderr / have (*bytes.Buffer, nil) / want (io.Writer)`.
- Executing linux-only tests is pointed at a linux container with the `goTags` from `Taskfile.dist.yaml`, mounting the repo and `$HOME/go/pkg/mod`.
- Nothing else in the paragraph changes: the e2e/integration requirements, `WERF_TEST_K8S_DOCKER_REGISTRY`, and the `REGISTRY_STORAGE_DELETE_ENABLED` note are untouched.

## Why

The claim was paid for. A downstream fork merge landed 1-argument calls to a 3-argument function in `pkg/buildah/native_linux_test.go`; `task format`, `task build`, `task lint` and `task test:unit` were all green, the change was pushed, and only an independent reviewer reading the file found it. An agent following the old sentence had no reason to look further, because it read as "this cannot be checked here" rather than "this is checked by a different command".


alexey-igrychev:
## Verification

Every claim was established by mutating the code and running the commands on darwin/arm64, then restoring:

- `pkg/buildah/native_linux.go` — added a bogus second parameter to `generateStdoutStderr`. `task build` → success; `task build:dev:linux:amd64:go` → `pkg/buildah/native_linux.go:497:52: not enough arguments in call to generateStdoutStderr / have (io.Writer) / want (io.Writer, int)`.
- `pkg/buildah/native_linux_test.go` — changed `generateStdoutStderr(logWriter)` to `generateStdoutStderr(logWriter, nil)`. `task test:unit paths="./pkg/buildah/..."` → `Ginkgo ran 1 suite / Test Suite Passed`; `GOOS=linux GOARCH=amd64 task test:unit paths="./pkg/buildah/..."` → `not enough arguments in call to generateStdoutStderr / have (*bytes.Buffer, nil) / want (io.Writer)`.
- The same file on an unbroken tree: `GOOS=linux GOARCH=amd64 task test:unit paths="./pkg/buildah/..."` → `fork/exec .../buildah.test: exec format error`, which is where the "compilation passed" reading comes from.
- `task build:dev:linux:amd64:cgo` → `linux_syscall.c:67:13: error: call to undeclared function 'setresgid'` / `linux_syscall.c:73:13: ... 'setresuid'` against the macOS SDK.
- Container execution: `docker run --rm -v "$PWD":/src -v "$HOME/go/pkg/mod":/go/pkg/mod -w /src -e CGO_ENABLED=0 golang:1.25 go test -tags="dfrunsecurity dfrunnetwork dfrunmount dfssh containers_image_openpgp" ./pkg/buildah/` → `ok github.com/werf/werf/v2/pkg/buildah`.
- `task lint:prettier` → `All matched files use Prettier code style!`

The `task lint` half of the third claim rests on the incident described in *Why* rather than on a mutation run here: a full `task lint` was green on a tree whose `pkg/buildah/native_linux_test.go` did not compile.

## Review focus

- Whether naming a raw `go test` inside a container is acceptable, given the repo rule that build and test go through `task`. There is no `task` target for executing a cross-GOOS test binary, and the sentence pins the tag list to `Taskfile.dist.yaml` so it cannot drift from what the task would pass.

PR #7850, opened 2026-09-09
fix(build, stapel): name the exit code of a failed stage command

alexey-igrychev:
## Summary

With the Docker Server backend a failed stapel stage command ends with `container run failed: ` and nothing after the colon, so the user is told the build failed but not why. Reproduces from a clean checkout:

```yaml
project: exitcode
configVersion: 1
---
image: app
from: registry.werf.io/base/ubuntu:22.04
shell:
  install:
  - exit 23
```

```sh
werf build --platform=linux/amd64
```

## What

- A stapel stage command exiting non-zero now ends with `container run failed: exit code 23`; before, everything after `container run failed: ` was empty. VERIFIED: the repro above, before and after, against a local Docker daemon.
- `--introspect-error` and `--introspect-before-error` again enter the introspection shell instead of failing the command with the shell's own exit status. Docker start failures (125, 126, 127) still abort as before. UNVERIFIED: the introspection paths were not exercised by hand; `werf build --introspect-error` on a failing stage would settle it.
- The error still wraps `cli.StatusError`, so a caller reading the exit code with `errors.As` keeps working.
- No flag, no output format and no other error text changes.

## Why

`docker/cli` used to render a container exit as `Status: , Code: 23`, and this code both printed that message and matched it as a string. Since the pin moved to v29, `cli.StatusError.Error()` returns the empty string when only `StatusCode` is set, which silently emptied the final error and turned two string matches into dead code: `strings.Contains(err.Error(), "Code: ")` never held, so the introspection paths treated every container exit as fatal, and `IsStartContainerErr` could never report a start failure.

Reading `StatusError.StatusCode` instead of the rendered message keeps the behavior independent of how the CLI decides to format its errors. Formatting the code into the message at the `docker.CliRun_LiveOutput` call sites was rejected: the exit status has to survive as a typed error for `werf run`, which propagates it as its own exit code.

### DIFF
--- a/pkg/container_backend/legacy_stage_container_exit_test.go
+++ b/pkg/container_backend/legacy_stage_container_exit_test.go
@@ -1,0 +1,56 @@
+package container_backend
+
+import (
+	"errors"
+	"fmt"
+
+	"github.com/docker/cli/cli"
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+)
+
+var _ = Describe("Legacy stage container exit status", func() {
+	DescribeTable("containerExitCode",
+		func(err error, expectedCode int, expectedOK bool) {
+			code, ok := containerExitCode(err)
+			Expect(ok).To(Equal(expectedOK))
+			Expect(code).To(Equal(expectedCode))
+		},
+		Entry("a bare status error", cli.StatusError{StatusCode: 23}, 23, true),
+		Entry("a wrapped status error", fmt.Errorf("container run failed: %w", cli.StatusError{StatusCode: 23}), 23, true),
+		Entry("a status error carrying a message", cli.StatusError{StatusCode: 125, Status: "no such image"}, 125, true),
+		Entry("any other error", errors.New("no such host"), 0, false),
+	)
+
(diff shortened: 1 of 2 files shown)


alexey-igrychev:
## Verification

- The repro in the description against a local Docker daemon on macOS/arm64: before, `error building stage install: container run failed: `; after, `error building stage install: container run failed: exit code 23`.
- Mutation: made `namedContainerExitErr` return its argument unchanged → `[It] reports the exit code through the run wrapping` failed on `container run failed: exit code 23`. Mutation: dropped the `code == 125/126/127` set from `IsStartContainerErr` → the three start-failure entries failed.
- Not run: `--introspect-error` / `--introspect-before-error` by hand, so the restored introspection behavior rests on the unit specs for `containerExitCode` alone.

## Review focus

- Whether `fmt.Errorf("exit code %d%w", code, err)` is acceptable for keeping the chain without a trailing separator, given the wrapped error renders as the empty string.
- `cmd/werf/run/run.go:306` asserts `err.(cli.StatusError)` on an unwrapped error; that path does not go through this wrapping, but it is the other place the exit code is read.

PR #7853, opened 2026-09-09
chore(deps): bump nelm to the head of its 2 branch

alexey-igrychev:
Moves the `werf/nelm` pin from `191171c` to `16e45e2`, the head of `nelm/2`. The range is a single commit, werf/nelm#706, and it changes only nelm's own `go.mod`/`go.sum`: it raises nelm's `werf/common-go` requirement to the version werf already requires, so module selection and the compiled result are unchanged and nothing is user-visible. `go mod tidy` additionally drops the stale `common-go` entry the old pin kept alive in `go.sum`.


alexey-igrychev:
## Verification

- `gh api repos/werf/nelm/compare/191171c87ea1...16e45e2e628b`: one commit, and its only changed files are `go.mod` and `go.sum` — no nelm source differs between the two pins.
- `go list -m github.com/werf/common-go` on the tree before and after the bump returns `v0.0.0-20260831162604-46c3b45163be` both times, so MVS already selected the version nelm's new requirement asks for and the bump moves no other module.

PR #7854, opened 2026-09-10
fix(build, buildah): stop re-compressing parent layers on stage push

Fral738:
## Summary

In Buildah mode, storing a Stapel stage into the repo re-exported and gzip-compressed every layer of the image — including all parent layers already present in the registry — before checking whether the registry had them. On multi-image projects with large base layers this made stages with a zero or byte-sized delta take minutes and write tens of gigabytes to disk per build. Reproduces from a clean checkout with any Stapel image whose `from` has several layers, once the parent stages are already in the repo and only a later stage is invalidated:

```yaml
project: blobcache
configVersion: 1
---
image: app
from: python:3.12-slim
shell:
  beforeInstall:
  - apt-get update && apt-get install -y --no-install-recommends curl && rm -rf /var/lib/apt/lists/*
  install:
  - echo v1 > /version
```

```sh
WERF_BUILDAH_MODE=auto werf build --repo <repo>
# change the install script, empty the local containers storage, rebuild:
WERF_DEBUG_BUILDAH=1 WERF_BUILDAH_MODE=auto werf build --repo <repo>
```

## What

- In Buildah mode `WERF_DEBUG_BUILDAH=1` no longer prints `Error creating a SQLite blob info cache at .../blob-info-cache-v1.sqlite, using a memory-only cache: ... no such table: sqlite_schema`; the file is created and populated on the first pull or push.
  - VERIFIED: on a rootless Buildah host with the repro above, pushing the invalidated `install` stage exports and compresses 1 layer instead of 7 (`Compressing blob on the fly` 7 → 1, `Skipping blob ... (already present)` for every parent layer).
- Parent layers pulled from the same registry earlier in the run are reused on push by their known compressed digest — a `HEAD` per layer and no export or compression.
  - The cache lives in `$XDG_DATA_HOME/containers/cache` (default `~/.local/share/containers/cache`); an ephemeral home still gets the benefit within one werf run, since the pull of the base stage seeds it.
- The Docker backend, the pull/push protocol, and the produced image digests do not change.
- A Linux/CGO unit test in `pkg/buildah` fails whenever the default blob-info cache degrades to memory-only, so a future dependency bump that reintroduces an incompatible SQLite breaks CI instead of silently slowing builds.

## Why

`containers/image` v5.30 stores the blob-info cache in SQLite and probes it with `SELECT 1 FROM sqlite_schema`, a table name that exists only since SQLite 3.33. werf linked SQLite 3.30.1: `go.mod` resolved `github.com/mattn/go-sqlite3` to the mistagged `v2.0.1+incompatible` (a 2019 snapshot) because `jmoiron/sqlx`, reached via `3p-helm-for-werf-helm`, requires that tag and `+incompatible` v2 wins module selection over the `v1.14.22` that `containers/image` requires. On the schema error `containers/image` silently falls back to a per-call in-memory cache, so with `DockerRegistryPushPrecomputeDigests` (enabled in werf) every push has to re-derive the compressed digest of every layer before it can ask the registry for it.

Pinning `go-sqlite3` to `v1.14.22` via `replace` restores the version `containers/image` is tested against. Excluding the bad tag instead would leave module selection with no valid v2 candidate for `sqlx`; upgrading `sqlx` is out of scope since the driver is only used in its tests.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -415,5 +415,6 @@ replace (
 	github.com/deislabs/oras => github.com/werf/3p-oras v0.9.1-0.20240115121544-03962ecbd40a // upstream not maintained
 	github.com/docker/buildx => github.com/werf/3p-docker-buildx v0.13.0-rc2.0.20241111114615-d77c2e1444ad // upstream not maintained
 	github.com/jaguilar/vt100 => github.com/tonistiigi/vt100 v0.0.0-20190402012908-ad4c4a574305 // upstream not maintained
+	github.com/mattn/go-sqlite3 => github.com/mattn/go-sqlite3 v1.14.22 // v2.0.1+incompatible is a mistagged 2019 release bundling SQLite 3.30.1; containers/image blob-info cache needs sqlite_schema (SQLite >= 3.33)
 	github.com/spf13/cobra => github.com/andremueller/cobra v0.0.0-20241025091859-0d550c15a8a4 // remove after merge to upstream
 )
--- a/go.sum
+++ b/go.sum
@@ -104,7 +104,6 @@ github.com/Microsoft/hcsshim v0.12.2 h1:AcXy+yfRvrx20g9v7qYaJv5Rh+8GaHOS6b8G6Wx/
 github.com/Microsoft/hcsshim v0.12.2/go.mod h1:RZV12pcHCXQ42XnlQ3pz6FZfmrC1C+R4gaOHhRNML1g=
 github.com/ProtonMail/go-crypto v1.0.0 h1:LRuvITjQWX+WIfr930YHG2HNfjR1uOfyf5vE0kC2U78=
 github.com/ProtonMail/go-crypto v1.0.0/go.mod h1:EjAoLdwvbIOoOQr3ihjnSoLZRtE8azugULFRteWMNc0=
-github.com/PuerkitoBio/goquery v1.5.1/go.mod h1:GsLWisAFVj4WgDibEWF4pvYnkVQBpKBKeU+7zCJoLcc=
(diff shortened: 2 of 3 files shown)


Fral738:
## Verification

- CGO binary built from this branch on a rootless Buildah host (Ubuntu 22.04, buildah 1.23.1, local plain-HTTP registry), repro from the description: emptied containers storage and `~/.local/share/containers/cache`, invalidated only `install`, rebuilt with `WERF_DEBUG_BUILDAH=1`. Store section of `install`: `exporting filesystem layer` 1, `Compressing blob on the fly` 1, `Skipping blob ... (already present)` 12, no cache errors, `blob-info-cache-v1.sqlite` created. Same scenario with a CGO binary built from `origin/main` on the same host: 3× `Error creating a SQLite blob info cache`, 7 layers exported and compressed. Both runs pushed successfully; werf exit code 0.
- `task format` and `task lint` clean, no changes produced.
- Not run: `task test:unit` reported `Test Suite Failed` for `pkg/true_git` while all 51 specs pass; this is a pre-existing host quirk reproducible on a clean `origin/main` worktree, not this diff.
- Mutation: dropped the `go-sqlite3` replace directive → `pkg/buildah` `blob info cache persists digest mappings across cache instances` fails with `fell back to a memory-only cache` (run with CGO on the Linux host; 37/37 pass with the pin).

## Review focus

- The `replace` directive pins the module for the whole build, including `3p-helm-for-werf-helm`'s `sqlx` import — confirm nothing outside `containers/image` and `sqlx` tests links the driver.
- Wall-clock gain on the reporter's workload (1.6 GiB images, 42 stages, 360 s Store for a +34 B delta) is inferred from the removed work, not measured.

## Follow-up

- [ ] Backport to branch `3` (same `go-sqlite3 v2.0.1+incompatible` pin in its `go.mod`).

PR #7856, opened 2026-09-10
feat(build): collect operations statistics for the whole command run

Fral738:
## Summary

Operations statistics (`--build-report-operations` / `--log-debug`) now cover the whole command run instead of only the conveyor: werf config render, giterminism initialization and pre-build registry/git calls are measured, and the summary is printed once before the command exits. Previously a command spending most of its time before the build reported a summary covering a fraction of it (observed: `build time: 0.14s` out of an 11.8s command, with a 4.2s config render invisible).

## What

- The operations summary blocks are printed before the command exits on the eight commands that save a build report (`build`, `bundle publish`, `converge`, `helm get-autogenerated-values`, `lint`, `plan`, `render`, `stages copy`); the console line is `command time:` instead of `build time:`. For `converge` the single summary prints after the deploy.
- Two new operations appear in the report and the summary: `config render` (werf.yaml templating) and `giterminism init` (git worktree/submodules initialization).
- Pre-build registry, git and docker calls are now counted: the existing hooks were no-ops until the collector entered the context.
  - VERIFIED: smoke on macOS with the Docker backend — the summary lists `config render` and `giterminism init`, the JSON report `Operations` section contains them alongside build-phase operations, and without the flag no summary is printed and no sections appear.
- With `--follow`, one summary covers the whole follow session, not each iteration.
- Commands not saving a build report (`compose`, `run`, `kube-run`, `export`) keep the previous conveyor-scoped behavior under `--log-debug`.
- The JSON report schema does not change; `Operations`/`StageCache` stay opt-in behind the same gate.

## Why

The collector was created inside `Conveyor.Build()`/`ShouldBeBuilt()`, coupling the statistics scope to the conveyor lifetime: everything before it — config render, giterminism init, sync clientID registry calls — ran with no collector in the context, so the already-placed measurement hooks recorded nothing and the printed total covered only the build. Command-scoped installation with collector reuse in the conveyor closes the gap without touching the hooks or the report schema; the alternative of leaving per-conveyor scope and wrapping each pre-build phase separately would keep the misleading `build time` total and duplicate gate logic in every command.


Fral738:
## Verification

- Smoke on macOS, Docker backend, binary from `task build`: `werf build --build-report-operations --save-build-report` on a minimal stapel project — summary printed once with `command time:`, `config render` and `giterminism init` present both in the console block and in the JSON `Operations`; a rerun without the flag prints no summary and the report carries no `Operations`/`StageCache`.
- Linux run with a branch `task build` binary on a real multi-image project (deckhouse `chrony/*` images, Docker backend, local registry): `command time: 60.71s` matches the command's `Running time 60.71 seconds` exactly; `config render` (4.00s) and `giterminism init` (0.12s) are reported alongside 14 `stage build` ops matching `StageCache` `built: 14` + 1 `found in repo`; parallelism factors shown for pull (×1.7) and push (×1.3). The run failed in a source-cloning stage for a reason unrelated to statistics, and the summary was still printed before exit — including a 0.38s run that died on config render, which reported `giterminism init` and `config render` alone.
- Mutation: `FromContext` reuse branch disabled in `newOperationsCollector` → the reuse spec in `pkg/build` failed; `InitOperationsStatistics` hardwired to skip installation → the helper specs in `cmd/werf/common` failed. Both mutations reverted, suites green (Build 22/22, Cmd Common 17/17).
- Not run: full `task test:unit` on the macOS host fails only in `pkg/true_git` with all specs passing — known pre-existing local environment issue, reproduced on clean main earlier.

## Review focus

- Summary printing moved verbatim from `pkg/build/conveyor.go` to `pkg/opstats/log_summary.go` with the time label parameterized — the conveyor path still prints `build time:` for non-migrated commands under `--log-debug`.
- The `docs/_includes/reference/cli/*.md` diffs are `task doc:gen` output (flag help text change).

## Follow-up

- [ ] Close the "Follow-up" checkbox in the PR #7833 comment once this merges.

alexey-igrychev (commented):
## Code Review

**Base:** `main` (merge-base `cc3ef724d`) · 26 files, +216/-69
Verified with: `task build`, `task doc:gen` (empty diff), `task lint:golangci-lint` on the touched packages, `task test:unit`, mutation runs against the new tests, a macOS + Docker smoke with a local registry, and a behavioural diff against a base-built binary.

### Verdict

The mechanics are right: the collector is installed at command scope, the conveyor reuses it, and `LogSummary` is extracted without changing the output format. But moving the print into a `defer` collides with a global side effect of nelm's `log.SetupLogging`, and the JSON report semantics under `--follow`/retry changed silently.

### DoD

| Criterion (inferred from the PR description) | Met? | Evidence |
| :--- | :--- | :--- |
| Summary covers the whole command run, `command time:` label | ✅ | `cmd/werf/common/operations_statistics.go:24`; smoke: `command time: 25.90s` |
| New `config render` / `giterminism init` operations | ✅ | `pkg/config/parser.go:83`, `cmd/werf/common/common.go:1467`; both present in report.json |
| Summary printed on all eight commands | ❌ | `render` — never; `helm get-autogenerated-values` — never |
| Conveyor does not create a second collector | ✅ | `pkg/build/conveyor.go:792`; mutating away the early return fails the test |
| JSON report schema unchanged | ✅ | `pkg/build/build_report.go:89` untouched |
| `--follow`: one summary per session | ✅ | 1 block over a 25s session vs 9 on base |
| Docs regenerated | ✅ | `task doc:gen` → empty diff |

---

### Major — `werf render --build-report-operations` no longer prints the summary

`cmd/werf/render/render.go:406` calls `log.SetupLogging(ctx, cmp.Or(GetNelmLogLevel(...), action.DefaultChartRenderLogLevel), ...)`, and `DefaultChartRenderLogLevel = log.ErrorLevel`. nelm then does `logboek.Context(ctx).SetAcceptedLevel(level.Error)` on the global logger (`nelm/pkg/log/logger_logboek.go:194`). The deferred `logOperationsSummaryFn` (`render.go:166-167`) runs after that, so the Default-level `LogBlock` is dropped.

Reproduction (local registry, project already built):

```
base: werf render --repo … --build-report-operations --log-verbose  → "┌ Operations summary … build time: 0.10s"
PR:   werf render --repo … --build-report-operations --log-verbose  → nothing
PR:   werf render --repo … --build-report-operations --log-debug    → printed
```

This is a regression against main: the flag silently does nothing. The same mechanism works in the opposite direction for `converge`/`plan`/`lint` (`InfoLevel` → `SetAcceptedLevel(level.Default)`) — summary visibility now depends on nelm's log level rather than on the werf flag.

Options: print the summary before invoking the nelm action, or capture the accepted level in `InitOperationsStatistics` and restore it inside the returned closure before printing.

### Major — the JSON build report became cumulative under `--follow`

`pkg/build/build_report.go:332` takes the collector from ctx; it is now shared across the whole command, so the report rewritten on every follow iteration contains the sum of all iterations:

```
PR   (10s follow --dev): Operations {'config render': 9, 'giterminism init': 10, …}  StageCache {'found in local stages storage': 9}
base (10s follow --dev): Operations {'docker daemon API': 2, 'local image inspect': 1} StageCache {'found in local stages storage': 1}
```

`--build-report-path` is parsed in CI, and `StageCache` no longer matches the last build. Same for `WithRetryBlock`: operations of a failed attempt land in the successful attempt's report. Neither the PR description nor `docs/pages_*/usage/build/process.md` mentions this.

### Minor — unbounded collector growth in a follow session

`GetGiterminismManager` is called on every follow iteration (`cmd/werf/common/follow.go:25`, 1s sleep), and each one now records an interval: 26 `giterminism init` ops over a 25-second session. That is ~86k intervals per day (~4 MB), and `Summary()` on every build becomes O(n log n) over the accumulated set. The summary is polluted by the polling loop too: `giterminism init` looks like a hot operation when it is just idle polling.

### Minor — the test on the returned `finish` asserts nothing

`cmd/werf/common/build_report_operations_test.go:72,79` — `Expect(finish).NotTo(BeNil())` on a function literal is always true. The closure's behaviour (printing the summary, the `command time` label) is uncovered: turning it into a no-op keeps the tests green. Mutating the gates themselves, by contrast, does fail the tests — verified in both files.

### Minor — `Conveyor.logOperationsSummary` degenerated into a one-line wrapper

`pkg/build/conveyor.go:804-806` — the method now only forwards to `opstats.LogSummary` with a constant label, across two call sites. It can be inlined and removed.

### Minor — the PR description contradicts the facts

werf squashes PRs, so the description becomes the commit body. "The operations summary blocks are printed before the command exits on the eight commands" is false for `render` (see Major) and for `helm get-autogenerated-values`: it uses `SetupLogOptionsDefaultQuiet` (`get_autogenerated_values.go:93`), accepted level Error, so the summary is not printed either before or after this PR. The VERIFIED section covers `build` only.

---

### Risks

| № | Risk | Type | Likelihood | Severity | Location |
| :--- | :--- | :--- | :--- | :--- | :--- |
| 1 | The flag silently produces no output | UX/Product | Likely | High | `cmd/werf/render/render.go:166,406` |
| 2 | CI parses a cumulative report | Operational | Possible | Medium | `pkg/build/build_report.go:332` |
| 3 | Memory growth and noise in the summary | Technical | Possible | Low | `cmd/werf/common/common.go:1467` + `follow.go:25` |

### Not verified

- Buildah paths were not compiled (macOS); e2e/integration were not run.
- `converge`/`plan` against a real cluster were not exercised — the nelm log-level effect there is derived from the code, with the mechanism proven on `render`.
- `task test:unit ./cmd/werf/common/...`: 17/17 Passed but the runner exits 1 — **the same on base** (14/14 Passed, exit 1), a pre-existing issue, not introduced by this PR.

Fral738:
Thanks for the thorough review — all findings addressed in a2c71231b, both Major mechanisms re-verified against your repros:

- **render summary regression**: the accepted log level is captured in `InitOperationsStatistics` and restored in the finish closure before printing. Your repro now prints the block (`werf render --build-report-operations --log-verbose` → `Operations summary … command time: 0.09s`). Covered by a regression spec that lowers the level to Error after installation and asserts the summary is still printed; making the closure a no-op or dropping the restore fails the tests (mutation-verified).
- **cumulative report under `--follow`/retry**: the collector got flush-delta summaries (`FlushSummary`/`FlushEventSummary`), and `createBuildReport` writes only the operations recorded since the previous report. Your repro: a 25s `--follow --dev` session prints one console summary with `giterminism init: 13`, while the last report carries `giterminism init: 1` and a single-iteration `StageCache`. Retry attempts land in the succeeding attempt's report — documented in the description and in `process.md` (en/ru) as intended: the retry is part of producing that build.
- **`finish` test asserting nothing**: replaced with output assertions (block header, operation row, `command time:` label, level-restore case); mutation `LogSummary(NewCollector())` → 4 failures.
- **`logOperationsSummary` wrapper**: inlined and removed.
- **description contradictions**: rewritten — the eight-commands claim scoped to where the accepted verbosity permits, `helm get-autogenerated-values` default-quiet behavior stated explicitly, VERIFIED lines updated for `render` and `--follow`.

Not fixed, with rationale:

- **unbounded collector growth in a follow session**: accepted for now — the collection is opt-in diagnostics, ~4 MB/day at a 1s polling loop, and the flush-delta change already keeps reports and their `Summary()` cost per-iteration. The console summary listing 26 `giterminism init` ops over a session is factual polling cost, not noise. If a bound turns out to be needed, interval compaction in `Collector.add` is the follow-up I'd take.

Fral738:
Second pass addressed in 5bf04197d:

- **Lost statistics on a failed report write (Major)**: `FlushSummary`/`FlushEventSummary` are split into `PendingSummary`/`PendingEventSummary` (no mark advance) and `CommitFlush`, called in `createBuildReport` only after `os.WriteFile` succeeds (or immediately when no report path is set). Regression test reproduces your scenario — failed write, then successful report carries counts `[2, 2]`; moving the commit before the write fails it (mutation-verified).
- **Quiet not protected (Minor)**: added a spec where the command starts with accepted level Error, an operation is recorded, the level is raised afterwards, and `finish()` must print nothing. Your surviving `acceptedLevel := level.Default` mutation now fails 2 specs.
- **`--follow` docs overpromise (Minor)**: reworded in en/ru — a report covers everything recorded since the previous report, explicitly including inter-build polling and failed retry attempts.

Fral738:
Third pass addressed in 03c025c5c:

- **Cumulative StageCache unprotected (Minor)**: the cumulative spec now records events before and after `CommitFlush` and asserts the total via `EventSummary`; your surviving mutation (`summarizeEvents(c.events, c.flushedEvents)` in `EventSummary`) now fails 2 specs — re-ran it to confirm.
- **API convention (Minor)**: `PendingSummary`/`PendingEventSummary`/`CommitFlush` now take `context.Context` first, call sites updated. `Summary`/`EventSummary` predate this PR (shipped in #7833) and are left as is to avoid scope creep.
- **Stale PR description claim (Minor)**: the `--follow` sentence now matches the docs — a report covers everything since the previous report, polling and failed retries included, plus the write-failure retention guarantee.

PR #7859, opened 2026-09-14
feat(build): include werf version in the build report

Fral738:
## Summary

The JSON build report now records which werf version produced it. Until now a report attached to an issue or compared across CI runs carried the backend and the in-container flag, but not the werf release, so the version had to be recovered from pipeline logs.

## What

- `Runtime` in the JSON build report (`--save-build-report`) gains `WerfVersion`, always present, placed first, before `Backend` and `InContainer`. The value is the same string `werf version` prints; a locally built binary reports `dev`.
- The `envfile` report format does not change: it never carried `Runtime`.
- Existing keys, their names and values in `Runtime` do not change; consumers that parse the report by key are unaffected.
- The build report section of the docs (`docs/pages_*/usage/build/process.md`) describes the field and the example report shows it.

## Why

`Runtime` already answers "what built this" for the backend and the container flag; the werf release is the remaining piece of that context and is what a reader needs first when a report is used to reproduce or diagnose a build. The field is named `WerfVersion` rather than `Version` so it cannot be read as a version of the report schema, mirroring `WerfImageName` next to `DockerImageName`.


Fral738:
## Verification

- Built `./bin/werf` from this branch and ran `werf build --save-build-report` on a one-image Dockerfile project against the local Docker daemon (macOS): `Runtime` in the written report starts with `"WerfVersion": "dev"`, followed by `Backend` and `InContainer`.
- Mutation: dropped the `WerfVersion:` line in `BuildPhase.BeforeImages` → `Simple build` failed on `checking build report runtime` with `""` vs `dev` (`task test:e2e paths="./test/e2e/build/..." labelFilter="build && simple"` on a Linux host with Docker).

## Review focus

- Whether always emitting `WerfVersion` (no `omitempty`) is the desired contract, given `Backend` and `InContainer` behave the same way.

PR #7860, opened 2026-09-14
feat(dev): publish JSON Schemas for werf configuration files

Fral738:
## Summary

werf now publishes JSON Schemas for `werf.yaml`, `werf-giterminism.yaml` and `werf-includes.yaml`, so an editor with a YAML language server (VS Code + Red Hat YAML, JetBrains, Neovim, …) offers completion, hover documentation and validation for these files. The schemas live in `schemas/` and are served by the documentation site at `https://werf.io/docs/v3/schemas/<name>.json`; the reference pages tell users how to attach them.

### Published schemas

- `schemas/werf.json` describes every `werf.yaml` document: the meta section (`configVersion`), a Dockerfile image (`dockerfile`) and a Stapel image (`image`), as the parser in `pkg/config` accepts them on this branch.
- Unknown keys fail validation at every level, matching werf's `UnmarshalStrict` behavior; `werf-includes.yaml` is the one exception, see below.
- Mutually exclusive directives fail when combined: `from`/`fromImage`, `before`/`after` on `import` and Stapel `dependencies`, `env`/`src`/`value` on secrets and `basicAuth.password`, `branch`/`tag` on `keepPolicies.references`, `from`/`fromPath` on `mount`; `branch`/`tag`/`commit`/`basicAuth` on a `git` mapping require `url`.
- Deprecated directives (`fromImage`, `image` in `import`/`dependencies`, `imageSpec.config.clearWerfLabels`, Dockerfile `ssh`) validate and carry `deprecationMessage`, so VS Code and IntelliJ strike them through.
- `cacheVersion`, `fromCacheVersion` and the per-stage `shell.*CacheVersion` accept a bare number as well as a string, because yaml.v2 coerces `cacheVersion: 1` and repository fixtures rely on it.
- `imageSpec.config.healthcheck` keys are `test`, `interval`, `timeout`, `startperiod`, `retries` — lower-case `startperiod` is what yaml.v2 reads from the untagged Go struct; the schema marks it UNDOCUMENTED.
- `schemas/werf-giterminism.json` is the schema werf already validated `werf-giterminism.yaml` against, moved out of a Go string constant and embedded via `//go:embed`; each directive gained a `description`. Runtime validation results do not change: the go-openapi validator ignores the added `$schema`, `title` and `description` keywords.
- `schemas/werf-includes.json` requires `git`, `add`, `to` and exactly one of `branch`/`tag`/`commit`; `add`/`to` must start with `/`, `includePaths`/`excludePaths` items must not. It rejects unknown keys although `pkg/includes` silently ignores them, so a misspelled directive that werf drops is flagged in the editor.

### Publishing

- The docs `assets` image adds `/schemas` (`*.json`) to the Jekyll source; Jekyll copies the files through byte-for-byte, so after the next docs deploy of this branch they are reachable at `https://werf.io/docs/v3/schemas/werf.json`, `…/werf-giterminism.json`, `…/werf-includes.json`. VERIFIED: files present and identical to the source in `_site/main/` and `_site/ru/` of a local `werf build` of the docs image. UNVERIFIED: the public URL answers 200 — it depends on the production deploy; nginx maps `/docs/v3/<tail>` to `/<tail>` and the ingress already routes the `/docs/v3/` prefix here.
- The `docs:deploy:latest` and `docs:validate` workflows also trigger on `schemas/**`.
- Reference pages `reference/werf_yaml.html`, `reference/werf_giterminism_yaml.html`, `reference/werf_includes_yaml.html` (en and ru) end with a "JSON Schema" section: the URL, a `yaml.schemas` snippet for VS Code and the `# yaml-language-server: $schema=` modeline. The URL is built from `site.canonical_url_prefix`, so it follows the branch's docs version. The `werf.yaml` page states that the schema describes the rendered YAML, not the Go template.
- Nothing is registered in SchemaStore yet; users attach the schema by hand until then.

### Tests

- `pkg/config` validates every non-templated `test/**/werf.yaml` (86 files) against `schemas/werf.json`, plus tables of accepted documents covering every directive and 34 rejected documents; a directive added to the parser and used in a fixture without a schema update fails `task test:unit`.
- A new root package `schemas` compiles every `schemas/*.json` as draft-07 with `santhosh-tekuri/jsonschema/v6` (the validator IDEs' behavior is closest to), validates all `test/**/werf-giterminism.yaml` fixtures, and holds the accept/reject tables for `werf-includes.yaml` including the documentation example.
- `github.com/santhosh-tekuri/jsonschema/v6` moves from indirect to direct in `go.mod`; it was already in the module graph and is used by tests only.

## Why

Without a schema the only feedback on a typo in `werf.yaml` is a werf run failing, and the only documentation is the reference page. Editors already know how to consume JSON Schema for YAML, so publishing one turns the reference into completion and hover text at the point of typing.

The schema is served under the versioned docs prefix rather than a single `werf.io/schemas/` URL because the site root belongs to werf/website and this repository only owns `/docs/<version>/` in the ingress; a per-version URL also means the schema and the parser it describes always come from the same branch, so no cross-version union with a deprecated tail is needed. The giterminism schema was moved instead of copied because a second copy of a validator input is a drift waiting to happen; embedding the published file makes the IDE and the CLI validate against the same bytes by construction.


Fral738:
## Verification

- Docs build: `cd docs && ../bin/werf build assets --dev`, then `werf run assets --dev -- ls /app/_site/main/schemas` and `cmp` against `/app/docs/schemas/*.json` — all three files present in `_site/main/` and `_site/ru/`, byte-identical. Rendered HTML of the six reference pages inspected for the resolved URL (`https://werf.io/docs/v3/schemas/...`) and the `yaml.schemas` snippet.
- Mutation: renamed the `dependencies` keyword on `gitMapping` in `werf.json` → `rejects a document / local git mapping with branch` failed.
- Mutation: `additionalProperties: true` on `meta` → `rejects a document / meta with unknown directive` failed.
- Mutation: removed `healthcheck.startperiod` → `accepts a document / dockerfile image with every directive` failed.
- Mutation: `"required": "giterminismConfigVersion"` (string instead of array) in `werf-giterminism.json` → `are valid draft-07 schemas` and the fixture test failed.
- Mutation: removed `properties.cli` from `werf-giterminism.json` → fixture test failed.
- Mutation: removed the `branch`/`tag`/`commit` `oneOf` from `werf-includes.json` → `include without ref` and `include with branch and tag` failed.
- Mutation: `"pattern": "^/"` → `".*"` on `add`/`to` in `werf-includes.json` → `relative add` and `relative to` failed.
- Mutation (review round): stripped every `minLength` from `werf.json` → the 9 new empty-value reject entries failed and nothing else; removed the `contextAddFile`/`contextAddFiles` `not` → `dockerfile image with both contextAddFile and contextAddFiles` failed; stripped `minLength` from `werf-includes.json` → `include with empty branch`, `basicAuth password with empty source` failed; restored the flat root `oneOf` → `reports a typo in a stapel image against the stapel branch only` failed.
- `https://werf.io/docs/latest/reference/werf_yaml.html` and `.../usage/build/stapel/base.html` answer 200 on the first hop (no redirect), so the `#anchor` fragments in hover links survive.
- Mutation (round 2): `scalarMap` values back to `type: string` → parity entry `bare numbers and booleans where werf reads strings` fails on `/imageSpec/config/env/PORT: got number, want string`; an unknown key injected into the same parity document → the parser side fails with ``unknown fields: `unknownDirective`!``, proving `parseWerfDocument` validates rather than just unmarshals.
- Plain `task test:unit` now prints `Schemas Suite - 22/22 specs`; plain `task lint` invokes golangci-lint with `./schemas/...`.
- Mutation (round 3): `cmd`/`healthcheck.test` back to `stringArray`, `gitMapping.tag` and `regexpOrString` back to `string` → `bare numbers and booleans where werf reads strings`, `numeric git refs`, `numeric branch in cleanup policy` fail; `includes.branch` back to `string` → `include pinned to a numeric branch` fails.
- `task build:dev:linux:amd64:go`, `task lint`, `task test:unit` unscoped — green.

## Review focus

- `schemas/werf.json`: `minLength: 1` sits on the shared `regexpOrString` definition, so besides `keepPolicies.references.branch/tag` it also forbids an empty item in `imageSpec.config.removeLabels`/`removeEnv` — the parser accepts `[""]` there, though the value is meaningless. Putting the constraint next to `$ref` is not an option in draft-07 (siblings of `$ref` are ignored); say if you would rather see an `allOf` wrapper on the two `references` properties instead.
- `schemas/werf-includes.json` is stricter than `pkg/includes` (which ignores unknown keys). Confirm that is the intended user experience rather than a bug to fix on the parser side first.
- `pkg/giterminism_manager/config/openapi_spec.go`: the JSON now carries draft-07 keywords go-openapi (Swagger 2.0 dialect) does not know; existing testify tests pass, but check nothing else feeds this schema into a stricter loader.
- Docs/code mismatches found while writing the schemas, left untouched here: `docs/_data/werf_yaml.yml` and `werf_includes_yaml.yml` document `basicAuth.password.file` while the code reads `src`; `docs/pages_*/usage/build/images.md` shows `healthcheck.interval: "30s"` while the Go type is `int` seconds.

## Follow-up

- [ ] Once `3` becomes `main` (and `docs.currentRoot` is `v3`), open a PR in `SchemaStore/schemastore` adding catalog entries for `werf.yaml`/`werf.yml`, `werf-giterminism.yaml`, `werf-includes.yaml` pointing at `https://werf.io/docs/v3/schemas/*.json`.
- [ ] After the docs deploy of this branch, `curl -I https://werf.io/docs/v3/schemas/werf.json` and settle the UNVERIFIED claim.
- [ ] Fix the `password.file` → `src` and `healthcheck.interval` type mismatches in `docs/` (separate docs PR).

schemas/werf.json
```diff
@@ -0,0 +1,1160 @@
...
+                    "$ref": "#/definitions/final"
+                },
+                "cacheVersion": {
+                    "$ref": "#/definitions/cacheVersion"
+                },
+                "platform": {
+                    "description": "List of target platforms for this image (for example ['linux/amd64', 'linux/arm64', 'linux/arm/v8']).",
+                    "$ref": "#/definitions/platform"
+                },
+                "from": {
```
alexey-igrychev: `from: ""` satisfies `{"required": ["from"]}`, so this document passes the schema:

```yaml
image: a
from: ""
```

while `StapelImageBase.validate` (`pkg/config/stapel_image_base.go:127`) rejects it with ``​`from: IMAGE` required!``. Add `"minLength": 1` here and to `fromImage`, and to every other string that a `oneOf` discriminates on: `secret.env/src/value`, `import.from/image`, `mount.from/fromPath`, `dependencyBase.from/image`, `keepPolicies.references.branch/tag`, `basicAuth.password.env/src/value`. Worth a reject-table entry per group — the tables currently only cover the missing-key case, not the empty-value one.

schemas/werf.json
```diff
@@ -0,0 +1,1160 @@
...
+                },
+                "context": {
+                    "description": "Build context path inside the project directory.",
+                    "type": "string"
+                },
+                "contextAddFiles": {
+                    "description": "Untracked files and directories to add to the build context. Paths are relative to the context path.",
+                    "$ref": "#/definitions/stringOrStringArray"
+                },
+                "contextAddFile": {
```
alexey-igrychev: The description says "the two cannot be used together", but the schema does not enforce it — `contextAddFile` and `contextAddFiles` side by side validate fine. The parser does enforce it (`pkg/config/raw_image_from_dockerfile.go:105-107`, "only one out of contextAddFiles and contextAddFile directives can be used at a time").

```json
"not": { "required": ["contextAddFile", "contextAddFiles"] }
```

on `dockerfileImage`, plus a reject entry.

schemas/werf.json
```diff
@@ -0,0 +1,1160 @@
+{
+    "$schema": "http://json-schema.org/draft-07/schema#",
+    "title": "werf.yaml",
+    "description": "werf configuration file. The file consists of several YAML documents separated by `---`: exactly one meta section (identified by `configVersion`) and any number of image sections (identified by `dockerfile` or `image`).\nhttps://werf.io/docs/v2/reference/werf_yaml.html",
+    "type": "object",
+    "oneOf": [
```
alexey-igrychev: `pkg/config/parser.go:847-870` dispatches on a single key — `configVersion` → meta, `dockerfile` → dockerfile image, `image` → stapel image. A flat three-branch `oneOf` throws that information away, and yaml-language-server then reports the failures of all three branches at once. This is what a single typo in a stapel image produces today:

```
- at '': additional properties 'image', 'contextAddFiles', 'target', 'args', 'addHost', 'network', 'dockerfile', 'cacheVersion', 'final', 'staged', 'context', 'platform', 'dependencies', 'secrets', 'imageSpec' not allowed
- at '': additional properties 'staged' not allowed
- at '': additional properties 'contextAddFiles', 'target', 'args', 'addHost', 'dockerfile', 'staged', 'context' not allowed
```

`allOf` of three `if: {"required": ["configVersion"]}, then: {"$ref": ...}` mirrors the parser exactly and reports from the branch the user actually meant. (See microsoft/vscode-json-languageservice#60 for why `oneOf` reads this way.)

schemas/werf.json
```diff
@@ -0,0 +1,1160 @@
+{
+    "$schema": "http://json-schema.org/draft-07/schema#",
+    "title": "werf.yaml",
+    "description": "werf configuration file. The file consists of several YAML documents separated by `---`: exactly one meta section (identified by `configVersion`) and any number of image sections (identified by `dockerfile` or `image`).\nhttps://werf.io/docs/v2/reference/werf_yaml.html",
```
alexey-igrychev: 39 links across the three schemas point at `werf.io/docs/v2/...` (30 here, 7 in `werf-giterminism.json`, 2 in `werf-includes.json`), zero at v3 — while the schemas are published under `/docs/v3/schemas/` on this branch. This is hover documentation, so a v3 user reads v2 pages.

`https://werf.io/docs/latest/reference/werf_yaml.html` returns 200, and a version-neutral link also survives merges between release branches, where a hardcoded `v2` silently rots.

schemas/werf-includes.json
```diff
@@ -0,0 +1,95 @@
...
+              },
+              "oneOf": [
+                { "required": ["env"] },
+                { "required": ["src"] },
+                { "required": ["value"] }
+              ]
+            }
+          }
+        },
+        "branch": {
```
alexey-igrychev, line 90: `branch: ""` satisfies `{"required": ["branch"]}`, so the schema accepts a document that `pkg/includes/config.go:135` rejects with "specify only `branch` or `tag` or `commit`" (`exactlyOne` tests `!= ""`). Same for `tag` and `commit`. `"minLength": 1` on all three.

schemas/werf-includes.json
```diff
@@ -0,0 +1,95 @@
...
+              "description": "Username.",
+              "type": "string",
+              "minLength": 1
+            },
+            "password": {
+              "description": "Password source. Exactly one of `env`, `src` or `value` must be set.",
+              "type": "object",
+              "additionalProperties": false,
+              "properties": {
+                "env": { "description": "Environment variable name.", "type": "string" },
```
alexey-igrychev: Same empty-string hole: `password: {env: ""}` passes the `oneOf` but fails `pkg/includes/config.go:101` ("specify only env or src or plain as basic auth password source"). `"minLength": 1` on `env`, `src` and `value`.

.github/workflows/docs_validate.yml
```diff
@@ -7,6 +7,7 @@ on:
       - "Taskfile.dist.yaml"
       - "cmd/werf/**"
       - "docs/**"
+      - "schemas/**"
```
alexey-igrychev, line 10: `schemas/**` is added here and in `docs_deploy_latest.yml`, but not to the `workflow_proceed` paths filter in `.github/workflows/tests.yml:30-42`. Since `doc_unit` in this workflow only runs `packages: cmd/werf/docs`, a PR that touches nothing but `schemas/*.json` goes green without running the Config Suite or the Schemas Suite — the very tests this PR adds to catch schema drift. Please add `- 'schemas/**'` to that filter.

pkg/config/helpers_test.go
```diff
@@ -0,0 +1,72 @@
...
+		}
+		if entry.IsDir() || entry.Name() != "werf.yaml" {
+			return nil
+		}
+
+		data, err := os.ReadFile(path)
+		if err != nil {
+			return err
+		}
+		if strings.Contains(string(data), "{{") {
```
alexey-igrychev, line 107: This skips 18 of the 104 `test/**/werf.yaml` fixtures (86 remain). Sensible as a default, but it means a directive used only in a templated fixture escapes the guard. Worth noting in the PR description alongside the "86 files" figure.

Also: `stapel/werf.yaml` at the repository root is non-templated and outside `test/`, so it is not in the corpus. I validated it against the schema by hand and it passes — cheap to include by walking `.` with the existing filter.

schemas/schemas.go
```diff
@@ -0,0 +1,11 @@
...
+//
+// The files are published for IDEs and validated by the unit tests of the
+// packages that parse the corresponding configuration; the schemas werf
+// itself validates against at runtime are embedded here.
+package schemas
+
+import _ "embed"
+
+//go:embed werf-giterminism.json
+var GiterminismConfig []byte
```
alexey-igrychev: `[]byte` is an exported mutable package-level slice — any caller can scribble on the embedded schema. `//go:embed` into a `string` is immutable and works unchanged here, since `json.Unmarshal` takes `[]byte(GiterminismConfig)`.

alexey-igrychev (commented):
## Review

Reviewed against `origin/3` at `8c9b969`. Ran `task build`, `task test:unit` (Config Suite 143/145, Schemas Suite 19/19), `task lint:golangci-lint` (0 issues) on a worktree at the PR head.

### What I verified

- **Coverage is complete.** I walked every `yaml:` tag of every `raw*` struct in `pkg/config` (`raw_meta*`, `raw_stapel_image`, `raw_image_from_dockerfile`, `raw_git*`, `raw_export_base`, `raw_import`, `raw_mount`, `raw_shell`, `raw_secrets`, `raw_dependency*`, `raw_image_spec`, `raw_stage_dependencies`) plus `git_repo.BasicAuthCredentials`/`PasswordSource` against `schemas/werf.json`. No missing and no extra keys.
- **The giterminism move is behavior-neutral.** I parsed the old `schemaYaml` constant from `origin/3:pkg/giterminism_manager/config/openapi_spec.go` and the new `schemas/werf-giterminism.json` and diffed the trees with `$schema`/`title`/`description` stripped: identical.
- **The `//go:embed` is really wired in.** Replacing `cli` with `{"type":"string"}` in `schemas/werf-giterminism.json` fails `TestValidFullConfig` and `TestUnknownNestedKeyRejected/cli`.
- **The tests discriminate.** Four mutations, each caught by the right test: deleting `staged` from `dockerfileImage` (fails the fixture corpus on `test/e2e/build/_fixtures/copy_flags/werf.yaml` and the accept table), deleting `gitMapping.dependencies` (fails `local git mapping with branch`), deleting `pattern: ^/` from `werf-includes.add` (fails `relative add`), corrupting the giterminism schema (above).
- **`cacheVersion: 1` as a number** — correct, yaml.v2 assigns the raw scalar text to a string field regardless of the resolved type.
- **Publishing.** nginx `location ~* ^/docs/(?<ver>v\d+…)/(?<tail>.*)` → `try_files /$tail` maps `/docs/v3/schemas/werf.json` to `/schemas/werf.json`; Jekyll copies `schemas/*.json` as static files (not in `exclude`). `check_broken_links` is green and that job actually builds and runs the `assets` image, so the new git mapping does not break the docs build.
- **The IDE claims hold.** `deprecationMessage` is honored by yaml-language-server (`yamlCompletion.ts`) and by JetBrains (Inspectopedia `YAMLSchemaDeprecation` explicitly names the non-standard extension). A bare `"werf.yaml"` pattern in `yaml.schemas` does match: `vscode-json-languageservice/jsonSchemaService.ts` prepends `**/`. Multi-document YAML is validated per document. The `# yaml-language-server: $schema=<https url>` modeline is current syntax.

### Blocking

**1. Empty strings bypass every mutual-exclusion rule.** The parser tests `!= ""`, the schema tests for key presence. All three of these pass the schema and fail in werf (checked with `santhosh-tekuri/jsonschema/v6`):

```yaml
image: a
from: ""                # schema OK -> "`from: IMAGE` required!"
```
```yaml
includes:
- git: https://x/y
  branch: ""            # schema OK -> "specify only `branch` or `tag` or `commit`"
  add: /
  to: /
```
```yaml
basicAuth: {username: bot, password: {env: ""}}   # schema OK -> "specify only env or src or plain"
```

Same class for `secret.env/src/value`, `import.from/image`, `mount.from/fromPath`, `dependencies.from/image`, `cleanup.references.branch/tag`. Uniform fix: `"minLength": 1` on every string that participates in a `oneOf`, plus reject-table entries — none of the current tables cover an empty value.

**2. A schema-only PR runs none of these tests.** This PR adds `schemas/**` to `docs_deploy_latest.yml` and `docs_validate.yml` but not to the `workflow_proceed` paths filter in `.github/workflows/tests.yml:30-42` (`go.mod`, `Taskfile.dist.yaml`, `.github/**`, `cmd/werf/**`, `pkg/**`, `scripts/**`, `test/**`). `docs_validate.yml` only runs `doc_unit` with `packages: cmd/werf/docs`. So editing `schemas/werf.json` alone goes green without executing either the Config Suite or the Schemas Suite — exactly the drift guard this PR builds. One line fixes it.

### PR description

It becomes the commit body, so two claims need adjusting:

- "Mutually exclusive directives fail when combined" — only for non-empty values (see 1).
- "a directive added to the parser and used in a fixture without a schema update fails `task test:unit`" — true locally (verified by mutation), but in CI only when the PR also touches `pkg/**` or `test/**` (see 2).
- The list of enforced exclusions omits `contextAddFile`/`contextAddFiles`, while the schema's own `description` claims it is enforced.

### Not verified

- `https://werf.io/docs/v3/schemas/werf.json` returning 200 — currently a 302 to `/docs/v3/usage/project_configuration/overview.html` since the file is not deployed yet. The route is correct on paper; the 200 depends on the production deploy.
- Live behavior in VS Code / GoLand — checked against language-server sources and JetBrains docs, not by hand.
- A local docs image build — relying on the green `check_broken_links` job, which performs one.

schemas/werf.json
Fral738: Done in 91493ed6c: `minLength: 1` on every string a `oneOf` discriminates on (`from`/`fromImage`, `secret.env/src/value`, `import.from/image`, `mount.fromPath`, `dependencyBase.from/image`, `basicAuth.password.env/src/value`, and `regexpOrString` for `references.branch/tag`), plus a reject entry per group. Stripping all `minLength` fails exactly those 9 entries.

schemas/werf.json
Fral738: Done: `"not": {"required": ["contextAddFile", "contextAddFiles"]}` on `dockerfileImage` plus `dockerfile image with both contextAddFile and contextAddFiles` in the reject table.

schemas/werf.json
Fral738: Done: root is now `allOf` of `if: required[configVersion|dockerfile|image] → then: $ref`, with the `image` branch guarded by `not: anyOf[configVersion, dockerfile]` so `dockerfile` wins over `image` as in `parser.go`, plus a final `anyOf` so a document with none of the three keys still fails. A test in `schemas/` asserts the `shel:` typo yields exactly one `additional properties` error and no mention of `configVersion`/`dockerfile`; restoring the flat `oneOf` fails it.

schemas/werf.json
Fral738: Done: all 39 links now point at `werf.io/docs/latest/...`. Checked that `/docs/latest/reference/werf_yaml.html` and `/docs/latest/usage/build/stapel/base.html` answer 200 on the first hop, so `#anchor` fragments survive.

schemas/werf-includes.json
Fral738: Done: `minLength: 1` on `branch`, `tag`, `commit` + `include with empty branch` reject entry.

schemas/werf-includes.json
Fral738: Done: `minLength: 1` on `env`, `src`, `value` + `basicAuth password with empty source` reject entry.

.github/workflows/docs_validate.yml
Fral738: Done: `schemas/**` added to `workflow_proceed` in `tests.yml`.

pkg/config/helpers_test.go
Fral738: Done: the corpus now walks the repository root (skipping dot-directories) and asserts `stapel/werf.yaml` is in it — 87 of 108 files, 21 templated ones skipped. Both figures are in the PR description now.

schemas/schemas.go
Fral738: Done: `//go:embed` into a `string`; the call site wraps with `[]byte(...)`.

alexey-igrychev:
## Verification

- Local: `task format`, `task build`, `task lint`, and unscoped `task test:unit` passed on `90db1a121` (52 suites); `task doc:check-broken-links:ru` passed after the parallel docs run hit local Docker `ENOSPC`.
- Mutation: allowing unknown properties in a nested schema object, reverting numeric scalar support, removing remote Git ref exclusivity, removing dependency target `minLength`, and removing absolute-path constraints each made its focused test fail.
- Independent review found remote Git ref parity, stale `password.file` docs, remaining scalar coercions, path validation, and accept-table parser coverage; all findings were addressed in `f904b59e6..90db1a121`.

## Review focus

- `schemas/werf.json`: deliberate editor-only strictness is listed in the PR description; parser/schema parity tests keep those exceptions separate.

alexey-igrychev (approved):
Schema/parser parity, documentation, CI wiring, and regression-test mutations reviewed. All findings addressed; local format/build/lint/unit/docs checks pass. The remaining integration_main failure is the existing origin/3 host-purge regression tracked by #7885.

alexey-igrychev:
Merged as `25bfb25d61f08a4dd42ed4e6a639b56b81864929`. Production docs deployment passed for EU and RU; all six schema URLs return HTTP 200 with valid JSON. The overridden `integration_main` check failed only on the existing origin/3 host-purge regression tracked and fixed separately in #7885.

PR #7863, opened 2026-09-16
fix(build, cleanup): keep the parallel log streaming and its progress numbers in order

Fral738:
## Summary

During a parallel `werf build`, the log could go completely silent for minutes while images were still building, then dump every other worker's output at once at the end with the `(N/Total)` progress numbers out of order. It happens whenever the first worker runs out of images before the others do — e.g. it finishes its share and the remaining images all depend on something a slower worker is still building. Reproduces from a clean checkout with `--parallel-tasks-limit=3`:

```yaml
project: parallel-log
configVersion: 1
---
image: a-short
from: alpine:3.20
shell:
  install: [ 'for i in $(seq 1 5); do echo "a tick $i"; sleep 1; done', 'echo a > /a.txt' ]
---
image: b-long
from: alpine:3.20
shell:
  install: [ 'for i in $(seq 1 40); do echo "b tick $i"; sleep 1; done', 'echo b > /b.txt' ]
---
image: c-long
from: alpine:3.20
shell:
  install: [ 'for i in $(seq 1 40); do echo "c tick $i"; sleep 1; done', 'echo c > /c.txt' ]
---
image: d-after-a
from: alpine:3.20
import: [ { image: a-short, add: /a.txt, to: /a.txt, after: install } ]
shell:
  install: [ 'for i in $(seq 1 5); do echo "d tick $i"; sleep 1; done' ]
---
image: e-after-bc
from: alpine:3.20
import:
- { image: b-long, add: /b.txt, to: /b.txt, after: install }
- { image: c-long, add: /c.txt, to: /c.txt, after: install }
shell:
  install: [ 'for i in $(seq 1 5); do echo "e tick $i"; sleep 1; done' ]
```

## What

- The oldest still-running image is streamed live, and the ones started after it follow in start order once it finishes; the printer is never again parked on a worker that merely sits idle. The log is quiet only while that oldest image itself produces no output. Under the repro above the log used to go quiet for ~30 s after `d-after-a` finished; now `b-long` streams its ticks right away.
- The `(N/Total)` prefix on every image block and in `Build summary` is the position in which the image actually started building, and image blocks are printed in that order, so the numbers appear ascending. Before, `(2/5)` and `(3/5)` could come after `(5/5)`.
- Each image's output is still one uninterrupted block; lines from concurrently building images are never interleaved. A goroutine that outlives its image and keeps logging through the old context is dropped, not printed inside the next image's block on the same worker. (This covers the image's own logger; output such a goroutine pushes through the worker's docker cli is attributed to whatever image that worker is building at that moment, as it always was.)
- Output of the docker cli (Docker backend: `docker pull`, `docker run` of Stapel stages, and so on) carries the same block indentation as the image's own log lines and stays inside that image's block. Every byte the image's buffer accepted before it closed is printed inside that block, and the block always ends on a line boundary: a last line without a trailing newline is terminated, never glued onto the next block. A write racing with the close — output of a goroutine the image left running and writing through the image's logger — is either terminated inside the image's block or dropped; it is never attributed to another image. Bytes still held back by the image's logger at that moment (an unfinished line written after the final flush) are the ones dropped.
- Task output wraps at the same width as before: the width available inside an image block equals what a sub-logger derived directly from `werf build`'s logger would have had (the caller's indentation is subtracted once, not once per clone).
- Each worker claims its first image only after the previous worker (0, 1, 2, …) has claimed its own, so the first block of each worker and its number do not depend on goroutine scheduling. Which image a worker picks up after its first one is decided by real timing, so from there on the order can differ between runs.
- When an image fails, its block is still printed last; when the failing image is the one being streamed, the blocks of images that were canceled behind it are dropped. Relative order among the surviving blocks is now start order, not worker order. When the run is stopped by an error that belongs to no image (a scheduling error, a failure to allocate the next log buffer), the queue is printed as is: no innocent image's block is moved to the end or dropped in its place.
- A log write of the streamed image never waits for the printer's disk reads: the buffer is read outside its lock, as the per-worker buffers were.
- Same rules apply to every other parallel operation that shares this printer (stage export, manifest operations, cleanup). `werf cleanup`'s per-image `(N/Total)` and the multiplatform manifest `(N/Total)` in `werf build` are numbered by start order too, so they stay ascending; before this change they would have read `1/4, 3/4, 2/4, 4/4` under two workers. VERIFIED: `werf cleanup --parallel-tasks-limit=3` over a 6-image project in a local registry prints `(1/6)` … `(6/6)` in that order; with task-ID numbering the static partition would have printed `1, 3, 5, 2, 4, 6`.
- The Buildah backend behaves the same as Docker: VERIFIED with the repro build under `WERF_BUILDAH_MODE=auto` on a Linux host — the long images stream tick by tick, blocks come out `1..5`, no line inside a block loses its prefix, buffer temp files go 3 → 1 → 0 during the run.
- Open file descriptors used for buffering are bounded by the number of workers plus one, not by the number of tasks: a finished task waiting to be printed holds none, and a task whose output the printer had already fully read when it finished holds none either. VERIFIED: 256 one-line tasks on a single worker under `RLIMIT_NOFILE=128` complete; the first revision of this branch died at task 114 with `too many open files`.
- Temp files in werf's tmp dir (`--tmp-dir` / `$WERF_TMP_DIR`) used for buffering are now `parallel-worker-<pid>-<worker>-<n>-*.log`, one per task instead of one per worker, and each is removed as soon as its block has been printed, not at the end of the run. At any moment the disk holds only output not printed yet — the running images plus the finished ones still queued behind the one being streamed — instead of the log of the whole build. VERIFIED: sampling the tmp dir every 5 s during the repro build shows 3 files while 5 images build, 0 afterwards; before, all 5 stayed until the end.
- Not changed: how many images build concurrently, in what order they are scheduled, `--parallel` / `--parallel-tasks-limit` semantics, one docker cli per worker. The unit suite of `pkg/util/parallel` is race-free under `-race`.

## Why

The parallel printer buffered output per worker and drained workers strictly in ID order: worker 0 was live, workers 1..N wrote to temp files that were only replayed after worker 0 had closed. With the dependency-graph scheduler introduced in v3, a worker that has nothing runnable blocks waiting for a dependency rather than exiting, so worker 0 could sit idle for the rest of the build with the printer still parked on it. In v2 this was masked because each level was a separate parallel run and buffers were flushed at every level boundary; in v3 the whole build is one run, so the buffers grow for its entire duration. For a CI job this looks like a hang, and some CI systems kill jobs on output inactivity.

The alternative — emitting a heartbeat or progress line from the scheduler into the main stream while worker 0 idles — would keep the log alive but leave the per-worker dump and the scattered numbering in place, and the heartbeat would land in the middle of whatever block worker 0 last printed. Buffering per task and draining in start order fixes the silence and the ordering with one mechanism. Binding the docker cli per task as well was rejected: each cli carries its own API client and connection pool, and cleanup runs thousands of tasks.


Fral738:
## Verification

- Real build of the repro `werf.yaml` from the description with `--parallel-tasks-limit=3`, cold stage cache and `alpine:3.20` removed first so `docker pull` runs through the cli relay, Docker backend on macOS, every stdout line stamped with its arrival time: `b-long` streams tick by tick right after `a-short` finishes; the only gap longer than 1 s is the 4 s `alpine` pull. Zero lines inside image blocks without a block prefix (`│`/`┌`/`└`). Numbers `1..5` ascending. No `parallel-worker-*` temp files remain in the tmp dir afterwards. On `3` the same run has a 29 s silence after `d-after-a` and 156 lines dumped in one second at the end.
- Descriptor bound: throwaway spec setting `RLIMIT_NOFILE=128` and running 256 one-line tasks on one worker via `parallel.DoTasks` passes; the first review reported it failing at task 114 on `c436196`. Spec not kept — it changes the process rlimit.
- Mutation: one buffer per worker again and half-close only at worker exit (the old behavior) → `DoTasksDynamic streams a running task's output while another worker idles waiting for a dependent task` fails with `context deadline exceeded`; `parallel task should print log concurrently…` fails on block order.
- Mutation: `Printer.FailFast` not moving the failed task to the end → the fail-fast spec fails (order `[0, 1, 2, 3]` instead of `[0, 2, 3, 1]`).
- Mutation: the first-task gate replaced with a skew that lets higher-ID workers start first → the same fail-fast spec fails (order `[3, 0, 2, 1]`).
- Mutation: `TaskOutput` nils `writer`/`reader` but skips the real `Close()` calls (the second review's mutation) → all three `TaskOutput descriptor lifecycle` specs fail on `Stat` not returning `os.ErrClosed`.
- Mutation: `HalfClose` no longer releases an already-drained reader → `releases a reader that already drained everything when the task then finishes` fails.
- Mutation: task logger bound to the worker instead of the task's own buffer → `drops a finished task's late output…` fails with `late output from task A` inside B's block.
- Mutation: cli relay pointed at the raw `TaskOutput` instead of the task logger's streams → `worker-level stream relay` fails: `cli line in A` loses its `│` prefix (the second review's symptom).
- Race: `task test:unit paths="./pkg/util/parallel/..." mode=cgo cgoLDFlags='-w' parallel=false -- -race` is clean on this revision. Mutation: task loggers cloned from the caller's logger instead of the per-worker template → the same run reports `WARNING: DATA RACE` (the third review's finding).
- Mutation: `terminateTaskOutput` made a no-op → both entries of `terminates a line the cli left unfinished…` fail with the partial line lost (`\nB line\n`); with only the terminating newline removed they fail with `\nA partialB line\n`.
- Mutation: the newline termination in `TaskOutput.HalfClose` disabled → `TaskOutput.HalfClose leaves the output on a line boundary` (`a\nlate` instead of `a\nlate\n`), both `terminates a line…` entries (`\nA partialB line\n`) and the UTF-8 `flushes a trailing incomplete rune` spec fail. The `a\n` + `late` entry is the fourth review's interleaving — a write landing after the logger flush and before half-close — expressed at the level where the fix lives.
- Mutation: `Unlock → runtime.Gosched → Lock` inserted between appending the terminator and closing the writer inside `HalfClose` (the fifth review's mutation) → `TaskOutput.HalfClose keeps the line boundary against a racing writer` fails on iteration 0 with `a\nxxxxxxxxxxxxxxx` lacking the suffix. A bare `Unlock → Lock` with no yield survived one run of the same spec: it is a stress test over 200 iterations, not a proof over every interleaving.
- Mutation: `SetWidth` correction on the task logger removed → `task loggers keep the width a directly derived sub-logger would have had` fails (`76` instead of `78` for a parent of width 80 under one indent).
- Mutation: image log index from a 1-based local counter instead of `parallel.TaskStartOrder` → `TestDoImagesInParallel_AssignsBuildOrderIndexByRealDequeueNotStaticTopology` fails (`expected 1, actual 2` for "slow"). A 0-based local counter is NOT caught, and no fixture can make it fail deterministically: it diverges from the start order only when two workers race inside the few instructions between `printer.Enqueue` and entering the task closure, and there is no seam there a test can hold open without adding a test-only hook to `runWorkers`. The assertion pins the numbering source to the printer's position, which is what makes the race impossible by construction; it does not pretend to observe the race.
- Mutation: the early `out.Cleanup()` in `Printer.Print` removed → `Printer removes a task's temp file as soon as its block is printed…` fails with `A's temp file … still exists while B runs: context deadline exceeded`.
- Temp-file lifetime on the real build: `ls $TMPDIR | grep -c parallel-worker` sampled every 5 s during the repro build with 3 workers reads `3` throughout and `0` after; on the previous revision it read `5` until the end.
- Buildah: the same repro build on a Linux host (Ubuntu 22.04, buildah 1.23.1, rootless, `WERF_BUILDAH_MODE=auto`, `--repo` on a local `registry:2`, `--parallel-tasks-limit=3`), CGO binary from this branch (`task build:dev:linux:amd64:cgo`, verified by code markers), every line stamped on arrival: `b-long` ticks arrive one per second while `a-short` has finished, longest silence 3 s, blocks `(1/5)`…`(5/5)` in order, 0 unprefixed lines inside blocks, `parallel-worker-*` files in `$TMPDIR` sampled every 5 s: `3` while building, `1`, then `0`.
- `werf cleanup` on the same host: 6-image project built into the local registry, `cleanup --without-kube --parallel-tasks-limit=3 --save-cleanup-report`: the `Cleanup unused stages` block prints `(1/6) image …` through `(6/6) image …` in ascending order, exit 0, no temp files left.

## Review focus

- `runWorkers` in `pkg/util/parallel/parallel.go`: the first-task gate (`firstTaskStarted`) is released in the goroutine's `defer` as well as on first `Enqueue`, so a worker that never takes a task or is canceled before its first `next` still unblocks the next one; this is the piece that turns a data race into a deadlock if it regresses.
- `TaskOutput.releaseDrainedReader` in `pkg/util/parallel/task_output.go` is the single place a reader is closed and is called from both `Read` and `HalfClose`; `Close()` covers a drain interrupted mid-way. All under the one mutex shared with `Write`.
- `Worker` in `pkg/util/parallel/worker.go` no longer owns a logger: it relays the worker-level docker cli's streams to the current task logger's streams (`bindTaskStreams` / `endTask`). `docker.NewContextWithStreams` exists for this; `docker.NewContext` is unchanged for the other callers.
- `TaskOutput.HalfClose` in `pkg/util/parallel/task_output.go` now also terminates an unfinished last line, under the mutex that stops writes. The task logger's `LogF("")` flush in `runWorkers` runs before it, outside the lock; a write between the two is terminated by `HalfClose`, a write after it is dropped, and a partial line the logger caches after the flush stays in that logger (lost, not glued) — the invariant is "the block ends on a line boundary", not "every late byte is printed".
- `runWorkers` relies on two logboek facts: `LogF("")` flushes a held-back incomplete line (it formats with `cacheIncompleteLine=false`), and all level streams of one logger share the fitter state, so one flush covers stdout and stderr. Both are exercised by the `terminates a line…` table; if logboek changes either, that table is what breaks. Third fact: `NewSubLogger` sets the clone's width to the parent's content width — hence the `SetWidth` after cloning from the template, pinned by the width spec.
- Every caller of `parallel.DoTasks` (storage manager, cleanup, export phases) now gets start-ordered output; `pkg/cleaning/cleanup.go` and `pkg/build/build_phase.go` switched their progress index to `parallel.TaskStartOrder` for that reason.
- `TaskOutput.Cleanup` is now idempotent and is called twice per printed task: by `Printer.Print` right after the block is emitted, and by the final sweep in `runWorkers`. Outputs the printer never reaches — those truncated by `FailFast` behind a failed head — are removed only by the sweep; that path is exercised by the fail-fast specs, which run to completion with a clean tmp dir.

alexey-igrychev:
Reviewed the branch (head `73a27ca54` at the time) and pushed three follow-up commits — `0298098b2`, `ea9761ee0`, `95a8b9876` — fixing what the review found. Verified: repro build from the description on the Docker backend (`--parallel-tasks-limit=3` — blocks come out `(1/5)…(5/5)`, longest silence 9 s, temp files 3 → 0), a failing build (`exit 7` in an image — failed block printed last, exit 1, no leftover temp files), `pkg/util/parallel` under `-race` (34/34), and the fixes below by mutation (each reverted piece makes a test fail).

**`0298098b2` — the failed-task highlight could grab an innocent image's log.** `printer.FailFast(workers[workerErr.ID].Output())` ran after `g.Wait()`, i.e. always outside a task, and `endTask` never cleared `Worker.current` — so `Output()` returned the worker's *last* task, despite its doc saying "nil outside a task". For an error raised by a task this happened to be the right output; for an error raised *between* tasks (`beginTask` failing to allocate a buffer, a non-cancel error from `NextTaskFunc`, a deadline-exceeded that `isCanceledErr` doesn't treat as cancel) it was the worker's previous, successful task — which then got moved to the end of the log or, if it was the head, silently truncated every queued block behind it. The failed output is now recorded at the moment `taskFunc` returns an error, and an error that belongs to no task leaves the queue alone (new spec in `dynamic_test.go` covers it; with the old wiring it prints `a, c, b` instead of `a, b, c`).

**`ea9761ee0` — log writes stalled on the printer's disk reads.** `TaskOutput.Read` held the buffer mutex across `os.Open`/`ReadAt`, so every `Write` of the streamed task (and every relayed docker-cli write) contended with the printer's read loop, which spins without sleeping while output flows. The base code read lock-free; the `ReadAt` now runs outside the lock. Also: a read landing exactly on the end of a drained task no longer reopens the file just to close it.

**`95a8b9876` — the integration tests raced wall-clock margins.** `TestDoImagesInParallel_DependentImageDoesNotWaitForUnrelatedSlowImage` failed with `"3" is not less than 0` in up to a third of `task test:unit paths="./pkg/build/... ./pkg/util/... ./pkg/cleaning/... ./pkg/docker/..."` runs on a loaded machine (the per-task buffer/logger setup this PR adds per image is enough to eat the 150 ms margin); `…AnnotatesEachImageWithARealWorkerID`'s "both workers used" can lose the same race. Both are now gated on channels — a wave-barrier regression deadlocks into the test timeout instead of racing a margin. Interleaved A/B on the same machine after the fix: head 0/10 failed runs vs base `3` 1/10. Also retitled the relay spec (it never covered "drops what a finished task left behind" — a leaked goroutine writing through the worker's docker cli lands in the current image's block, which the `Worker` doc now states) and downgraded the start-order assertion's comment to the wiring check it actually is: both sides read the same context value, so it cannot fail on its own (verified — reverting `conveyor.go` to the old `atomic.Int64` counter keeps the whole suite green).

Notes, no action taken:
- The printer streams the *oldest running* image; if that image is silent (long `docker pull`, quiet `RUN`) the log is quiet even though other images are producing output. Inherent to coherent per-image blocks, same as the pre-v3 behavior within a worker — but worth knowing when reading CI logs. Description adjusted accordingly.
- `NewTaskOutput` creates the temp file in the constructor (CODESTYLE says constructors do no filesystem I/O) — inherited from `NewWorker`, left as is.
- `pkg/build` has a residual ~1/10 flake that exists on base `3` as well (invisible in ginkgo's parallel mode: the suite fails with specs 30/30 green — it's a plain `go test` failing in one of the parallel procs); not introduced here.

Not verified on this machine: Buildah backend (macOS), `werf cleanup` against a real registry, `golangci-lint` (not installed here).

alexey-igrychev:
Reviewed the branch at `9555152de` (two independent review passes over the whole diff on top of my own) and pushed two follow-up commits — `db2cde77a`, `5a1025071`. Verified: `go build ./...`, `go test ./pkg/util/parallel/... -race -count=1` (37/37), `./pkg/build/... ./pkg/cleaning/... -race` — all specs green. `TestBuild` reports a data race in `pkg/build/build_phase_test.go:28` vs `:76`; it reproduces on base `9f52c5cee`, so it is not from this PR.

**`db2cde77a` — a temp log file was leaked whenever its removal failed.** `TaskOutput.Cleanup` set `removed = true` before `os.Remove`, so a failed removal was final: `Printer.Print` only warns about it, and the final sweep in `runWorkers` then saw the output as already removed. The flag is now set after a successful removal (`os.ErrNotExist` counts as success), so the sweep retries. New spec in `task_output_internal_test.go` — with the old ordering it fails, leaving the file on disk after the second `Cleanup`.

**`5a1025071` — wording only.** A buffer that could not be created or closed costs the task's log, not the task, but the warning read `failed to half-close task N output`, i.e. as a shutdown failure. And `Image.SetWorkerID`'s doc still said the printer "prints one worker's whole output before moving to the next", which stopped being true with per-task blocks in start order.

Two things left for you to decide, no action taken:

- `Printer.FailFast` (`printer.go:73-74`): when the failed task is the one being printed, everything queued behind it is dropped as "partial output is noise". With per-task buffering the tasks behind the head are not necessarily partial — the head is typically the slowest task, and tasks that started later may have finished successfully with a complete block ("Use previously built image" and the like). Those blocks are now discarded too. Not a regression (per-worker buffering discarded the same output), but more likely to bite. Keeping the blocks that are already half-closed and dropping only the still-running ones would be closer to the stated intent.
- `AfterImages` in `build_phase.go`: only the multiplatform branch was moved to `parallel.TaskStartOrder`; the single-platform branch still labels via `img.LogDetailedName()`, i.e. with the *build phase's* start order, while `AfterImages` runs its own `DoTasks` with its own start order. So `(N/Total)` in the final-repo copy blocks can still come out non-ascending. Pre-existing, and out of scope if you want to keep this PR to the build log itself.

PR #7864, opened 2026-09-17
fix(cleanup): stop import metadata cleanup leaking goroutines

### LINKED ISSUE #7862: cleanup leaks protected-stage sets through abandoned Iter goroutines (29 GiB live heap, 68k stages)
# cleanup leaks protected-stage sets through abandoned Iter goroutines (29 GiB live heap, 68k stages)

## Version

Observed on werf v2.76.0, tag commit `9ff07f6887845e4963c3bbc2f0b4aabd25a31946`.
The dependency is `github.com/deckarep/golang-set/v2 v2.6.0`.

As of 2026-09-16 the same loop and dependency are still present in v2.78.1
(`3c7a1b8921c2626162f8c7f31b23b92197fdc331`) and v2.78.2
(`ce7c1a0471efc744de4045c1d3010a791992d149`, marked prerelease).
The cleanup changes since v2.76.0 add report plumbing (#7806), not an iterator fix.
Issue searches for cleanup memory leaks, goroutines, and import metadata did not
find a matching report.

## Result

A Stapel project with imports has 68,538 stage tags. A successful cleanup takes
about 79 minutes and deletes zero stages:

```text
ancestor                 47106
import source            21275
used in Kubernetes          83
git policy                  59
built within last 2 hours   15
Saved stages tags (68538/68538)
```

The runner normally has about 31 GiB RAM and no swap; the job makes the shared
host unresponsive. A temporary ~63 GiB VM allowed collecting a complete log.
`GODEBUG=gctrace=1` reveals that the graph itself is not the tens-of-GiB live set:

```text
Fetching imports metadata: 514.07 seconds
Processing relative stages for saved stages: 2501.84 seconds
gc 57 @3081.995s: 455->455->424 MB
Saved stages tags (68538/68538)
gc 75 @3341.247s: 5436->5463->3094 MB
gc 87 @3824.937s: 21183->21281->12036 MB
gc 95 @4709.521s: 45269->46182->28980 MB
Cleanup unused stages: 4569.47 seconds
Running time: 4753.35 seconds
```

The large heap growth starts **after** logging all saved stages, during the
subsequent import metadata scan. No production heap profile is available;
the source-level mechanism below has been reproduced independently.

## Source-level cause

In [deleteUnusedImportsMetadata](https://github.com/werf/werf/blob/v2.76.0/pkg/cleaning/cleanup.go#L1043),
each source stage reconstructs the complete protected set:

```go
outerLoop:
for sourceStageID, importMetadataIDs := range m.sourceStageIDImportIDs {
    for protectedStageDesc := range m.stageManager.GetProtectedStageDescSet().Iter() {
        if sourceStageID == protectedStageDesc.StageID.String() {
            continue outerLoop
        }
    }
    importMetadataIDsToDelete = append(importMetadataIDsToDelete, importMetadataIDs...)
}
```

`GetProtectedStageDescSet` creates sets and unions them.
The dependency's [threadSafeSet.Iter](https://github.com/deckarep/golang-set/blob/v2.6.0/threadsafe.go#L206)
starts a goroutine sending values into an unbuffered channel. An early
`continue outerLoop` abandons the receiver. The producer remains blocked,
holding the newly allocated set and an RLock.

This retains roughly O(protected stages × protected import sources) set storage,
in addition to quadratic allocation/work. GC cannot free those reachable sets.

There are similar early exits from `.Iter()` in
[protectRelativeStageDescSetByStageDesc](https://github.com/werf/werf/blob/v2.76.0/pkg/cleaning/cleanup.go#L1170)
(legacy parent lookup and image-index lookup) and `cleanupFinalStages`.
They should be audited as well. The legacy parent lookup additionally scans
the complete stage set for every visited stage.

## Minimal reproduction without a registry

Use a separate directory with this `go.mod`:

```go
module cleanup-iter-repro

go 1.23

require github.com/deckarep/golang-set/v2 v2.6.0
```

And `main.go`:

```go
package main

import (
    "flag"
    "fmt"
    "runtime"

    mapset "github.com/deckarep/golang-set/v2"
)

func main() {
    indexed := flag.Bool("indexed", false, "use one index instead of abandoning Iter")
    flag.Parse()
    protected := mapset.NewSet[int]()
    for i := 0; i < 4096; i++ {
        protected.Add(i)
    }
    runtime.GC()
    beforeGoroutines := runtime.NumGoroutine()
    var before, after runtime.MemStats
    runtime.ReadMemStats(&before)
    if *indexed {
        ids := make(map[int]struct{}, protected.Cardinality())
        for stage := range protected.Iter() {
            ids[stage] = struct{}{}
        }
        for source := 0; source < 1000; source++ {
            if _, ok := ids[source]; !ok {
                panic("unexpected unprotected source")
            }
        }
    } else {
    outer:
        for source := 0; source < 1000; source++ {
            for stage := range protected.Clone().Iter() {
                if stage == source {
                    continue outer
                }
            }
        }
    }
    runtime.GC()
    runtime.Gosched()
    runtime.ReadMemStats(&after)
    fmt.Printf("goroutine_delta=%d heap_before=%.2fMiB heap_after=%.2fMiB\n",
        runtime.NumGoroutine()-beforeGoroutines,
        float64(before.HeapAlloc)/(1<<20), float64(after.HeapAlloc)/(1<<20))
    runtime.KeepAlive(protected)
}
```

Run separately:

```sh
go mod tidy
go run .
go run . -indexed
```

The equivalent parameterized reproducer on Go 1.25.5 darwin/arm64 produced:

| Protected stages | Source checks | Mode | Goroutine delta | Heap after GC |
| ---: | ---: | --- | ---: | ---: |
| 4096 | 1000 | abandoned Iter | 1000 | 141.92 MiB |
| 2048 | 1000 | abandoned Iter | 1000 | 71.39 MiB |
| 4096 | 500 | abandoned Iter | 500 | 71.11 MiB |
| 4096 | 1000 | indexed | 0 | 0.27 MiB |

Exact figures vary by Go version and iteration order. This is a reduced
reproduction of the iteration pattern, not a full werf integration benchmark.

## Expected result / suggested correction

Construct an index of protected stage IDs once, then check membership for each
source stage. Preserve existing deletion semantics. Use indexed parent lookups
and safe, fully consumed iterations (or cancellable iterators) elsewhere.
Replacing just `.Iter()` with `.ToSlice()` prevents that goroutine leak but
still recreates the full set for every source.

Regression coverage should verify equivalent deletion candidates, no growing
goroutine count after repeated scans, and memory scaling with the number of
stages plus imports rather than their product.

## Related retention behaviour / workaround question

Independently of the leak, the default resolver maps a checksum to **all** source
stage IDs with that imported-content checksum, then protects their ancestors.
The existing `should keep both by import checksum` e2e test documents this.
Our project imports small, potentially unchanged subtrees from frequently rebuilt
intermediate images, which can make this closure retain almost everything.
The exact high-fanout checksum has not been measured in our registry yet.

`WERF_EXPERIMENTAL_IMPORT_BY_SOURCE_IMAGE_TAG=true` from #7392 appears to avoid
this fanout and skip `deleteUnusedImportsMetadata` entirely. Current default
builds already write both checksum and source-stage-ID labels.

Could maintainers confirm the supported migration procedure for using this flag
**only during cleanup** against older default-mode builds, especially:

- legacy stages missing direct source labels;
- retained import metadata pointing at stages removed/rebuilt by later runs;
- how to retire old import metadata, since the flag skips its cleanup;
- whether a separate cleanup-only direct-source policy is appropriate?

We have not enabled the flag in production, changed retention policies, or
purged the registry. Fixing the leak alone does not fix checksum-based
over-retention; both need to be addressed for effective cleanup.

pujitha24:
Motivation:
deleteUnusedImportsMetadata decided whether each source stage was
still protected by ranging over stageManager.GetProtectedStageDescSet().Iter()
and jumping out with `continue outerLoop` as soon as it found a match.
GetProtectedStageDescSet() rebuilds a brand-new unioned set on every
call (not cached), so this ran that O(n) rebuild once per source stage,
turning the whole function into O(protected stages x source stages).
Worse, the underlying set's Iter() (github.com/deckarep/golang-set/v2)
starts a goroutine that RLocks the set and blocks forever on an
unbuffered channel send once nobody keeps draining it; abandoning the
range early via `continue outerLoop` does exactly that, leaking one
goroutine plus its pinned set object per early match. On a project
with many protected stages and many import sources, this accumulates
into unbounded goroutine and heap growth during cleanup.

Approach:
Build a map of protected stage IDs once, via ToSlice() (synchronous,
no goroutine involved), before iterating the source stages, then use
a plain map lookup for membership instead of re-scanning and
abandoning the protected set on every source stage. Deletion semantics
are unchanged: a source stage is still skipped exactly when its ID is
in the protected set.

Validation:
- go test ./pkg/cleaning/ -run TestDeleteUnusedImportsMetadata -v:
  added TestDeleteUnusedImportsMetadata_SkipsProtectedKeepsUnprotected
  (deletion candidates still match protected/unprotected status) and
  TestDeleteUnusedImportsMetadata_NoGoroutineLeak (asserts
  runtime.NumGoroutine() does not grow across a scan with many
  protected/matching stages). Confirmed both fail against the
  pre-fix code (the leak test fails with ~200 extra goroutines) and
  pass against the fix, i.e. this is a real regression test, not a
  tautology.
- task build and task test:unit paths="./pkg/cleaning/..." both pass.
- task lint:golangci-lint golangciPaths="./pkg/cleaning/..." reports
  pre-existing "undefined: git"/Ginkgo-DSL typecheck errors in
  test/legacy_e2e and pkg/cleaning/git_history_based_cleanup that
  reproduce identically on an unmodified checkout of the base branch;
  this is an environment/tooling issue unrelated to this change, not
  a regression it introduces.
- Not reproduced here: the full production scenario from the report
  (tens of thousands of stages, tens of GiB of heap growth). The unit
  test demonstrates the same leak mechanism (an abandoned Iter() goroutine
  per early match) at a scale that fits a fast local test.

Report: https://github.com/werf/werf/issues/7862
Signed-off-by: Pujitha Paladugu <10557236+pujitha24@users.noreply.github.com>
Assisted-by: claude-sonnet-5 (via Claude Code)

Fixes #7862


alexey-igrychev (changes requested):
The fix itself is correct and minimal, and I verified the tests are load-bearing rather than tautological: reverting the implementation to the pre-fix body makes `TestDeleteUnusedImportsMetadata_NoGoroutineLeak` fail 3/3 runs, and inverting the membership condition (`ok` -> `!ok`) makes `TestDeleteUnusedImportsMetadata_SkipsProtectedKeepsUnprotected` fail. `task build` and `task test:unit paths="./pkg/cleaning/..."` pass, and the new tests are stable under `-race -count=5`.

Two things block a merge as-is.

**1. `Fixes #7862` closes an issue that names three leak sites; this PR fixes one.**

Still present on current `main`:

- `protectRelativeStageDescSetByStageDesc`, `pkg/cleaning/cleanup.go:1213` — `for platformStageDesc := range m.stageManager.GetStageDescSet().Iter()` with a `break` at :1217, inside the loop over index platforms. `GetStageDescSet()` returns a `Clone()` (`pkg/cleaning/stage_manager/manager.go:296` -> `stages.go:63`), so every abandoned iteration pins a **full copy** of the stage set forever, not just a goroutine stack. For a multi-platform project at 68k stages this is the same order of magnitude as the reported heap growth — arguably a bigger leak than the one fixed here.
- The legacy parent lookup, `pkg/cleaning/cleanup.go:1244`, with a `break` at :1248. The set is hoisted out of the loop, so it leaks one goroutine per visited stage rather than a clone, but this is also the O(n x n) scan the report calls out.
- One-shot abandons at `cleanup.go:210` (`continue loop`) and `cleanup.go:343` (`continue Loop`) — one goroutine each, cosmetic.

`cleanupFinalStages`, also named in the issue, was already fixed on `main` by 72162e50c (#7867); this branch still carries the old `FilterOutFinalStages` version because it is based on 101565201.

Please either fix :1213 and :1244 with the same technique (build a `map[digest]*image.StageDesc` index once, `ToSlice()` instead of `Iter()`), or change `Fixes #7862` to `Refs #7862` so the issue stays open for them.

**2. The PR description overstates what is fixed.**

In the reported run, `Processing relative stages for saved stages: 2501.84 seconds` comes from the legacy parent lookup at :1244, which this PR does not touch — the 79-minute cleanup will not get meaningfully faster. What this change fixes is the heap growth in the import-metadata phase. werf squashes, so this description becomes the commit body; please narrow the claim there.

Note also that the branch now conflicts with `main` in `pkg/cleaning/cleanup_test.go` (both appended tests at the end of the file); `cleanup.go` merges cleanly.

pujitha24:
@alexey-igrychev I've pushed changes addressing your review — the branch is now at `b091951` and CI is green. Could you take another look when you have a moment? Happy to keep iterating if anything is still off.

alexey-igrychev (approved):
Thank you, @pujitha24, for the focused fix and regression coverage! The import-metadata cleanup fix is correct and approved as-is. This supersedes my earlier request for changes.

PR #7865, opened 2026-09-17
fix(deploy): render included charts at the project root

Fral738:
## Summary

Fix empty successful renders when `deploy.helmChartDir: .` selects a chart supplied by includes at the project root. The affected setup has a committed include with `to: /` supplying `Chart.yaml` and `templates/`, while the local project directory contains only `werf-includes.yaml` and its committed lock file. With that project under `ci/`, `werf render --dir ci` now renders the included templates instead of silently omitting them.

## What

- Included chart files are loaded for relative root paths `.`, `./`, and `chart/..`; equivalent subdirectory paths such as `chart` and `./chart/` select the same files.
- Included templates, metadata, and values at the selected chart path now participate in chart loading, so root-chart projects can produce resources and values that were previously ignored.
- Local files continue to take precedence over included files at the same path; selecting `chart/` does not include files from `chart-other/`.
- VERIFIED: a root-mounted included chart renders its ConfigMap without access to Docker or a Kubernetes cluster on macOS.
- No flags, configuration defaults, or requirements for an explicit `Chart.yaml` are added.

## Why

The include filter previously treated the root directory as the literal prefix `./`, but include destinations are normalized relative paths such as `Chart.yaml` and `templates/cm.yaml`. Every included chart file was therefore rejected. Local include configuration and lock files kept the loaded directory nonempty, while default chart metadata allowed rendering to succeed without the missing templates.

Normalize the selected directory and use an empty prefix for the project root, sharing that prefix between file selection and chart-relative names. Rejecting root charts or requiring an explicit `Chart.yaml` would change existing supported behavior rather than repair the include filter.


Fral738:
## Verification

- Manual CLI smoke on macOS: create two temporary committed Git repositories, include `/helm-chart` from the source repository into `/` of an application project under `ci/`, run `werf includes update`, commit the lock file, then run the freshly built `werf render --dir <app>/ci`. With `DOCKER_HOST` pointing to a nonexistent socket and `KUBECONFIG` pointing to a nonexistent file, render exited successfully and emitted ConfigMap `included-root-chart` from the include. This was a local CLI check, not a Linux or cluster deployment test.
- Mutation: restore the original root-prefix handling -> 5 of 7 `LoadChartDir` table entries fail on missing files.
- Mutation: retain project-relative names instead of chart-relative names -> all 3 subdirectory entries fail.
- Mutation: remove the processed-file guard -> all 3 local-override entries fail on duplicate `values.yaml`.
- Mutation: remove the trailing slash from the directory-membership prefix -> all 3 subdirectory entries fail because sibling include files are admitted.
- Each mutation was reverted immediately; the restored suite passed 7/7 after every mutation. These four mutations were rerun after switching the expected-data reads to `os.ReadFile`.
- Local full unit verification was not green: the 46-suite run reported failure only for `pkg/true_git`, despite that suite reporting 51/51 passing specs. An isolated rerun then failed one setup step because GPG signing returned `Cannot allocate memory`. No test exclusions or signing configuration changes were applied to hide those failures.

## Review focus

- Check that the same normalized prefix controls both include selection and chart-relative names, while local-file precedence and subdirectory boundaries remain unchanged.

pkg/giterminism_manager/file_manager/file_manager_test.go
```diff
@@ -0,0 +1,52 @@
...
+
+var _ = ginkgo.Describe("LoadChartDir", func() {
+	ginkgo.DescribeTable("loads included chart files relative to the chart directory",
+		func(ctx ginkgo.SpecContext, chartDir, destination string, localOverride bool) {
+			manager, projectDir, includedFiles := newChartFileManager(ctx, destination, localOverride)
+
+			files, err := manager.LoadChartDir(ctx, chartDir)
+			gomega.Expect(err).NotTo(gomega.HaveOccurred())
+
+			expected := includedFiles
```
reyreavman: Minor: `expected := includedFiles` — алиас на карту, которую вернул хелпер, дальше она мутируется (`expected["values.yaml"]`, `expected[name]`). Сейчас безопасно (хелпер создаёт свежую карту на каждый spec), но `maps.Clone(includedFiles)` сделал бы намерение явным и защитил бы от будущего переиспользования фикстуры между записями таблицы.

pkg/giterminism_manager/file_manager/file_manager_test.go
```diff
@@ -0,0 +1,52 @@
...
+			if localOverride {
+				expected["values.yaml"] = "source: local\n"
+			}
+			if destination == "/" {
```
reyreavman: Наблюдение (не дефект PR): при `to: /` в чарт попадают `werf-includes.yaml` и `werf-includes.lock` из локального проекта — тест это фиксирует как ожидаемое поведение. Это pre-existing: локальный каталог загружается целиком через `fileReader.LoadChartDir`, и до PR было так же. Просто отмечаю, что тест закрепляет это как контракт.

reyreavman (commented):
approve

reyreavman (approved):

PR #7866, opened 2026-09-17
fix(cleanup): avoid slow tag deletion in GitLab registries

Fral738:
## Summary

GitLab registry cleanup can spend most of its time deleting stages when the legacy tag endpoint is unavailable and werf falls back to deleting manifests by digest. Prefer the supported manifest-by-tag API so deleting a selected tag does not require digest deletion. This affects existing repositories with many tags, particularly when GitLab Registry uses filesystem metadata.

### Tag deletion

- The GitLab driver tries `DELETE /v2/<repo>/manifests/<tag>` first and accepts HTTP 200/202 without a new flag.
- Subsequent tag deletions reuse the successful endpoint and authorization scope; concurrent workers share the cache safely without serializing HTTP requests.
- VERIFIED: on GitLab Registry v4.41.0-gitlab with filesystem metadata, deleting selected tags preserves other tags pointing to the same manifest.
- Digest-only references use digest deletion, including after a tag-deletion method has been cached, rather than deleting a synthesized `latest` tag.

### Compatibility and errors

- An endpoint-level 404, HTTP 405, or structured `DIGEST_INVALID` retains the previous `/tags/reference/<tag>` fallback, followed by the existing digest fallback if the legacy endpoint is unavailable.
- HTTP 404 with `MANIFEST_UNKNOWN` or `NAME_UNKNOWN` is a successful no-op in both tag APIs, including cached calls, without escalating to digest deletion.
- Missing-resource codes on HTTP 500 remain errors; codes mentioned only in diagnostic text do not trigger fallback or suppress errors.
- Modern tag deletion retries `UNAUTHORIZED` with wildcard scope after push/pull/delete scopes; denial of both attempts stops deletion instead of trying another API.
- Deletion errors use `DELETE <path>: <registry error>` and preserve typed HTTP status and registry codes, without adding a duplicate status prefix.

### Cleanup policies

- Retention policies, dependency traversal, dry-run decisions, and other registry drivers remain unchanged.
- UNVERIFIED: cleanup duration on a repository with approximately 68,000 tags has not been measured; rerun that workload and compare completion time and registry DELETE logs.

## Why

The GitLab driver previously tried only the legacy tag-deletion route before falling back to manifest deletion by digest. Current GitLab Registry exposes tag deletion on the manifest route, while digest deletion on filesystem storage also looks up referencing tags. Requesting the selected tag directly avoids that digest-deletion work and preserves sibling tags.

Keep compatibility fallbacks rather than replacing the legacy route outright: older registries can reject a tag where they expect a digest. Distinguish a missing resource from an unsupported endpoint in both tag APIs so a disappearing tag cannot escalate into deletion of a shared manifest. Treat an already absent resource as successfully deleted, and publish the selected method atomically because cleanup workers share the driver.


Fral738:
## Verification

- Live smoke rerun after the review fixes: a temporary client compiled with `task build` used the public werf GitLab driver against `registry.gitlab.com/gitlab-org/build/cng/gitlab-container-registry:v4.41.0-gitlab` on local Docker with filesystem storage and `database.enabled=false`. Four tags shared one manifest. Registry logs showed DELETE by tag: 202 for `drop` and `next`, then 404 `MANIFEST_UNKNOWN` for an already deleted `drop` through a fresh driver. The last call returned nil; `keep` and `latest` remained readable. No digest DELETE was sent.
- The live smoke did not reproduce the reported `prefer` startup state: the fresh test storage required a reachable metadata database in that mode. Filesystem deletion was exercised with the database explicitly disabled.
- All 24 focused GitLab cases passed with `task test:unit paths=./pkg/docker_registry -- -focus=GitLab -race`, including eight simultaneous cold-cache deletions followed by a warm-cache wave. Endpoint compatibility fixtures send real HTTP 404/405 responses with a custom reason phrase.
- Mutation: change recognized 404/405 statuses -> three fallback cases failed; match error codes in rendered text -> the misleading diagnostic case detected a swallowed server error; disable missing-resource no-op -> all six modern/legacy missing-resource cases detected extra requests or errors; remove atomic cache publication -> the concurrent test produced DATA RACE on cache writes; remove the 404 restriction -> both 500 missing-code cases detected swallowed errors; restore the duplicate-status prefix -> the bodyless 500 case counted two status strings instead of one. Restored the implementation after every mutation and reran the focused suite successfully.
- Initial implementation tests also caught reverting to the obsolete endpoint, removing the digest-reference guard, and removing wildcard-scope retry. The revised missing-resource tests now require successful no-op behavior, not the earlier missing-image error.
- Formatting, build, and full lint passed. The full unit gate failed in untouched `pkg/true_git`: two BeforeEach setups could not create git commits because `gpg: signing failed: Cannot allocate memory`. On isolated rerun, all 51 specs passed but Ginkgo still returned a suite-level failure. Other suites passed.

## Review focus

- The supplied six minor findings are addressed: typed fallback matching, legacy missing-resource handling, atomic cache publication, idempotent deletion, concise errors, and an accurate endpoint-test name. Independent read-only challenge review found no implementation blockers; its two suggested error-boundary assertions were added and mutation-checked.
- Legacy API compatibility is exercised through HTTP fixtures, not a live older GitLab registry.

## Follow-up

- [ ] Run cleanup on a representative large GitLab filesystem registry and compare the DELETE endpoints and total duration.

pkg/docker_registry/gitlab_registry.go
```diff
@@ -52,8 +53,42 @@ func newGitLabRegistry(options gitLabRegistryOptions) (*gitLabRegistry, error) {
...
-		return r.deleteRepoImageFunc(ctx, repoImage)
+	if !isReferencedByTag(repoImage) {
+		return r.defaultImplementation.DeleteRepoImage(ctx, repoImage)
+	}
+
+	if deleteFunc := r.deleteRepoImageFunc.Load(); deleteFunc != nil {
+		return deleteFunc.(func(context.Context, *image.Info) error)(ctx, repoImage)
+	}
+
+	var deleteFuncErr error
```
reyreavman: Nit: `deleteFuncErr` is only ever compared with nil and the stored error is never returned — it acts as a bool flag ("already retried with universal scope"). A `bool` would express the intent more directly.

pkg/docker_registry/gitlab_registry.go
```diff
@@ -64,18 +99,22 @@ func (r *gitLabRegistry) DeleteRepoImage(ctx context.Context, repoImage *image.I
 	} {
 		if err := deleteFunc(ctx, repoImage); err != nil {
 			reference := strings.Join([]string{repoImage.Repository, repoImage.Tag}, ":")
-			if strings.Contains(err.Error(), "404 Not Found") {
+			var transportErr *transport.Error
+			if !errors.As(err, &transportErr) {
+				return err
+			}
+			if transportErr.StatusCode == http.StatusNotFound {
```
reyreavman: Non-blocking, pre-existing: the legacy loop still `continue`s on every `UNAUTHORIZED` without the second-failure short-circuit added to the modern loop above (L79-83). A persistent 401 here still costs the extra digest attempts before failing. Consistent handling would be nice but is out of scope for this fix.

pkg/docker_registry/gitlab_registry_test.go
```diff
@@ -0,0 +1,302 @@
...
+	ginkgo.Entry("accepts an OK response", http.StatusOK, "", http.StatusAccepted,
+		[]string{"DELETE /v2/project/image/manifests/stage", "DELETE /v2/project/image/manifests/next"}, ""),
+	ginkgo.Entry("falls back to the legacy tag endpoint on 404 and reuses it", http.StatusNotFound, "", http.StatusAccepted,
+		[]string{"DELETE /v2/project/image/manifests/stage", "DELETE /v2/project/image/tags/reference/stage", "DELETE /v2/project/image/tags/reference/next"}, ""),
+	ginkgo.Entry("falls back to the legacy tag endpoint on 405", http.StatusMethodNotAllowed, "", http.StatusAccepted,
+		[]string{"DELETE /v2/project/image/manifests/stage", "DELETE /v2/project/image/tags/reference/stage", "DELETE /v2/project/image/tags/reference/next"}, ""),
+	ginkgo.Entry("falls back when a registry requires a digest", http.StatusBadRequest, `{"errors":[{"code":"DIGEST_INVALID","message":"provided digest did not match uploaded content"}]}`, http.StatusAccepted,
+		[]string{"DELETE /v2/project/image/manifests/stage", "DELETE /v2/project/image/tags/reference/stage", "DELETE /v2/project/image/tags/reference/next"}, ""),
```
reyreavman: Suggestion: there is no entry pinning the "return immediately" branch on the modern endpoint for a bodiless 401 (no structured `UNAUTHORIZED` code). Today it returns without falling back, same as before the PR; an entry like `(http.StatusUnauthorized, "", …, "401 Unauthorized")` would make that explicit.

reyreavman (approved):

PR #7867, opened 2026-09-17
fix(cleanup): preserve retained final images

Fral738:
## Summary

With a stage published to both `--repo` and `--final-repo`, cleanup currently deletes its final counterpart even when cleanup policies retain the primary stage, unless Kubernetes protects the final image independently. Conversely, final images without a primary counterpart are retained indefinitely. Restore policy-based retention for the final repository.

## What

- Retain a final stage when the same stage digest and creation timestamp remain in the primary set after cleanup policy selection, without cloning and scanning the primary set for every final stage.
- Delete unprotected final stages whose primary counterparts are absent, including counterparts selected for deletion during the same cleanup.
- Previously accumulated, unprotected final-only stages now become eligible for deletion without an opt-in; preview the selection with `--dry-run` before cleaning an existing final repository.
- Preserve independent Kubernetes protection for final stages, including orphans, and keep its `used in Kubernetes` reason when a primary counterpart also exists.
- Record retained counterparts in cleanup JSON reports with reason `found in repo` instead of protecting orphans with `not found in repo`; the report schema is unchanged.
- Apply the same final-stage selection in dry-run without deleting images from either repository.
- Leave registry deletion APIs and error handling unchanged: final retention follows policy selection, not confirmation of primary deletion, so an unprotected final may be deleted after a nonfatal failure to delete its primary candidate.
- Leave primary retention policies and registry-specific effects on other tags sharing a deleted manifest unchanged.

## Why

The stage-set refactor replaced accumulation of orphaned final stages in a deletion list with marking those same stages as protected, without reversing the matching branch. Primary cleanup already removes deletion candidates from its in-memory stage set before final cleanup, so the refactor inverted the intended result: retained counterparts were deleted and orphans were protected.

Match by stage ID because primary and final descriptions refer to different repositories, using the existing primary stage ID index that is updated when deletion candidates are forgotten. Restore the original selection while preserving existing Kubernetes protection; changing deletion-failure bookkeeping or registry deletion semantics is outside this fix.

Fixes: d8ff9d9a7171183dc53a7d37dcab4ed5824ecdb4 ("refactor(cleanup): introduce StageDescSet and unify array handling")


Fral738:
## Verification

- Linux/Docker run against `registry:2` with `REGISTRY_STORAGE_DELETE_ENABLED=true`: the complete cleanup e2e suite passed 3/3 on `d90329bf270e`. The final-repository fixtures now have no Git remote. They verify retained pairs, pre-existing orphans, counterparts removed by the same cleanup, reports, and registry contents in dry-run and real cleanup. Command: `task test:e2e mode=go paths=./test/e2e/cleanup labelFilter=cleanup parallel=false` with a freshly built binary in `WERF_TEST_BINARY_PATH`.
- Mutation rerun after removing origin setup: run both final-repository CLI tests with a binary built from base `851eb025fb18` -> both failed because the retained final stage was missing from `kept`; rerun with the fixed binary -> both passed.
- Mutation on the indexed implementation: negate the StageID membership condition -> 10/14 unit cases failed on paired retention or orphan deletion; the four independently Kubernetes-protected cases survived as expected. Restore the implementation -> 14/14 passed.
- Earlier mutation checks: ignore final protection -> both Kubernetes-protected orphan cases failed; force replacement of the protection reason -> both paired Kubernetes cases failed; bypass dry-run -> the orphan dry-run case detected a real deletion call. Each mutation was restored before proceeding.
- Full formatting, build and lint passed after the review fixes. Full unit testing failed only in untouched `pkg/true_git`: one BeforeEach could not sign its test commit (`gpg: signing failed: Cannot allocate memory`). An isolated rerun hit the same setup failure in three specs; 48/51 passed.
- macOS CLI execution remains blocked by Docker Desktop access to the local registry, as established in the original run; executable CLI verification was completed on Linux. Live Kubernetes discovery and actual multiplatform registry publication were not exercised.

## Review focus

- All three supplied minor findings are addressed: use the existing StageID index, remove unnecessary origin/fetch setup from fixtures with Git history policy disabled, and align the fake storage manager receiver name.
- The primary index is cleared when deletion candidates are forgotten; the same-run deletion e2e checks this boundary. No new index or cache is introduced.
- Independent review found no blocker for restoring policy-based retention. Existing partial-deletion behavior remains: primary candidates are forgotten after nonfatal deletion errors, so an unprotected final can still be deleted. This limitation is stated in the description.
- The previous report-only test deliberately changes its expectation because it encoded orphan retention rather than the original cleanup contract.

## Follow-up

- [ ] Define confirmation-based final retention after a nonfatal primary deletion failure in `werf/werf`, with a regression that injects primary DELETE failure and checks final deletion behavior.

reyreavman (approved):

